PR #134, opened 2026-04-25
feat(updater): auto-check on every command with Homebrew awareness

shahariaazam:
## Summary

Implements #128. Adds an opt-out, cache-backed update check that runs before every `agento` subcommand, with first-class awareness of Homebrew-managed installs.

- New `internal/updater` package wraps release detection (24h on-disk cache, 5s timeout, semver comparison) and install dispatch (native self-update via `go-selfupdate`, or Homebrew-managed short-circuit).
- New `PersistentPreRunE` in `cmd/root.go` runs the check before each subcommand. Skipped for: dev/unknown builds, `update`/`help`/`completion`/`__complete` subcommands, non-TTY environments (stdin AND stdout), and `AGENTO_SKIP_UPDATE_CHECK=1|true`.
- When an update is available and the install is Homebrew-managed, prints `brew upgrade agento` and continues without modifying the binary. Otherwise prompts on stderr; on confirmation, self-updates and exits 0.
- `agento update` refactored onto the same package and passes `forceFresh=true` so the explicit command always bypasses the 24h cache.

## Acceptance criteria (from #128)

- [x] Auto-check on `agento web`/`ask`/etc., at most once per 24h via on-disk cache
- [x] Yes/no prompt when an update is available
- [x] Homebrew installs print `brew upgrade agento` instead of self-updating
- [x] Non-TTY / CI silently skipped (both stdin and stdout)
- [x] `AGENTO_SKIP_UPDATE_CHECK=1` disables entirely
- [x] `agento update` bypasses the 24h cache
- [x] Cross-platform: Homebrew detection short-circuits on Windows; cache uses `filepath.Join`
- [x] No regression to web UI `GET /api/version/update-check`
- [x] Dev/unknown builds skipped via `ErrNotReleaseBuild` and explicit guard

## Test plan

- [x] `go build ./...`
- [x] `go test ./...` (15 new unit tests covering cache hit/miss/stale, dev-build skip, version mismatch, force-fresh, no-cache mode, corrupted cache fallthrough, timeout propagation, and all Homebrew path layouts incl. negative cases)
- [x] `make lint` (0 issues)
- [ ] Manual smoke on a release build: `AGENTO_SKIP_UPDATE_CHECK=1 agento web` skips; non-TTY (`agento ask < /dev/null`) skips; Homebrew layout prints upgrade hint without modifying binary

Closes #128

---
_Generated by [Claude Code](https://claude.ai/code/session_01CbSNk8rhf2DAxGHgUxZjJ6)_

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -1,22 +1,51 @@
 package cmd
 
 import (
+	"context"
+	"errors"
 	"fmt"
 	"os"
+	"strings"
 
+	"github.com/mattn/go-isatty"
 	"github.com/spf13/cobra"
 
 	"github.com/shaharia-lab/agento/internal/build"
 	"github.com/shaharia-lab/agento/internal/config"
+	"github.com/shaharia-lab/agento/internal/updater"
 )
 
+// skipUpdateCheckEnv is the environment variable users can set to disable the
+// pre-run update check entirely (useful in CI, scripts, or any environment
+// where prompting or network calls are unwanted).
+const skipUpdateCheckEnv = "AGENTO_SKIP_UPDATE_CHECK"
+
+// updateCheckSkipCommands lists subcommand names that must never trigger the
+// auto-update check. The "update" command runs its own (uncached) check, and
+// help/version are non-interactive metadata commands.
+var updateCheckSkipCommands = map[string]struct{}{ //nolint:gochecknoglobals
+	"update":     {},
+	"help":       {},
+	"completion": {},
+	"__complete": {}, // cobra's hidden shell-completion command
+}
+
 // NewRootCmd returns the root cobra command wired with the provided AppConfig.
+//
+// The root command attaches a PersistentPreRunE hook that performs an opportunistic
+// update check before any subcommand runs. The hook is intentionally best-effort:
+// it never returns an error to the caller, so a failed network call or a missing
+// cache directory cannot prevent the user's command from running.
 func NewRootCmd(cfg *config.AppConfig) *cobra.Command {
 	root := &cobra.Command{
 		Use:     "agento",
 		Short:   "Agento — AI Agents Platform",
 		Long:    "A platform for running Claude agents defined in YAML configuration files.",
 		Version: build.String(),
+		PersistentPreRunE: func(cmd *cobra.Command, _ []string) error {
+			runAutoUpdateCheck(cmd, cfg)
+			return nil
+		},
 	}
 	root.SetVersionTemplate("{{.Version}}\n")
 	return root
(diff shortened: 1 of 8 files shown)


shahariaazam:
@claude can you please review?

shahariaazam:
Thanks for the review. Pushed `09fc275` addressing the actionable items.

### Addressed

| # | Severity | Finding | Resolution |
|---|----------|---------|------------|
| 1 | MUST FIX | `agento` (no subcommand) and `--help` could trigger the prompt | `shouldRunAutoCheck` now returns `false` when `cmd == nil`, `!cmd.Runnable()` (bare `agento`), or when `--help` is set on any subcommand. New tests `TestShouldRunAutoCheck_SkipNonRunnable` and `TestShouldRunAutoCheck_SkipHelpFlag` cover both paths. |
| 2 | SHOULD FIX | `fmt.Scanln` rejects "yes" and leaves bytes in stdin | Switched to `bufio.NewReader(os.Stdin).ReadString('\n')` + `strings.TrimSpace(strings.ToLower(...))`. Accepts `y` / `yes` case-insensitively and consumes the full line so downstream `agento ask` reads from a clean stdin. |
| 4 | SHOULD FIX | Cache file shared across machines could return assets for the wrong platform | Added `Platform string` (`runtime.GOOS + "/" + runtime.GOARCH`) to `CheckResult`. Cache entries with a different (or empty/legacy) platform are invalidated. New tests `TestChecker_Check_BypassesCacheWhenPlatformDiffers` and `TestChecker_Check_LegacyCacheMissingPlatformIsInvalidated` cover this. |
| 7 | CONSIDER | `Install()` ignored `latestVersion` and could install a newer release than the user agreed to | Added a guard: if `release.Version()` doesn't match `latestVersion`, fail with "release on GitHub (X) no longer matches the version you confirmed (Y); please re-run". |
| 8 | CONSIDER | Cache dir was `0o750` while the file was `0o600` | Tightened directory perms to `0o700` for consistency. |
| Q | QUESTION | `vdev` test case origin | Confirmed the build pipeline never injects that — removed the case from `TestShouldRunAutoCheck_SkipDevBuild`. |

### Deferred (with rationale)

- **#3 — `os.Exit(0)` after self-update.** Worth a separate discussion. The current behavior matches the intent of issue #128 ("update completed; user re-runs the new binary"), and changing the exit code is observable to anyone scripting around `agento`. Happy to switch to `os.Exit(2)` (or print + return) if you have a preference — let me know which you'd like.
- **#5 — Duplicate cache between `internal/api/version.go` and `internal/updater`.** Acknowledged. The new `updater.Checker` already supports a `CacheTTL` field that would let the API handler reuse it (1h vs. 24h). Tracking as a follow-up so this PR stays scoped to the issue. Want me to open the issue?
- **#6 — Snooze / "skip this version" mechanism.** Not in the original issue spec; opening a follow-up issue makes sense.
- **#9 — Observability (`AGENTO_DEBUG`).** Defer to a follow-up — would touch every silent error path.
- **#10 — Refactor `runAutoUpdateCheck` for injectability.** The orchestration is thin (skip-rules → checker → prompt → install) and each piece is independently tested. Would prefer to leave it until a real bug motivates a wider refactor.

### Verification
- `go build ./...` ✅
- `go test ./...` ✅ (15 new+modified unit tests)
- `golangci-lint run ./...` → 0 issues

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #167, opened 2026-08-07
chore: restrict dependabot to patch-only updates and pin all npm deps to exact versions

vibexp-agent:
## Why
Reduce supply-chain risk: no automatic minor/major dependency upgrades, and every frontend dependency pinned to an exact patch-level version.

## Changes
- **`.github/dependabot.yml`** — gomod and npm ecosystems now ignore `semver-major` and `semver-minor` updates, so dependabot only opens patch-level PRs (still grouped, weekly, max 1 open). GitHub Actions updates stay grouped; all actions are already pinned to full commit SHAs (#158).
- **`frontend/package.json`** — all 36 dependencies/devDependencies pinned from `^x.y.z` to the exact versions currently resolved in `package-lock.json`. No version changes — verified by diffing the lockfile (only an extraneous `react-is` entry was dropped).
- **`frontend/.npmrc`** — `save-exact=true` so future installs stay pinned.
- **Go** — `go.mod` already pins every module to an exact semver including patch; no changes needed.

## Verified
- `npm run typecheck` passes
- `go build ./...` passes


vibexp-agent:
Superseded by #169 — this branch was based on main before #164 merged, so its pinned versions were stale.

PR #195, opened 2026-08-08
fix(claude-sessions): count conversational turns, not JSONL events (#182)

### LINKED ISSUE #182: fix(claude-sessions): message_count counts tool_result carrier events — 93% of "messages" are not messages
## Problem

`message_count` in the Claude Sessions list is not a message count. It counts raw JSONL events, and the overwhelming majority of `user` events are now tool-result carriers rather than anything a human typed.

`processSummaryEvent` (`internal/claudesessions/scanner.go:585-601`) increments `MessageCount` for **every** non-sidechain `user` event and **every** `assistant` event.

## Evidence

Across 739 local session files:

- **59,357 of 63,798** user events (**93%**) carry only `tool_result` blocks — no human input at all
- assistant events include every pure tool-call turn, not just visible replies

So the number the UI labels "messages" is really an API-round-trip count, inflated roughly an order of magnitude relative to what a user would count as a message.

The insight pipeline already gets this right: `isTurnStart` (`internal/claudesessions/processor.go:167`) explicitly returns false for user events whose content contains a `tool_result` block. The scanner simply never got the same treatment, so `session_insights.turn_count` and `claude_session_cache.message_count` disagree with each other for the same session — by a large factor.

## Impact

Low severity but high visibility: it is one of the few numbers shown on every row of the sessions list, and it is wrong enough to be misleading when comparing sessions.

## Proposed direction

Extract the `isTurnStart` predicate into one shared helper used by both the scanner and the processor, so the two can never drift again. Fix `message_count` to mean genuine conversational turns, and keep the raw event count under a separate, explicitly-named column rather than silently repurposing the existing one.

Backfill note: existing rows will not refresh on their own — file mtimes have not changed, so `diffDiskAndCache` classifies every file as unchanged. A forced full re-read is required.

Found during a JSONL schema-drift audit of `internal/claudesessions/`.

---
## WHAT

**Summary:** Make `message_count` mean conversational turns rather than raw JSONL events, using the same predicate the insight pipeline already applies, and keep the raw count under an explicitly-named column.

**Scope:**
- Extract the existing `isTurnStart` predicate into one shared helper used by both the scanner and the processor
- Redefine `message_count` as genuine user turns + assistant messages containing at least one `text` block
- Add `event_count` preserving today's raw value
- Force a one-time re-read so historical rows are recomputed

**Affected Components:**
- `internal/claudesessions/processor.go` — `isTurnStart` becomes the shared helper
- `internal/claudesessions/scanner.go` — `processSummaryEvent`, `processDetailUserEvent`, `processDetailAssistantEvent`, `upsertCacheRow`, `loadAllSessions`
- `internal/claudesessions/types.go` — `ClaudeSessionSummary.EventCount`
- `internal/claudesessions/cache.go` — `loadAll`
- `internal/storage/sqlite.go` — migration: one column
- `frontend/src/types.ts`, `ClaudeSessionsPage.tsx`

**Acceptance Criteria:**
- [ ] For a session whose events are 1 real user prompt, 8 assistant turns (3 with text, 5 tool-call-only) and 5 tool_result carrier events, `message_count == 4` and `event_count == 14`
- [ ] `message_count` for a given session equals `session_insights.turn_count` + the assistant text-reply count — the scanner and the insight pipeline no longer disagree
- [ ] A `user` event whose content is a plain JSON string (not a block array) still counts as a turn
- [ ] `isSidechain` events remain excluded from both counters
- [ ] After migration, existing rows are recomputed without a manual cache wipe
- [ ] The sessions list label reads as message/turn count consistently with the new semantics

**Out of Scope:**
- Changing `session_insights.turn_count` or any insight metric
- Counting sub-agent messages — that rolls up via #179
- Reworking how the detail view paginates or renders messages

## WHY

**Business Value:** `message_count` appears on every row of the sessions list. A number that is wrong by roughly an order of magnitude makes cross-session comparison meaningless, which is the main thing the list is for.

**Problem Statement:** `processSummaryEvent` (`scanner.go:585-601`) increments `MessageCount` for every non-sidechain `user` event and every `assistant` event. In the reference corpus 59,357 of 63,798 user events (93%) carry only `tool_result` blocks — no human input at all — and the assistant side counts every pure tool-call turn. The insight pipeline already excludes these via `isTurnStart` (`processor.go:167`), so the same session reports two very different "sizes" depending on which table is read.

**Expected Impact:** Displayed counts drop sharply and become comparable across sessions. Two code paths that encode the same concept collapse into one helper, removing a standing drift risk. No effect on tokens, cost or insights.

**Context:**
- Related: #179 — sub-agent messages roll into the parent there, not here
- Depends on the `scanner_version` forced-rescan mechanism specified in #181; if #181 has not landed, add it here and #181 reuses it
- Verified against `main` at `b72f09b`

## HOW

**Technical Approach:** Promote `isTurnStart` from an unexported processor helper to the package-level predicate both call sites use, then split the counter in two. The predicate already exists and is already correct — this is deduplication plus a column, not new logic.

**Implementation Details:**

### 1. Shared predicate — `internal/claudesessions/processor.go`
`isTurnStart` (`processor.go:167`) currently takes a `ProcessableEvent`. Retarget it at the two fields it actually reads so both event structs can use it:
```go
// isUserTurn reports whether a user event represents genuine human input —
// non-sidechain, and not merely a carrier for tool_result blocks.
func isUserTurn(isSidechain bool, content json.RawMessage) bool

// isAssistantReply reports whether an assistant message contains at least one
// text block, i.e. something the user actually saw.
func isAssistantReply(content json.RawMessage) bool
```
Keep `isTurnStart(ev ProcessableEvent)` as a thin wrapper so `turn_processor.go`, `autonomy_processor.go` and `depth_processor.go` are untouched.

### 2. Count — `internal/claudesessions/scanner.go`
In `processSummaryEvent` (`scanner.go:585`):
```go
case "user":
    if ev.IsSidechain { return }
    summary.EventCount++
    if ev.Message != nil && isUserTurn(ev.IsSidechain, ev.Message.Content) {
        summary.MessageCount++
        if summary.Preview == "" { … }   // unchanged
    }
case "assistant":
    summary.EventCount++
    if ev.Message != nil && isAssistantReply(ev.Message.Content) {
        summary.MessageCount++
    }
    …                                     // model + usage unchanged
```
Note the preview assignment moves inside the turn check — a bonus fix, since today the preview can be seeded from a `tool_result` carrier. #181 makes preview a last-resort fallback anyway, but the two changes are independent and both correct.

Apply the same split in `processDetailUserEvent` (`scanner.go:696`) and `processDetailAssistantEvent` (`scanner.go:712`) so `ClaudeSessionDetail.MessageCount` agrees with the summary. The message list itself keeps every event — only the counter changes.

### 3. Persist — `internal/storage/sqlite.go`
Migration at the next free version (11 unless #179/#180/#181 land first):
```sql
ALTER TABLE claude_session_cache ADD COLUMN event_count INTEGER NOT NULL DEFAULT 0;
```
Add `event_count` to `upsertCacheRow` (`scanner.go:420`), `loadAllSessions` (`scanner.go:465`) and `Cache.loadAll` (`cache.go:130`).

### 4. Recompute history
Bump `CurrentScannerVersion` (see #181) so `diffDiskAndCache` re-reads every cached file. No insight version bump needed — insights are unaffected.

### 5. Frontend
- `frontend/src/types.ts:531` area — add `event_count`.
- `ClaudeSessionsPage.tsx` — keep showing `message_count`; expose `event_count` only in the detail view or a tooltip, so the list stays scannable.

### Testing Strategy
- Unit: `scanner_test.go` — the fixture from the first acceptance criterion, asserting both counters exactly.
- Unit: string-content user event counts as a turn (guards the `parseContentBlocks` early return on non-array content at `processor.go:155`).
- Unit: cross-check — run the processor pipeline and the scanner over one fixture and assert `message_count == turn_count + assistantReplies`.
- Regression: sidechain events excluded from both counters.

- Doc comment on `ClaudeSessionSummary.MessageCount` (`types.go:32`) restated — it currently says "user + assistant top-level messages", which is what made the drift easy to miss.
- No new dependencies, env vars or config.

**Files to Create:** none

**Files to Modify:** `internal/claudesessions/{processor,scanner,types,cache}.go`, `internal/storage/sqlite.go`, `frontend/src/types.ts`, `frontend/src/pages/ClaudeSessionsPage.tsx`

**Risks:**
- *Users notice counts dropping and read it as data loss* → `event_count` preserves the old number and the release note explains the change.
- *`isAssistantReply` hides genuinely tool-only turns from the count* → that is the intent; `event_count` and `tool_calls_total` both remain available for the fuller picture.
- *Migration number collision with #179 / #180 / #181* → whichever lands first takes 11; rebase the rest.

**Complexity:** Low

---

vibexp-agent:
Closes #182

## What

`message_count` was never a message count. `processSummaryEvent` incremented it for **every** non-sidechain `user` event and **every** `assistant` event, so it reported API round-trips.

This redefines it as conversational turns and preserves the old number under a new `event_count` column:

- **`message_count`** — user events carrying genuine human input, plus assistant events containing at least one `text` block.
- **`event_count`** — the raw top-level user + assistant event total, i.e. exactly what `message_count` used to hold.

## Why

Across 739 local session files, **59,357 of 63,798 user events (93%)** carry only `tool_result` blocks — nothing a human typed — and the assistant side counted every pure tool-call turn. The number was inflated roughly an order of magnitude on one of the few figures shown on every row of the sessions list, which made cross-session comparison meaningless.

The insight pipeline already got this right via `isTurnStart`, so the scanner and `session_insights.turn_count` disagreed by a large factor for the same session. Two code paths encoding one concept is a standing drift risk; this collapses them into a shared predicate.

## How

**Shared predicate** (`internal/claudesessions/processor.go`) — `isUserTurnContent` (genuine input, not a `tool_result` carrier) and `isAssistantReply` (contains a text block). `isTurnStart` becomes a thin wrapper, so `turn_processor.go`, `rhythm_processor.go` and `depth_processor.go` are untouched and no insight metric changes.

**Deviation from the issue's HOW, deliberate:** the HOW proposed `isUserTurn(isSidechain bool, content)`. The sidechain check is instead left to each caller, because the flag means "delegated work, skip" in a parent transcript but is set on *every* event of a sub-agent transcript — where `readSubagentSummary` must count those turns. Baking the flag into the predicate would have silently degraded sub-agent counts to assistant-only. This is the only shape both callers can share.

**Count** (`scanner.go`) — the user and assistant cases are extracted into `addSummaryUserEvent` / `addSummaryAssistantEvent` (keeping `processSummaryEvent` under the `gocyclo` threshold). The same split is applied to `processDetailUserEvent` / `processDetailAssistantEvent` so the detail view agrees with the list. **Every event still renders** — only the counters changed. Seeding the preview is now gated on the same predicate, so it can no longer be taken from a `tool_result` carrier.

**Persist** — migration **14** adds `event_count` to `claude_session_cache`; threaded through `upsertCacheRow`, `sessionSummarySelect` and its positional `Scan`.

**Recompute history** — `CurrentScannerVersion` **2 → 3**. Existing rows hold the old inflated number and their file mtimes never change, so only a version bump re-reads them.

**Frontend** — `event_count` added to `ClaudeSessionSummary`; the list keeps showing `message_count` and exposes the raw total in a tooltip, so the row stays scannable.

## Acceptance criteria

- [x] Fixture of 1 user prompt + 8 assistant turns (3 text, 5 tool-call-only) + 5 carriers → `message_count == 4`, `event_count == 14`
- [x] `message_count` equals `turn_count` + assistant text replies — scanner and insight pipeline agree
- [x] A `user` event with plain JSON string content still counts as a turn
- [x] `isSidechain` user events excluded from both counters
- [x] Existing rows recomputed after migration without a manual cache wipe
- [x] The list label reads consistently with the new semantics

## Tests

`internal/claudesessions/counts_test.go` covers each criterion, plus a detail-vs-summary agreement test and the scanner-version backfill path. The sub-agent fixture gained a `tool_result` carrier so sub-agent counting is pinned too. Each was verified to fail when the underlying fix is reverted.

## Deferred

- **#196** — sub-agent `EventCount` is computed but has no `claude_subagent_cache` column to persist to.
- **#197** — injected string-content user events (`system-reminder`, `command-message`, …) still count as turns; filtering them would change `turn_count`, which #182 put out of scope.

## Note for the release

Displayed counts drop sharply (~5.6× on a 120-transcript sample). That is the fix, not data loss — `event_count` retains the previous number.


vibexp-agent:
Review round 1 findings applied. Deferred to #196 (sub-agent `event_count` column) and #197 (injected string-content user events still counted as turns).

PR #200, opened 2026-08-08
feat(insights): attribute tool calls to skill, plugin and MCP server (#184)

### LINKED ISSUE #184: feat(insights): break down tool usage by skill, plugin and MCP server using attribution* fields
## Problem

Claude Code now attributes each assistant turn to the skill, plugin, MCP server and MCP tool that produced it. Agento's tool breakdown counts only the raw `tool_use` block name, so all plugin, skill and MCP activity is flattened into anonymous tool names.

`ToolUsageProcessor.Process` (`internal/claudesessions/tool_processor.go:20`) increments `toolBreakdown[b.Name]` and nothing else. The attribution fields are not in `rawEvent` (`scanner.go:23`) or `ProcessableEvent` (`processor.go:15`).

## Evidence

New top-level fields on `assistant` events, with occurrence counts across 739 local session files:

| Field | Events | Top values |
|---|---|---|
| `attributionMcpServer` | 64,474 | `claude.ai VibeXP` (29,615), `vibexp_io_vibexp_team` (20,501), `claude-in-chrome` (6,411) |
| `attributionMcpTool` | 64,474 | `vibexp_io_reply_to_feed_item` (27,863), `vibexp_io_post_to_feed` (11,549), `computer` (3,243) |
| `attributionSkill` | 57,160 | `lab-workflow:github-issue-to-pr` (27,248), `lab-workflow:review-pr` (12,937), `vibexp:prime` (4,222) |
| `attributionPlugin` | 55,151 | `lab-workflow` (44,238), `vibexp` (7,455), `lab-marketing` (2,337) |
| `effort` | 47,129 | `high` |

Sub-agent transcripts additionally carry `attributionAgent` and `agentId`.

## Impact

This is the single richest new signal in the JSONL for Agento's stated purpose. It turns "you made 59,361 tool calls" into "62% of your tool calls came from the `lab-workflow` plugin, and `github-issue-to-pr` alone drove 27k of them" — the kind of breakdown that lets a user actually decide what to keep, tune, or drop.

`effort` is a smaller but free addition: a per-turn reasoning-effort dimension that currently has no representation anywhere in Agento.

## Proposed direction

Extend the insight pipeline with per-skill / per-plugin / per-MCP-server breakdowns alongside the existing `tool_breakdown`, stored in `session_insights` and aggregated on the insights page. Requires a `CurrentProcessorVersion` bump (`processor.go:11`, currently `1`) so existing rows reprocess.

Found during a JSONL schema-drift audit of `internal/claudesessions/`.

---
## WHAT

**Summary:** Attribute every tool call to the skill, plugin and MCP server that produced it, using the `attribution*` fields Claude Code now stamps on assistant events, and expose the breakdowns on the insights page.

**Scope:**
- Parse `attributionSkill`, `attributionPlugin`, `attributionMcpServer`, `attributionMcpTool`, `attributionAgent` and `effort` from assistant events
- New insight processor producing per-skill / per-plugin / per-MCP-server call counts alongside the existing `tool_breakdown`
- Aggregate across sessions on the insights page, matching how `tool_breakdown` already aggregates
- Reprocess existing sessions via a processor-version bump

**Affected Components:**
- `internal/claudesessions/processor.go` — `ProcessableEvent`, `SessionInsight`, `CurrentProcessorVersion`
- `internal/claudesessions/attribution_processor.go` — new
- `internal/claudesessions/processor_registry.go` — register it
- `internal/claudesessions/scanner.go` — `rawEvent` gains the attribution fields
- `internal/storage/` — `session_insights` columns + `InsightStorer.GetSummary` aggregation
- `internal/api/`, `frontend/src/types.ts`, insights page

**Acceptance Criteria:**
- [ ] For a session where 10 assistant events carry `attributionSkill: "lab-workflow:review-pr"` and 5 carry none, `skill_breakdown["lab-workflow:review-pr"] == 10` and unattributed calls are counted under an explicit `""`/`built-in` key rather than dropped
- [ ] `mcp_server_breakdown` and `mcp_tool_breakdown` populate independently, so a server with several tools is countable both ways
- [ ] The sum of all `skill_breakdown` values plus the unattributed count equals the number of assistant events processed — no call is double-counted or lost
- [ ] The insights page shows a top-N skills and top-N plugins panel driven by SQL aggregation, consistent with how `TopToolTotals` is already merged in `GetSummary`
- [ ] Bumping `CurrentProcessorVersion` causes `NeedsProcessing` to return every existing session, and the worker reprocesses them without a restart
- [ ] Sessions with no attribution fields (older files) process cleanly with empty breakdowns

**Out of Scope:**
- Attributing sub-agent transcripts — those become readable in #179; this issue covers the parent session's events, and picks up sub-agents for free once #179 feeds them through the same registry
- Cost attribution per skill/plugin (needs #180's corrected pricing first)
- Any change to the existing `tool_breakdown` semantics

## WHY

**Business Value:** This is the single richest new signal in the JSONL for what Agento exists to do. It turns "you made 59,361 tool calls" into "62% of them came from the `lab-workflow` plugin, and `github-issue-to-pr` alone drove 27k" — the level of detail that lets someone actually decide what to keep, tune, or drop.

**Problem Statement:** `ToolUsageProcessor.Process` (`tool_processor.go:20`) increments `toolBreakdown[b.Name]` from the `tool_use` block and nothing else, so all plugin, skill and MCP activity is flattened into anonymous tool names. The attribution fields are absent from `rawEvent` (`scanner.go:23`) and `ProcessableEvent` (`processor.go:15`), so they never reach any processor.

**Expected Impact:** Measured on the reference corpus, this classifies 57,160 skill-attributed and 64,474 MCP-attributed assistant events that are currently indistinguishable from built-in tool use. `effort` adds a per-turn reasoning-effort dimension with no representation in Agento today (47,129 events, all `high` locally — more interesting on mixed workloads).

**Context:**
- Related: #179 — once sub-agent events flow through the same `ProcessorRegistry`, their `attributionAgent` and `agentId` are picked up by this processor with no extra work
- Related: #180 — per-skill *cost* attribution should wait for corrected pricing
- Field counts and top values across 739 session files, `2.1.177` → `2.1.224`
- Verified against `main` at `b72f09b`

## HOW

**Technical Approach:** Add the fields to the event structs, then follow the established processor pattern exactly — `ToolUsageProcessor` (`tool_processor.go`) is the template, including the `map[string]int` + `Finalize` + `Reset` shape and the JSON-blob storage `tool_breakdown` already uses.

**Implementation Details:**

### 1. Carry the fields — `scanner.go` + `processor.go`
Add to `rawEvent` (`scanner.go:23`) and `ProcessableEvent` (`processor.go:15`):
```go
AttributionSkill     string `json:"attributionSkill,omitempty"`
AttributionPlugin    string `json:"attributionPlugin,omitempty"`
AttributionMcpServer string `json:"attributionMcpServer,omitempty"`
AttributionMcpTool   string `json:"attributionMcpTool,omitempty"`
AttributionAgent     string `json:"attributionAgent,omitempty"`
Effort               string `json:"effort,omitempty"`
```
`ProcessableEvent` already carries `Raw json.RawMessage` (`processor.go:22`) for exactly this case, but promoting these to typed fields is cheaper than re-unmarshalling per processor, and they are needed on more than one.

### 2. New processor — `internal/claudesessions/attribution_processor.go`
```go
type AttributionProcessor struct {
    skills, plugins, mcpServers, mcpTools, efforts map[string]int
    unattributed int
}
func (p *AttributionProcessor) Name() string { return "attribution" }
```
`Process` counts one entry per **assistant** event (not per tool_use block — attribution is stamped at the turn level). An event with no `attributionSkill` increments `unattributed` so the totals reconcile. `Finalize` writes into new `SessionInsight` fields; `Reset` clears all maps — mirror `ToolUsageProcessor` (`tool_processor.go:37,52`) exactly.

Register in `processor_registry.go` alongside the existing eight.

### 3. Persist
`SessionInsight` (`processor.go:43`) gains `SkillBreakdown`, `PluginBreakdown`, `McpServerBreakdown`, `McpToolBreakdown`, `EffortBreakdown map[string]int` and `UnattributedCalls int`. Store the maps as JSON text columns, exactly as `tool_breakdown` is today:
```sql
ALTER TABLE session_insights ADD COLUMN skill_breakdown       TEXT NOT NULL DEFAULT '{}';
ALTER TABLE session_insights ADD COLUMN plugin_breakdown      TEXT NOT NULL DEFAULT '{}';
ALTER TABLE session_insights ADD COLUMN mcp_server_breakdown  TEXT NOT NULL DEFAULT '{}';
ALTER TABLE session_insights ADD COLUMN mcp_tool_breakdown    TEXT NOT NULL DEFAULT '{}';
ALTER TABLE session_insights ADD COLUMN effort_breakdown      TEXT NOT NULL DEFAULT '{}';
ALTER TABLE session_insights ADD COLUMN unattributed_calls    INTEGER NOT NULL DEFAULT 0;
```
Note this touches `session_insights`, not `claude_session_cache`, so it does **not** collide with the migrations in #179–#183 beyond the version number itself.

### 4. Reprocess
Bump `CurrentProcessorVersion` (`processor.go:11`) from `1` to `2`. `InsightStorer.NeedsProcessing` (`processor.go:129`) already returns rows below the current version, and `insight_worker.go`'s 5-minute rescan loop already drains them through the bounded pool — no new plumbing.

### 5. Aggregate & render
- `InsightAggregateSummary` (`processor.go:103`) gains `TopSkillTotals`, `TopPluginTotals`, `TopMcpServerTotals`, merged the same way `TopToolTotals` already is in `GetSummary`.
- `frontend/src/types.ts` — mirror the new fields.
- Insights page — two panels (top skills, top plugins) plus an MCP server row, reusing the existing tool-breakdown chart component.

### Testing Strategy
- Unit: `processors_test.go` — extend with an `AttributionProcessor` case; fixture of 15 assistant events (10 attributed to one skill, 5 bare) asserting the exact counts and that `10 + 5 == 15`.
- Unit: `Reset()` clears every map — the existing processor tests already cover this pattern for the other eight.
- Unit: events with no attribution fields produce empty maps, not nil, so JSON marshals to `{}` not `null`.
- Integration: seed a `session_insights` row at `processor_version = 1`, assert `NeedsProcessing` returns it and the worker fills the new columns.

- `CLAUDE.md` — the `internal/claudesessions/` bullet says "8 processors run in a single pass per file"; update the count and mention attribution.
- No new dependencies, env vars or config.

**Files to Create:** `internal/claudesessions/attribution_processor.go`

**Files to Modify:** `internal/claudesessions/{scanner,processor,processor_registry}.go`, `internal/claudesessions/processors_test.go`, `internal/storage/` insight store + migration, `internal/api/`, `frontend/src/types.ts`, insights page component, `CLAUDE.md`

**Risks:**
- *Reprocessing every session on upgrade is expensive* → it already happens by design on any version bump, through a bounded 4-goroutine pool with a 100-item queue (`insight_worker.go:18-25`); it is background work, not on a request path.
- *High-cardinality keys blow up the JSON blobs* → skill and plugin names are bounded by what the user has installed; cap stored keys at the top 50 per session and log if truncated rather than storing an unbounded map.
- *Attribution semantics change between Claude Code releases* → all fields are `omitempty`; absence degrades to the `unattributed` bucket rather than failing the scan.

**Complexity:** Medium

---

vibexp-agent:
Closes #184

## What

Claude Code stamps each assistant event with the skill and plugin whose instructions were in context. Agento counted only the raw tool name, so every plugin, skill and MCP call was flattened into an anonymous tool.

A new `AttributionProcessor` breaks tool calls down by **skill**, **plugin**, **MCP server**, **MCP tool** and **effort tier**, alongside the existing `tool_breakdown`. The insights page gains three panels reusing the existing chart, plus the built-in share.

## Why

This is the richest signal in the JSONL for what Agento is for. It turns *"you made 59,361 tool calls"* into *"`lab-workflow` drove 44k of them, and `github-issue-to-pr` alone 27k"* — the level of detail that lets someone decide what to keep, tune, or drop.

## How — and two deviations from the HOW

Both were measured against the local corpus **before** any code was written, and [documented on the issue](https://github.com/shaharia-lab/agento/issues/184#issuecomment-5227587500). The reviewer re-measured both independently over all 1,534 transcripts and found the evidence stronger than the original sample.

### 1. Counting is per `tool_use` block, not per assistant event

The HOW specified one count per assistant event. But Claude Code splits one assistant message into several JSONL events — thinking, text, each tool call — all sharing a `message.id` and therefore carrying **identical** attribution. Measured: 20,190 message ids span 2 events, 24,244 span 3; only 20,337 are single-event. Per-event counting would inflate every number by a variable factor.

Per-block is also the unit the WHY asks about, and it cannot double-count: 0 of 85,296 `tool_use` block ids repeat within a message id.

### 2. MCP attribution is parsed from the tool name, not the `attributionMcp*` fields

Those fields are **sticky** — they hold the last MCP tool touched and persist onto later, unrelated turns. Across all local transcripts, of the 32,762 MCP-attributed events that carry a `tool_use` block, only **1,623 (~5%)** agree with the block actually being called; the disagreements are dominated by `Bash` (15,485), `Edit` (4,656) and `Read` (3,353). The field also holds a display name (`claude.ai VibeXP`) that never equals the tool-name slug.

A breakdown built on them would confidently report numbers that are mostly wrong. Server and tool are instead parsed from the authoritative `mcp__<server>__<tool>` block name — still independently countable, satisfying AC 2. The fields are parsed and carried, just not counted.

### Consequences for AC 1 and AC 3

Unattributed calls are kept in their own `UnattributedCalls` counter rather than under a `"built-in"` key inside `skill_breakdown`, because a key inside the map would break AC 3's arithmetic. AC 3's reconciliation becomes **`sum(skill_breakdown) + unattributed == tool_calls_total`** — the same no-double-counting, no-loss invariant, against the meaningful denominator. Verified holding in 756/756 real sessions.

This matters at the UI too: **52% of tool calls locally are unattributed**, so the page states what share of tool calls the skills panel actually accounts for rather than implying it is the whole picture.

### Storage

Migration **16** adds six columns to `session_insights`, stored as JSON like `tool_breakdown`. `CurrentProcessorVersion` **3 → 4** makes `NeedsProcessing` return every existing session; the worker's existing rescan loop drains them with no restart.

## Acceptance criteria

- [x] 10 calls attributed to one skill + 5 bare → `skill_breakdown[skill] == 10`, unattributed counted explicitly
- [x] `mcp_server_breakdown` and `mcp_tool_breakdown` populate independently
- [x] `sum(skill_breakdown) + unattributed == tool_calls_total` — nothing double-counted or lost
- [x] Insights page shows top-N skills and plugins (plus MCP servers) from SQL aggregation, merged as `TopToolTotals` already is
- [x] The version bump makes `NeedsProcessing` return existing sessions
- [x] Sessions with no attribution fields process cleanly with empty breakdowns

## Tests

`processors_test.go` covers each criterion including the reconciliation, the per-block vs per-event distinction, MCP parsing against a deliberately stale field, and a raw-JSON decode through the real registry so a wrong `json:` tag cannot pass silently. `sqlite_session_insights_store_test.go` covers the upsert/scan round-trip, the aggregate column wiring (disjoint sentinel keys — a swapped pair fails), and the version-bump reprocessing path. Each was verified to fail when the corresponding code is reverted.

## Deferred

- **#201** — the aggregate does 4 scans of `session_insights` per request instead of 1.
- **#202** — `attributionAgent`, `effort_breakdown` and `mcp_tool_breakdown` are stored but not surfaced.

## Note

During the v3→v4 backfill window the unattributed share reads low, because the denominator includes not-yet-reprocessed rows while the numerator only counts reprocessed ones. It self-corrects as the worker drains the queue.


vibexp-agent:
Review round 1 findings applied: unattributed share surfaced (52% of tool calls locally), aggregate column wiring pinned by test, processor-version bump tied to a test, raw-JSON decode covered, frontend types synced. Deferred to #201 and #202.

Issue #189, opened 2026-08-08
feat(settings): model pricing management UI — view rate history and add effective-dated rates


vibexp-agent:
## Problem

Model rates are maintained by hand from provider pricing pages, but there is nowhere to maintain them. `pricingTable` (`internal/claudesessions/analytics.go:24`) is a compiled-in Go map, so every rate correction is a code change, a review, a release and an upgrade before a user's cost numbers become accurate again.

That cadence does not match how the data behaves. Provider rates move continuously and unpredictably: Opus went $15/$75 → $5/$25, Haiku $0.80/$4 → $1/$5, and Claude Sonnet 5 is on introductory pricing ($2/$10) that reverts to $3/$15 on 2026-09-01. Third-party providers (#187) move at least as often. A self-hosted single-binary tool cannot ship a release every time Moonshot adjusts a price.

There is also no way to **see** what rates are being applied. When a cost number looks wrong, the user has no path from "this seems too high" to "here is the rate that produced it" — the table is invisible at runtime, and after #186 lands there will be a rate *history* that is equally invisible.

## Impact

- Rates drift from reality between releases, and the user cannot tell that they have.
- Correcting a rate requires the maintainer, not the user, even though it is pure data entry.
- The effective-dating machinery from #186 has no user-facing surface, so its main benefit — "add a new rate, keep the old one for old usage" — is unreachable.
- Cost numbers are unauditable: nothing in the UI explains how a figure was derived.

## Proposed direction

A **Model Pricing** tab in Settings, alongside the existing tabs (`frontend/src/components/` — `AdvancedTab.tsx`, `MonitoringTab.tsx`, `NotificationsTab.tsx`, `ClaudeSettingsTab.tsx`, `AppearanceTab.tsx`), backed by CRUD endpoints over the #186 catalog.

Core capability:

- **List** every model the catalog knows, grouped by provider, showing the rate currently in effect.
- **Expand** a model to see its full rate history — each rate with its `effective_from` date and where it came from — so a past cost figure can be traced to the rate that produced it.
- **Add a new rate** for a model, effective from a chosen date. This is the central interaction and the whole point of #186: adding a rate must *append* to the history, never overwrite it, so last month's spend keeps last month's price. The form should make that consequence obvious rather than looking like an edit.
- **Add a model** that the catalog does not yet know, so a user on a provider Agento has never seen can price it themselves without waiting for a release.
- **Correct a rate** entered in error — distinct from adding a new one, and the UI must not blur the two.

[...]

Issue #186, opened 2026-08-08
feat(pricing): time-versioned model pricing catalog — usage must keep the price it was charged at


vibexp-agent:
## Problem

Agento prices every token at **today's** rate. `pricingTable` (`internal/claudesessions/analytics.go:24`) is a hardcoded Go map with one rate per tier and no notion of *when* that rate applied, and cost is computed on demand from session token totals (`buildSummary` `analytics.go:244`, `buildCostOverTime` `analytics.go:426`, `TokenProfileProcessor.Finalize` `token_processor.go:48`).

So when a provider changes its prices, **the entire history is silently re-priced at the new rate**. A user who spent $40 last month sees that month's number change the next time rates are edited, with no way to tell that it moved or why. For a spend dashboard this makes month-over-month comparison meaningless — the past is not supposed to be mutable.

This is not hypothetical. Anthropic's published rates have moved repeatedly, and at least one currently-live rate is explicitly time-boxed:

- **Claude Sonnet 5** is on introductory pricing of **$2 / $10** per MTok **through 2026-08-31**, reverting to **$3 / $15** on 2026-09-01. Today (2026-08-08) Agento would have to pick one, and whichever it picks is wrong for half the year.
- Opus dropped from **$15 / $75** (Opus 4.1 and earlier) to **$5 / $25** with Opus 4.5.
- Haiku moved from **$0.80 / $4** (3.5) to **$1 / $5** (4.5).

Because rates live in compiled Go, correcting or adding one also requires a code change and a release — which is the wrong shape for data that the maintainer refreshes by hand from provider pricing pages.

## Impact

- Historical cost figures mutate whenever rates are edited; there is no audit trail.
- The Sonnet 5 introductory rate cannot be represented at all — one of the two prices is always wrong.
- Adding or fixing a rate needs a recompile, so the table drifts from reality between releases.
- Cost is one of the headline numbers on the analytics page and per-session insight rows, so the error is user-visible.

## Proposed direction

Replace the hardcoded map with a **persisted, effective-dated pricing catalog**, and price usage at the moment it was actually spent.

- Each rate row carries an `effective_from` timestamp. Looking up a price takes `(model_id, spent_at)` and returns the latest rate whose `effective_from <= spent_at`. Adding a new rate never rewrites the old one, so last month stays priced at last month's rate.
- Rates live in SQLite, seeded from a built-in catalog on migration and re-seeded idempotently on startup for entries the user has not overridden.
[...]

Issue #180, opened 2026-08-08
fix(analytics): cache-write cost understated ~37% — all cache creation is now 1-hour ephemeral


vibexp-agent:
## Problem

Agento prices prompt-cache writes at the **5-minute** rate (1.25× input). In practice essentially all cache creation is now **1-hour** ephemeral cache, which bills at **2× input**. Cache-write cost is understated by ~37%.

`pricingTable` (`internal/claudesessions/analytics.go:24`) has a single `CacheWritePerMTok` per tier, and `rawUsage` (`scanner.go:43`) / `EventUsage` (`processor.go:34`) only decode the flat `cache_creation_input_tokens` — the nested `cache_creation` object that carries the split is never parsed.

## Evidence

Actual `message.usage.cache_creation` totals across the local corpus:

```
ephemeral_5m_input_tokens:         130
ephemeral_1h_input_tokens: 427,997,898   ← 99.99997%
```

At Opus rates that is $10/MTok versus the $6.25/MTok Agento assumes — roughly **$1.6K** of understated cache-write cost across this dataset alone.

Secondary pricing problems in the same table:

- `pricingForModel` (`analytics.go:34`) substring-matches `opus` / `sonnet` / `haiku` and falls back to Sonnet. Models present in real data that all silently price as Sonnet: `claude-fable-5` (7,841 events), `k3` (6,109), `glm-5.2` (8).
- The comment at `analytics.go:20-23` is stale — it cites Opus 4.5/4.6 and Haiku 4.5, while the data contains `claude-opus-4-8` and `claude-opus-5`.
- `<synthetic>` appears as a model on 51 events and should be excluded from cost entirely rather than priced as Sonnet.

## Impact

Cost is one of the headline numbers Agento shows. It is currently wrong in a consistent, one-directional way (always too low), across both the analytics page and per-session insight rows.

Found during a JSONL schema-drift audit of `internal/claudesessions/`.

---
## WHAT

**Summary:** Parse the nested `usage.cache_creation` split and price 1-hour cache writes at 2× input instead of the 5-minute 1.25× rate; stop silently pricing unknown models as Sonnet.

**Scope:**
- Decode `message.usage.cache_creation.{ephemeral_5m_input_tokens, ephemeral_1h_input_tokens}`
- Persist the split and bill each tier at its own rate
- Replace the substring-match model lookup with an explicit table plus an honest "unknown" path
- Exclude `<synthetic>` from cost entirely
[...]

PR #205, opened 2026-08-09
feat(pricing): price non-Anthropic models — Kimi K3, Qwen and GLM are billed as $0 (#187)

### LINKED ISSUE #187: feat(pricing): price non-Anthropic models — Kimi K3, Qwen and GLM are billed as $0
## Problem

Claude Code is routinely pointed at non-Anthropic, Anthropic-API-compatible backends, so Agento's session corpus contains models from several providers. Agento has pricing for none of them.

Distinct `message.model` values across the local `~/.claude/projects` corpus (740 sessions):

```
claude-opus-4-8                        98,307
claude-opus-5                          33,521
claude-fable-5                         13,193
k3                                      8,181   ← Moonshot Kimi K3
claude-haiku-4-5-20251001               5,889
<synthetic>                                69
glm-5.2                                     8   ← Z.ai / Zhipu GLM
mixedbread-ai/mxbai-embed-large-v1          7   ← embedding model
opus                                        4
claude-opus-4-7[1m]                         3
any                                         1
```

`k3` alone is **8,181 assistant events** — the third most-used model in the corpus, ahead of Haiku. Today `pricingForModel` (`analytics.go:34`) substring-matches and falls through to Sonnet, so Kimi K3 usage is invented as Anthropic spend. #180 correctly stops that lie by routing unpriced models to an "unknown pricing" bucket — but the result is that a top-three model contributes **$0.00** and the headline cost number is understated instead of overstated. Neither is the right answer: these models cost real money and their rates are published.

## Impact

- Roughly 8,200 events of real, billable usage report as either wrong-provider spend (today) or zero spend (after #180).
- Users running mixed Anthropic + third-party fleets — which is the whole point of Claude Code's provider flexibility — cannot see their true total.
- The "unknown pricing" bucket becomes permanently populated rather than being the rare exception it is designed to be.

## Proposed direction

Seed the pricing catalog with entries for the third-party models actually observed, sourced from each provider's published pricing page and recorded with the date the rate was retrieved.

Providers to cover, with rates retrieved 2026-08-08 (**re-verify at implementation time — these move**):

| Provider | Model | Input $/MTok | Output $/MTok | Cached input $/MTok |
|---|---|---|---|---|
| Moonshot | Kimi K3 (`k3`) | 3.00 | 15.00 | 0.30 |
| Z.ai / Zhipu | GLM-5.2 (`glm-5.2`) | 1.40 | 4.40 | 0.26 |
| Alibaba | Qwen3.8-Max | 2.00 | 6.00 | 0.25 |
| Alibaba | Qwen3.5 397B | 0.60 | 3.60 | — |
| Alibaba | Qwen3.5 Plus | 0.40 | 2.40 | — |
| Alibaba | Qwen3.5 Flash | 0.10 | 0.40 | — |
| Alibaba | Qwen3 Max | 0.78 | 3.90 | — |

Sources: Moonshot, Z.ai and Alibaba Cloud Model Studio published API pricing (International endpoints — note Alibaba's Mainland China endpoint is 60–70% cheaper, and Z.ai rates differ via OpenRouter, so the entry should record which endpoint it reflects).

Also worth resolving in the same pass, since they are the same class of "model ID that does not map to a rate":

- `<synthetic>` (69 events) — Claude Code's synthetic messages. #180 already zero-costs these; confirm it is a catalog entry rather than a special case.
- `mixedbread-ai/mxbai-embed-large-v1` (7 events) — an embedding model, not a chat model. Decide whether to price it or exclude it from cost entirely so it does not sit permanently in the unknown bucket.
- `opus` (4 events) and `any` (1 event) — alias values with no concrete model behind them.
- `claude-opus-4-7[1m]` — the 1M-context variant. Anthropic prices Opus 4.7's 1M window at standard rates with no long-context premium, so it should resolve to the base `claude-opus-4-7` rate via prefix matching rather than landing in the unknown bucket.

Cache-write pricing for third-party providers does not follow Anthropic's 1.25×/2× ephemeral split that #180 introduces; each provider's own cache semantics need to map onto the catalog's fields, or be recorded as unknown rather than guessed.

**Blocked by #186** (time-versioned pricing catalog) — that issue delivers the storage and resolver these rows go into, including the effective-dating that lets these rates be corrected later without rewriting history. **Depends on #180** for the unknown-pricing bucket these entries drain.

## Out of scope

- Auto-fetching rates from provider pricing pages. Rates are retrieved and entered manually; the catalog just has to make that cheap and auditable.
- Providers not present in the corpus (OpenAI, Gemini, DeepSeek, MiniMax, …) — add on demand.

---
## WHAT

**Summary:** Seed the pricing catalog with rates for the non-Anthropic models Claude Code is pointed at — Moonshot Kimi, Z.ai GLM, Alibaba Qwen — plus explicit handling for the alias and non-chat model IDs, so third-party usage stops reporting as $0.

**Scope:**
- Catalog entries for Moonshot Kimi K3, Z.ai GLM-5.2 and the Alibaba Qwen line, each with provider, published rate and a `source` recording the endpoint and the date retrieved
- Cache-read priced from each provider's published cached-input rate; cache-write seeded at the provider's plain input rate with the assumption recorded in `source`
- `mixedbread-ai/` prefix entry: zero-cost, non-billable — an embedding model, not chat usage
- `opus` bare alias resolves via the coarse family fallback and is flagged estimated; `any` stays unpriced
- `claude-opus-4-7[1m]` confirmed to resolve to the base `claude-opus-4-7` rate via prefix (Anthropic charges no long-context premium on Opus 4.7's 1M window)
- Confirm `<synthetic>` is a catalog row rather than a hardcoded special case

**Affected Components:**
- `internal/pricing/catalog.json` — new provider blocks; the bulk of this issue
- `internal/pricing/resolver.go` — a `Billable bool` on `Rate` so a zero-cost row is distinguishable from a $0.00 rate that simply hasn't been filled in
- `internal/pricing/resolver_test.go` — per-provider resolution cases
- `internal/storage/sqlite.go` — `billable` column if #186 did not already add it

**Acceptance Criteria:**
- [ ] `k3` resolves to Moonshot rates ($3.00 in / $15.00 out / $0.30 cached read per MTok) and the 8,181 corpus events on it report non-zero cost
- [ ] `glm-5.2` resolves to Z.ai rates ($1.40 / $4.40 / $0.26 cached read)
- [ ] Every Qwen entry resolves by its published model ID and records which endpoint (International vs Mainland China) the rate reflects — the two differ by 60–70%
- [ ] `mixedbread-ai/mxbai-embed-large-v1` contributes $0.00, is marked non-billable, and does **not** appear in the unknown-pricing bucket
- [ ] `opus` prices at the Opus-tier fallback and is reported as estimated; `any` remains in the unknown bucket
- [ ] `claude-opus-4-7[1m]` prices identically to `claude-opus-4-7`
- [ ] After this issue, no model ID present in the reference corpus sits in the unknown-pricing bucket except `any`
- [ ] Every seeded row carries a `source` naming the provider page and the date the rate was retrieved

**Out of Scope:**
- The catalog storage, resolver and effective-dating machinery (#186)
- Auto-fetching or scraping rates from provider pricing pages — entry stays manual
- Providers absent from the corpus beyond the Qwen line requested here (OpenAI, Gemini, DeepSeek, MiniMax) — add on demand
- Per-endpoint rate switching (International vs Mainland China) as a user setting; seed the International rate and record which one it is

## WHY

**Business Value:** Provider flexibility is a core reason to run Claude Code against non-Anthropic backends, and Agento's whole value proposition here is telling the user what their agent usage costs. A cost dashboard that silently reports $0 for the third-most-used model in the corpus is not doing that job.

**Problem Statement:** `pricingForModel` (`internal/claudesessions/analytics.go:34`) substring-matches on `opus`/`sonnet`/`haiku` and falls through to Sonnet, so Kimi K3's 8,181 assistant events are currently invented as Anthropic spend. #180 correctly stops that by routing unpriced models to an explicit unknown bucket — but the result swings the error the other way: a top-three model then contributes exactly $0.00. Neither number is defensible when the rates are published and public.

**Expected Impact:** ~8,200 assistant events (K3 8,181 + GLM 8) move from fabricated-or-zero to actually priced. The unknown-pricing bucket returns to being the rare exception it is designed to be — after this, only `any` (1 event) remains unresolved in the reference corpus. Users on mixed fleets get a true total for the first time.

**Context:**
- Blocked by: #186 — supplies the persisted, effective-dated catalog and the matcher these rows are entered into; without it there is nowhere to put them that survives a rate change
- Related: #180 — defines the unknown-pricing bucket this issue drains and the 5m/1h cache-write fields these rows must populate
- Decisions taken during refinement (2026-08-08): embedding models get a zero-cost non-billable row rather than a real rate or an unknown; `opus` uses the family fallback while `any` stays unknown; third-party cache-write is seeded at the provider's input rate with the assumption recorded rather than zeroed or flagged unknown.
- Verified against `main` @ `b72f09b`: model distribution re-counted from `~/.claude/projects` on 2026-08-08; `analytics.go` unchanged since #14.

## HOW

**Technical Approach:** This is predominantly data entry into the structure #186 builds, plus one small type change. The only code of substance is a `Billable` flag, because "priced at zero because it genuinely costs nothing" and "priced at zero because nobody filled it in" must not be the same state — conflating them is how a $0 row becomes an invisible bug.

**Implementation Details:**

### 1. `Billable` on `Rate`
```go
type Rate struct {
    // ...
    Billable bool // false = deliberately zero-cost (synthetic, embeddings)
}
```
A non-billable match returns `(rate, true)` so it does **not** land in the unknown bucket, but contributes `$0.00`. An unmatched model still returns `false`. Column `billable INTEGER NOT NULL DEFAULT 1` on `model_pricing` if #186 has not already added it.

### 2. Catalog entries — rates retrieved 2026-08-08, **re-verify before merging**
| Provider | `model_pattern` | Match | Input | Output | Cache read | Cache write |
|---|---|---|---|---|---|---|
| Moonshot | `k3` | exact | 3.00 | 15.00 | 0.30 | 3.00 † |
| Z.ai / Zhipu | `glm-5.2` | prefix | 1.40 | 4.40 | 0.26 | 1.40 † |
| Alibaba | `qwen3.8-max` | prefix | 2.00 | 6.00 | 0.25 | 2.00 † |
| Alibaba | `qwen3.5-397b` | prefix | 0.60 | 3.60 | — | — |
| Alibaba | `qwen3.5-plus` | prefix | 0.40 | 2.40 | — | — |
| Alibaba | `qwen3.5-flash` | prefix | 0.10 | 0.40 | — | — |
| Alibaba | `qwen3-max` | prefix | 0.78 | 3.90 | — | — |
| — | `mixedbread-ai/` | prefix | 0.00 | 0.00 | 0.00 | 0.00 (`billable: false`) |

† cache-write assumed equal to the input rate; record verbatim in `source`.

`match_type` is `exact` for `k3` deliberately — it is a two-character token and a prefix match would collide with any future model ID starting `k3`. Confirm the exact Qwen model-ID strings against Alibaba Cloud Model Studio at implementation time; the table above uses documented product names, and the corpus contains no Qwen usage yet to check against, so these are seeded ahead of need at the user's request.

Every row carries `"source": "<provider pricing page> · International endpoint · retrieved 2026-08-08"`. The endpoint matters: Alibaba's Mainland China endpoint is 60–70% cheaper and Z.ai's OpenRouter-hosted rate (~$0.95/$3.00) differs from direct.

### 3. Alias handling
`opus` (4 events) falls through to the coarse `contains "opus"` family fallback #180 defines, resolving to the current Opus-tier rate with the estimated flag set. `any` (1 event) matches nothing and stays unknown — it names no family, so any price would be fabricated. Neither needs a dedicated entry; both are assertions in the resolver test that the general rules produce the right outcome.

### Testing Strategy
- `internal/pricing/resolver_test.go` — one case per seeded provider asserting `(model_id, now)` → expected four-tuple; `mixedbread-ai/mxbai-embed-large-v1` → billable=false, cost 0, not unknown; `opus` → Opus fallback + estimated; `any` → not found; `claude-opus-4-7[1m]` → base Opus 4.7 rate
- `internal/pricing/catalog_test.go` — every embedded row parses, has a non-empty `source`, and has `billable=false` iff all four rates are zero (catches a half-filled entry)
- `internal/claudesessions/analytics_test.go` — a fixture session on `k3` produces non-zero cost and does not increment the unknown-pricing counter

- `docs/` — table of seeded providers with the endpoint each rate reflects and when it was retrieved, so the next refresh knows what it is replacing
- No new dependencies, env vars or config

**Files to Create:** none

**Files to Modify:** `internal/pricing/catalog.json`, `internal/pricing/{types,resolver}.go`, `internal/pricing/{resolver,catalog}_test.go`, `internal/storage/sqlite.go` (only if `billable` is not already present), `docs/`

**Risks:**
- *Published rates drift between refinement and merge* → re-verify every row at implementation time; `source` records the retrieval date so staleness is visible rather than assumed
- *Wrong endpoint seeded* → Alibaba Mainland is 60–70% cheaper and Z.ai differs via OpenRouter; the `source` string names the endpoint so a mismatch is diagnosable
- *Qwen model IDs guessed wrong* → no Qwen usage exists in the corpus to validate against; a wrong ID fails safe into the unknown bucket rather than mispricing, and #189's UI lets a user correct it without a release
- *Cache-write assumption overstates third-party cost* → documented in `source`; revisit if a provider publishes explicit cache-creation pricing

**Complexity:** Low

---

vibexp-agent:
Closes #187

## What

Prices the non-Anthropic models Claude Code is routinely pointed at, so third-party usage stops reporting `$0.00`.

Kimi K3 is the **third-most-used model in the reference corpus — 11,008 assistant events**, ahead of Haiku. #180 correctly stopped it being invented as Anthropic spend, but the result swung the error the other way: a top-three model contributed exactly nothing to the headline cost.

| Provider | Pattern | Match | Input | Output | Cache read |
|---|---|---|---|---|---|
| Moonshot | `k3` | exact | 3.00 | 15.00 | 0.30 |
| Moonshot | `kimi-k3` | prefix | 3.00 | 15.00 | 0.30 |
| Z.ai / Zhipu | `glm-5.2` | prefix | 1.40 | 4.40 | 0.26 |
| Alibaba | `qwen3.5-397b-a17b` | prefix | 0.60 | 3.60 | 0.06 |
| Alibaba | `qwen3.5-plus` | prefix | 0.40 | 2.40 | 0.04 |
| Alibaba | `qwen3.5-flash` | prefix | 0.10 | 0.40 | 0.01 |
| Alibaba | `qwen3-max` | prefix | 1.20 | 6.00 | 0.12 |
| — | `<synthetic>` | exact | non-billable | | |
| — | `mixedbread-ai/` | prefix | non-billable | | |

## Why the diff is bigger than "data entry"

The issue's HOW called this "predominantly data entry into the structure #186 builds, plus one small type change". Three things made that untrue, each verified against the code on `main` @ `413c2d8`:

1. **`catalog.json` had no cache fields.** `catalog.go` derived them as `input × 1.25 / 2 / 0.1` — Anthropic's TTL split and nobody else's. K3's $0.30 cached read coincidentally equals `3.00 × 0.1`, but GLM publishes $0.26 (derivation gives 0.14) and Alibaba prices caching as a **percentage rule**: explicit-cache creation 125% of input, hits 10%, with no TTL tiers at all. Cache rates are now overridable per rate, defaulting to the Anthropic multipliers.
2. **A zero rate was rejected outright**, so the non-billable embedding row could not be seeded. `Rate` gains `Billable`, and validation now enforces the invariant in both directions — a billable row must price every category, a non-billable one must price none. A half-filled entry fails the build instead of quietly under-reporting.
3. **The "coarse `contains "opus"` family fallback #180 defines" does not exist.** The resolver is anchored exact/prefix only, and the *exact* `opus` alias already resolved — but with `Estimated: false`, leaving "reported as estimated" unmet. Added an `estimated` catalog flag, OR'd into `Resolved.Estimated`.

`<synthetic>` was confirmed to be a hardcoded special case (`claudesessions/types.go:63`) and is now a catalog row, so the rule lives in one place.

## Rate re-verification (2026-08-09)

The HOW required re-verifying every rate before merge. **Three of seven rows did not survive it:**

- **Kimi K3's canonical API id is `kimi-k3`, not `k3`** — but Claude Code writes the bare `k3` into transcripts, so both are seeded. `k3` is an **exact** match on purpose: a two-character prefix would swallow any future id starting `k3`.
- **`qwen3-max` is $1.20/$6.00**, not the issue's $0.78/$3.90 — that figure is exactly 65% of list, a lapsed promotion.
- **Qwen3.8-Max is on no provider pricing page.** It is on Alibaba's marketplace with no published id and no published rate, two competing ids in circulation, and a claimed cached-input figure matching neither of Alibaba's documented cache percentages. Seeding it would be a guess, so it is omitted — a missing row is visible in the unknown bucket, a wrong one is not.
- Qwen ids corrected from product names to API ids (`qwen3.5-397b-a17b`, `qwen3.5-plus`, `qwen3.5-flash`).

Deferred to #206: Qwen3.8-Max once published, the newer 3.7 generation, and Alibaba's context-length tiers (which need a structural change to `Rate` — a rate would become a function of token count).

## Acceptance criteria

- [x] `k3` resolves to Moonshot rates and its 11,008 corpus events report non-zero cost
- [x] `glm-5.2` resolves to $1.40 / $4.40 / $0.26
- [x] Every Qwen entry resolves by its **published** model id and records the endpoint in `source` — Qwen3.8-Max omitted because it has none (#206)
- [x] `mixedbread-ai/mxbai-embed-large-v1` contributes $0.00, is non-billable, and is **not** in the unknown bucket
- [x] `opus` prices at the Opus-tier fallback and reports `Estimated`; `any` stays unknown
- [x] `claude-opus-4-7[1m]` prices identically to `claude-opus-4-7`
- [x] No corpus model id remains in the unknown bucket
- [x] Every seeded row carries a `source` naming the provider page, the endpoint, and the retrieval date

## Review round

One major finding, applied in `bf6559c`: `validateRate` guarded the built-in seed but **not `UpsertRate`** — the write path #189's settings UI will use — so it could store a billable row with no rates. The likelier accident is the inverse: `Billable`'s Go zero value is `false`, so the natural `Rate{InputPerMTok: …}` literal persists a *priced* row as non-billable. All four hand-built `Rate` fixtures in the test suite did exactly that, which is what the new validation surfaced.

## Migration

Migration **18** adds `billable INTEGER NOT NULL DEFAULT 1` and `estimated INTEGER NOT NULL DEFAULT 0`. Existing rows are real published Anthropic rates, so the defaults are correct for them. Both flags are in the FNV-1a revision hash, so toggling either re-costs affected sessions on the next `Cache.List`.

## Testing

`go build`, `go vet`, `golangci-lint` (v2.5.0, 0 issues), `go test -race ./...` — 18/18 packages green locally. New coverage: per-provider published-rate assertions, the billable-vs-unknown distinction, the alias/context-variant cases, catalog invariant enforcement (including proof it rejects a half-filled entry), cache-override-vs-derivation, and the `UpsertRate` validation rule.


vibexp-agent:
Deferred to #206 — the Qwen catalog gaps that could not be closed here: Qwen3.8-Max is on no provider pricing page, the 3.7 generation is unseeded, and Alibaba's context-length tiers need a structural change to `Rate`.

PR #207, opened 2026-08-09
feat(claude-sessions): show per-session cost in the session list and detail pages (#188)

### LINKED ISSUE #188: feat(claude-sessions): show per-session cost in the session list and detail pages
## Problem

Cost is only ever shown **in aggregate**. The Claude Sessions list and the session detail page show token counts but no money, so the one question a user actually asks — *"what did this session cost me?"* — has no answer anywhere in the UI.

Today:

- **`GET /api/claude-sessions`** (`internal/api/claude_sessions.go:17`) returns `ClaudeSessionSummary`, which has `usage` but no cost field (`internal/claudesessions/types.go:24`). `SessionRow` (`frontend/src/pages/ClaudeSessionsPage.tsx:348`) renders `12K↑ 16K↓` and stops there.
- **`GET /api/claude-sessions/{id}`** returns `ClaudeSessionDetail` — same story.
- The only per-session cost that exists is `SessionInsight.CostEstimateUSD` (`internal/claudesessions/processor.go:66`), produced by a *different* pipeline (the insight processors), stored in a *different* table (`session_insights`), gated behind the Insights page, and computed differently from the analytics page — `TokenProfileProcessor` (`token_processor.go:37`) prices the whole session with the single first-seen model, while `buildSummary` (`analytics.go:244`) prices per session from aggregate totals.

So there are already two cost code paths that can disagree, and neither is reachable from the session list.

## Impact

- A user browsing 740 sessions cannot sort, scan or spot the expensive ones.
- Drill-down from analytics lands on a session list with no cost, breaking the "which sessions made up this spike?" flow the heatmap/hourly drill-down was built for (`frontend/src/lib/drilldown.ts`).
- Two independent cost computations for the same session is a correctness hazard: whichever one a user sees first anchors them.

## Proposed direction

Make per-session cost a **first-class, stored property of the session**, computed once by the scanner and reused everywhere.

- Persist the cost breakdown (input / output / cache read / cache write, and total) on `claude_session_cache` alongside the token counts, computed during `readSessionSummary` (`scanner.go:542`) as messages are read.
- Expose it on `ClaudeSessionSummary` so both the list and detail endpoints carry it with no extra query.
- Render it on `SessionRow` next to the token badge, with a breakdown on hover matching the existing token tooltip, and prominently on the session detail page.
- Have the analytics page and the insight pipeline consume this stored value instead of each re-deriving it, collapsing the two divergent code paths into one.
- Surface honestly when a session contains models with no known rate (the "unknown pricing" bucket from #180) — a session that is partly unpriced should say so rather than presenting an understated number as complete.

Worth deciding during refinement: whether cost becomes a **sort option** on the session list, and whether the list shows a total across the current filter. Both are natural follow-ons but neither is required for the core ask.

**Blocked by #186** (time-versioned pricing catalog). Storing a cost per session only makes sense once pricing is effective-dated and per-message — otherwise the stored number silently becomes wrong the next time rates change, which is worse than computing it on the fly. #186 also introduces the pricing-revision marker that tells the scanner when a stored cost needs recomputing.

Related: #180 (explicit pricing lookup and unknown-model bucket), #187 (third-party rates, so third-party sessions do not display $0).

## Out of scope

- Cost on the Chats / Tasks / Job History pages (those have their own token accounting in `chat_sessions` and `job_history`).
- Budgets, alerting, or spend forecasting.

---
## WHAT

**Summary:** Store a per-session cost breakdown on the session cache and surface it on the Claude Sessions list rows and the session detail page, replacing the two divergent on-the-fly cost computations with one stored value.

**Scope:**
- Persist the four-part cost breakdown (input / output / cache read / cache write) plus total on `claude_session_cache`, computed during the scan
- Carry it on `ClaudeSessionSummary` so both list and detail endpoints return it with no extra query
- Render cost on `SessionRow` beside the token badge, with a hover breakdown matching the existing token tooltip
- Render cost in the existing token-usage banner on the detail page
- Flag a session whose models are not fully priced, rather than presenting an understated total as complete
- Point `AggregateAnalytics` and the insight pipeline at the stored value so the two cost paths collapse into one

**Affected Components:**
- `internal/claudesessions/types.go` — `ClaudeSessionSummary` (`:24`) gains a `Cost` struct and an unpriced-models indicator
- `internal/claudesessions/scanner.go` — `readSessionSummary` (`:542`) accumulates cost; `upsertCacheRow` (`:420`) and `loadAllSessions` (`:465`) persist and read it
- `internal/claudesessions/cache.go` — `loadAll` (`:130`) runs the same SELECT as `loadAllSessions` and must stay in sync
- `internal/storage/sqlite.go` — migration adding the cost columns
- `frontend/src/types.ts` — `ClaudeSessionSummary` (`:527`) mirrors the Go struct
- `frontend/src/pages/ClaudeSessionsPage.tsx` — `SessionRow` (`:348`), cost badge next to the existing `Zap` token badge (`:398`–`:425`)
- `frontend/src/pages/ClaudeSessionDetailPage.tsx` — token usage banner (`:487`–`:518`)
- `frontend/src/lib/format.ts` — a `formatCost` helper alongside `formatTokens` (`:6`)

**Acceptance Criteria:**
- [ ] `GET /api/claude-sessions` returns a per-session cost breakdown and total; `GET /api/claude-sessions/{id}` returns the same values for the same session
- [ ] Every row in the session list shows a cost, and hovering reveals the input/output/cache-read/cache-write split — matching the interaction of the existing token tooltip
- [ ] The detail page's usage banner shows total cost alongside the existing token figures
- [ ] A session containing a model with no known rate is visibly marked as partially priced; its displayed total is not presented as complete
- [ ] Sub-cent costs render meaningfully rather than as `$0.00` (a 3K-token session is not free)
- [ ] The list total shown for a given session equals the analytics page's contribution for that same session — one number, one source
- [ ] Cost columns populate for existing cached rows on upgrade without a manual cache wipe

**Out of Scope:**
- Cost on the Chats, Tasks or Job History pages — those have separate token accounting in `chat_sessions` and `job_history`
- Budgets, spend alerting or forecasting
- Sorting the session list by cost, and a filter-wide cost total — both natural follow-ons, neither required here (see Risks)
- Per-message cost in the transcript

## WHY

**Business Value:** "What did this session cost me?" is the single most direct question a user brings to a session list, and today the UI cannot answer it anywhere. Cost exists only in aggregate on the analytics page, so a user who spots a spend spike has no way to walk from the spike to the sessions that caused it — which is precisely the flow the heatmap and hourly drill-down (`frontend/src/lib/drilldown.ts`, shipped in #173) was built to serve.

**Problem Statement:** `ClaudeSessionSummary` (`internal/claudesessions/types.go:24`) carries `usage` but no cost, so `SessionRow` (`ClaudeSessionsPage.tsx:348`) renders token counts and stops. The only per-session cost that exists is `SessionInsight.CostEstimateUSD` (`processor.go:66`) — produced by a different pipeline, stored in a different table (`session_insights`), reachable only from the Insights page, and computed differently from the analytics page: `TokenProfileProcessor.Finalize` (`token_processor.go:37`) prices a whole session at its first-seen model, while `buildSummary` (`analytics.go:244`) prices from aggregate session totals. Two independent implementations of the same number is a correctness hazard in itself — whichever a user sees first anchors them, and nothing forces them to agree.

**Expected Impact:** Cost becomes visible at the grain users actually browse — 740 sessions in the reference corpus, currently all unlabelled. Analytics drill-down becomes end-to-end. The two cost code paths collapse to one, removing a class of "why do these two screens disagree" bug before it is reported.

**Context:**
- Blocked by: #186 — storing a per-session cost is only safe once pricing is effective-dated and accumulated per message; against a mutable global rate table a stored number silently rots, which is strictly worse than recomputing. #186 also supplies the pricing-revision marker that tells the scanner when a stored cost needs recomputing.
- Related: #180 — supplies the unknown-pricing bucket behind the partially-priced indicator
- Related: #187 — without third-party rates, K3 and GLM sessions would display a confident `$0.00`, which is worse than displaying nothing
- Verified against `main` @ `b72f09b` on 2026-08-08: `ClaudeSessionSummary` has no cost field, `SessionRow` renders tokens only, and the detail page's usage banner (`:487`) has room for cost without a layout change.

## HOW

**Technical Approach:** Cost becomes a stored property of the session rather than a derived view. The scanner already walks every assistant event to sum tokens (`processSummaryEvent`, `scanner.go:585`); #186 has it resolve a rate per event, so the breakdown falls out of the same pass at no extra I/O. Everything downstream then reads one persisted number instead of re-deriving it three times.

**Implementation Details:**

### 1. Types and storage
```go
type SessionCost struct {
    InputUSD      float64 `json:"input_usd"`
    OutputUSD     float64 `json:"output_usd"`
    CacheReadUSD  float64 `json:"cache_read_usd"`
    CacheWriteUSD float64 `json:"cache_write_usd"`
    TotalUSD      float64 `json:"total_usd"`
}

type ClaudeSessionSummary struct {
    // ...
    Cost           SessionCost `json:"cost"`
    UnpricedModels []string    `json:"unpriced_models,omitempty"`
}
```
`UnpricedModels` is the honest-reporting mechanism: a non-empty list means the total is a floor, not a figure. Migration adds five `REAL` columns plus a TEXT column for the unpriced list; existing rows re-cost via #186's `pricing_rev` mechanism, so no manual wipe.

`loadAllSessions` (`scanner.go:465`) and `Cache.loadAll` (`cache.go:130`) run the same SELECT — both column lists must be extended together, and this pair has drifted before.

### 2. Frontend — list
Add `formatCost` to `frontend/src/lib/format.ts` next to `formatTokens` (`:6`). It must not round sub-cent values to `$0.00`: below one cent, show more precision (e.g. `$0.004`) so short sessions read as cheap rather than free. Mirror the existing token-badge pattern in `SessionRow` (`:398`–`:425`) — a `Tooltip` wrapping a compact badge, breakdown rows inside — so the cost badge is visually a sibling of the token badge rather than a new idiom. When `unpriced_models` is non-empty, mark the badge (a `~` prefix or a muted warning icon) and name the offending models in the tooltip.

### 3. Frontend — detail
The usage banner (`ClaudeSessionDetailPage.tsx:487`–`:518`) already lays out input / output / cache-write / cache-read as labelled figures; add total cost as a further item, and the same partially-priced marker.

### 4. Collapse the duplicate cost paths
`buildSummary` (`analytics.go:227`) and `buildCostOverTime` (`:419`) sum `s.Cost` rather than re-pricing from tokens; `TokenProfileProcessor.Finalize` (`token_processor.go:37`) reads the stored value instead of computing its own. Much of this lands as part of #186 — this issue's job is to confirm no caller re-derives cost independently once the stored field exists.

### Testing Strategy
- `internal/claudesessions/scanner_test.go` — fixture session asserts the persisted breakdown sums to the total and that each component matches the hand-computed per-model figure; a fixture with an unknown model asserts `UnpricedModels` is populated and the total excludes it
- `internal/claudesessions/cache_test.go` — round-trip through `upsertCacheRow` → `loadAll` preserves every cost column (guards the two-SELECT drift)
- `internal/api/api_test.go` — list and detail responses expose identical cost for the same session
- `frontend` — unit test for `formatCost` covering sub-cent, cent and dollar magnitudes
- Cross-check: the analytics total for a date range equals the sum of the per-session costs the list shows for that range

- `CLAUDE.md` — note that per-session cost is stored on the session cache and is the single source for all cost display
- No new dependencies, env vars or config

**Files to Create:** none

**Files to Modify:** `internal/claudesessions/{types,scanner,cache,analytics,token_processor}.go`, `internal/storage/sqlite.go`, `frontend/src/types.ts`, `frontend/src/lib/format.ts`, `frontend/src/pages/{ClaudeSessionsPage,ClaudeSessionDetailPage}.tsx`, `CLAUDE.md`

**Risks:**
- *Displaying a confident total for a partly-unpriced session* → the `unpriced_models` marker is an acceptance criterion, not a nicety; a silently understated cost is the failure mode this issue must not ship
- *Row density on the session list* → the list already carries project, branch, relative time, message count and a token badge; if cost makes the meta row wrap on narrow viewports, prefer replacing the token badge with a combined badge over adding a sixth element
- *Users immediately want sort-by-cost* → deliberately out of scope to keep this shippable; the stored column makes it a small follow-up rather than a rework
- *`loadAllSessions` / `Cache.loadAll` column drift* → covered by the round-trip test above

**Complexity:** Medium

---

vibexp-agent:
Closes #188

## What

Per-session cost becomes a **stored property of the session**, computed once by the scan and read everywhere. The Claude Sessions list and the detail page now answer the question the list invites — *what did this session cost me?* — which previously had no answer anywhere in the UI.

- Cost badge on every session row, mirroring the token badge beside it, with the input / output / cache-read / cache-write split on hover.
- Total cost in the detail page's usage banner.
- A session containing a model with no published rate is marked `~` and names the offending models: its total is a floor, not a figure.
- `buildSummary` and `buildCostOverTime` read the stored value, so the list and the analytics totals are the same number by construction.

## Why the diff reaches past the UI

Two things the HOW assumed turned out not to hold on `main` @ `2228e41`:

**The `pricing_rev` re-cost mechanism did not exist.** The HOW says existing rows "re-cost via #186's `pricing_rev` mechanism" — but no pricing revision was stored per row or anywhere else. It lived only in an in-memory struct, and a rate edit needed no re-scan *precisely because* cost was computed at read time (`cache.go`: *"Costs are computed at read time, so the summaries loaded below already reflect the new rates"*).

Storing cost invalidates that guarantee, so the mechanism is built here: `claude_cache_metadata.pricing_rev` records the catalog fingerprint the costs were computed under, a drift invalidates the cached mtimes to force a re-read, and `Cache.List` skips its freshness short-circuit — otherwise cached rows would serve stale costs indefinitely. Re-reading is the only correct response: re-pricing needs each message's own model and timestamp, and the row keeps neither. This matters immediately, because #189 is a rate-editing UI.

**Sub-agent cost had to be stored too.** Analytics prices `TotalUsage()` (main + delegated), so a main-transcript-only cost would have silently dropped every sub-agent's spend from the headline, failing the "list total equals the analytics contribution" criterion. `Cost` / `SubagentCost` / `TotalCost()` now mirror `Usage` / `SubagentUsage` / `TotalUsage()` exactly.

Conversely, one HOW risk is already mitigated: `loadAllSessions` and `Cache.loadAll` no longer duplicate a SELECT — both wrap `querySessionSummaries` over one shared const, so the drift the issue warns about cannot recur.

## Collapsing the duplicate cost paths

The issue's real target. `TokenProfileProcessor` already priced per message (#186 fixed that), so the remaining divergence was `buildSummary`/`buildCostOverTime` pricing *aggregate* tokens at the *first-seen* model — an approximation that could not agree with the scanner's per-message figure. Both now read the stored value, which left `costForUsage`/`costFromPricing` with no production caller; they moved to the test file, which is what still exercises the resolver wiring.

`formatCost` already existed, private to `TokenUsagePage`. Lifted to `lib/format.ts` rather than written a second time — a second implementation would recreate exactly the duplication this issue exists to remove. Its smallest bucket was also wrong for per-session amounts: anything under `$0.0001` rounded to `$0.00`, so a real cost read as free. Now only an exact zero — which a session of purely non-billable models genuinely has — prints `$0.00`.

## Acceptance criteria

- [x] List and detail return the same cost for the same session — `ClaudeSessionDetail` embeds `ClaudeSessionSummary`, so they are the same fields by construction
- [x] Every row shows a cost; hovering reveals the four-way split
- [x] Detail banner shows total cost alongside the token figures
- [x] A partly-priced session is visibly marked and names the unpriced models
- [x] Sub-cent costs render meaningfully — `$0.0150`, `< $0.0001`, never a rounded `$0.00`
- [x] The list total equals the analytics contribution — one stored value, read by both
- [x] Cost populates for existing cached rows on upgrade with no manual wipe (scanner version 6)

## Review round

One finding, applied in `9bf2694`: `unpriced_models` and `unpriced_tokens` are written together on both cache tables but the sub-agent roll-up summed only the tokens. A session whose *sub-agent* used an unpriced model came back with `UnpricedTokens > 0` and an empty `UnpricedModels` — and since both badges key their disclosure off the model list, the row rendered a confident total while analytics reported excluded tokens belonging to no model. The original test missed it because its unpriced model sat on the parent transcript.

Deferred to #208: the pricing-triggered re-cost runs inline in `Cache.List`, so a rate saved from #189's UI would pay a full 1,556-transcript re-read in one request.

## Migration

Migration **19** adds the five cost columns plus `unpriced_models`/`unpriced_tokens` to both `claude_session_cache` and `claude_subagent_cache`, and `pricing_rev` to `claude_cache_metadata`. `CurrentScannerVersion` goes to **6** so existing rows — zeros in every cost column, indistinguishable from a genuinely free session — are re-read rather than left alone.

## Testing

`go build`, `go vet`, `golangci-lint` (v2.5.0, 0 issues), `go test -race ./...` — 18/18 packages green. Frontend: typecheck, ESLint, Prettier, build all clean; new `src/lib/format.test.ts` (8 cases) passes.

New backend coverage: the persisted breakdown round-trips through the shared SELECT (catching a column added to the INSERT but forgotten in the scan), multi-model per-message pricing, unpriced-model recording, non-billable-is-not-a-gap, sub-agent roll-up, delegated unpriced models, and both pricing-revision paths — drift forces a re-cost, unknown revision never does.

> **Unrelated pre-existing issue noticed:** `src/lib/streamingBlocks.test.ts` has 7 failing tests on `main`. CI's frontend job runs typecheck/lint/format/build but **not** `npm test`, which is how they went unnoticed. Untouched by this PR.


vibexp-agent:
Deferred to #208 — the pricing-triggered re-cost runs inline in `Cache.List`, so saving a rate from #189's UI would pay a full 1,556-transcript re-read inside one request. Consistent with the existing scanner-version path, but that fires once per release rather than on a routine UI action.

PR #212, opened 2026-08-09
fix(analytics): time buckets and date ranges are UTC — show everything in the browser timezone (#190)

### LINKED ISSUE #190: fix(analytics): time buckets and date ranges are UTC — show everything in the browser timezone
## Problem

Analytics are computed and labelled in **UTC**, but users read them in their own timezone. Storage in UTC is correct and should stay; presenting UTC to the user is not.

Session timestamps are parsed from JSONL `Z` values and never converted, so every bucketing helper in `internal/claudesessions/analytics.go` operates in UTC:

- `bucketKey` / `bucketLabel` (`analytics.go:452`) — `t.Format("2006-01-02T15")` on a UTC time
- `buildHeatmap` (`analytics.go:380`) — `s.LastActivity.Weekday()` / `.Hour()`
- `buildHourlyActivity` (`analytics.go:404`)
- `buildMostActiveDays` (`analytics.go:359`)
- `filterSessions` (`analytics.go:213`) against `from`/`to` parsed as UTC midnights (`internal/api/claude_analytics.go:33`, `parseAnalyticsDate` `:49`); the insights endpoint parses dates the same way (`internal/api/claude_session_insights.go:61,69`)

The UI then leans into it rather than correcting it — `GeneralUsagePage.tsx:326,392` literally tell the user "(UTC)", and `lib/drilldown.ts` builds its windows with `getUTCDay()` / `getUTCHours()` and labels them `"Mondays 14:00–15:00 UTC"`.

**Concretely, at `Europe/Berlin` (UTC+2, the machine this was investigated on):**

- Work done at 00:30 local on 8 Aug is attributed to **7 Aug** in "Most active days" — a day boundary off by one for every late-night session.
- The activity heatmap and hourly chart are shifted **two hours**, so a 9-to-5 working day renders as 7am-to-3pm and peak-hours conclusions are simply wrong.
- The "This month" preset can include or exclude a day at each edge.

There is a second, subtler bug in the same area: `fmt()` (`frontend/src/pages/analyticsShared.tsx:41`) derives the range with `d.toISOString().slice(0, 10)` — a **UTC** date extracted from a **local** `Date`. `presetToRange` builds "this month" / "last month" from *local* calendar arithmetic (`new Date(y, m, 1)`) and then serialises those local dates through a UTC formatter, so the two halves disagree at the boundary. Users west of UTC get an off-by-one day on every preset.

`formatDateLabel` (`analyticsShared.tsx:93`) compounds it by taking a UTC bucket key and rendering it through `new Date(d + 'T00:00:00')`, which parses as *local* — a UTC-labelled bucket displayed with a local-parsed date.

## Impact

- Every time-of-day and day-of-week insight is wrong by the user's UTC offset — and these charts exist specifically to answer "when do I work?".
- Drill-down from a heatmap cell selects the wrong sessions, because the windows it generates are UTC hours.
- Date-range presets are off by a day at the edges.
- Showing "(UTC)" in the UI is the current mitigation; it is a workaround for the bug, not a fix, and users still have to do the arithmetic in their heads.

Non-analytics timestamps are mostly fine already — `formatDateTime` / `formatRelativeTime` (`frontend/src/lib/utils.ts`), `SessionJourneyPage.tsx:31`, `JobHistoriesPage.tsx`, `TasksPage.tsx` and `TaskForm.tsx` all use `toLocale*` and render in the browser zone. The defect is concentrated in the analytics/bucketing path.

## Proposed direction

Keep **everything** in UTC in storage, transport and the backend. Convert only at the aggregation and presentation boundary, using the **browser's** IANA timezone.

- Frontend sends its zone (`Intl.DateTimeFormat().resolvedOptions().timeZone`) as a `tz` query param on the analytics and insights endpoints.
- Backend resolves it with `time.LoadLocation` and applies `.In(loc)` before bucketing — day, hour and weekday keys, plus the `from`/`to` day boundaries. Invalid or missing `tz` falls back to UTC so existing callers keep working.
- Because the binary ships standalone, embed the zone database (`import _ "time/tzdata"`) rather than depending on the host having `/usr/share/zoneinfo` — nothing currently imports it.
- `lib/drilldown.ts` switches from `getUTCDay()`/`getUTCHours()` to local equivalents, and the `"… UTC"` labels and the "(UTC)" copy in `GeneralUsagePage.tsx` come out.
- `fmt()` in `analyticsShared.tsx` builds a `YYYY-MM-DD` from local date parts so it agrees with the local arithmetic in `presetToRange`.
- `formatDateLabel` parses bucket keys consistently with how the backend now emits them.

Worth settling during refinement: whether the timezone should be a **user setting** (with "browser default" as the default) rather than purely implicit, since a user analysing a team's data or travelling may want to pin a zone. Also how DST transitions are handled at day/hour bucket boundaries — a 23-hour and a 25-hour day exist and the heatmap has a fixed 24-column grid.

Independent of the pricing chain (#180, #186, #187, #188, #189) — it touches the same `analytics.go` file, so whichever lands second should expect a small rebase, but there is no functional dependency.

## Out of scope

- Changing anything about how timestamps are stored or transported. UTC end to end below the presentation layer.
- Locale/formatting concerns beyond timezone (date format, number format, translations).

---
## WHAT

**Summary:** Bucket and label analytics in the browser's IANA timezone instead of UTC, keeping storage, transport and the backend in UTC, so day/hour/weekday charts and date-range presets reflect the user's actual working day.

**Scope:**
- `tz` query param on the analytics and insights endpoints, applied before bucketing
- Day, hour and weekday keys plus `from`/`to` day boundaries resolved in that zone
- Embedded tzdata so `time.LoadLocation` works in the standalone binary
- Drill-down windows generated in local time; `"… UTC"` labels and "(UTC)" copy removed
- `fmt()` derives its `YYYY-MM-DD` from local date parts so it agrees with `presetToRange`'s local arithmetic
- `formatDateLabel` parses bucket keys consistently with how the backend emits them
- Missing or invalid `tz` falls back to UTC

**Affected Components:**
- `internal/claudesessions/analytics.go` — `AnalyticsParams` gains a location; `bucketKey`/`bucketLabel` (`:452`), `buildHeatmap` (`:380`), `buildHourlyActivity` (`:404`), `buildMostActiveDays` (`:359`), `filterSessions` (`:213`), `fillTimeSeries` (`:463`), `buildCostOverTime` (`:419`)
- `internal/api/claude_analytics.go` — parse `tz`; `parseAnalyticsDate` (`:49`) and the end-of-day construction (`:33`) resolve in that location
- `internal/api/claude_session_insights.go` — same date parsing (`:61`, `:69`)
- `cmd/` or `main.go` — `import _ "time/tzdata"`
- `frontend/src/lib/api.ts` — send `Intl.DateTimeFormat().resolvedOptions().timeZone`
- `frontend/src/lib/drilldown.ts` — `collectWindows` (`:67`), `heatmapCellTarget` (`:86`), `hourlyBarTarget` (`:102`), `parseRangeBounds` (`:47`) switch to local; labels drop "UTC"
- `frontend/src/pages/analyticsShared.tsx` — `fmt` (`:41`), `formatDateLabel` (`:93`)
- `frontend/src/pages/GeneralUsagePage.tsx` — "(UTC)" copy (`:326`, `:392`)

**Acceptance Criteria:**
- [ ] At `Europe/Berlin` (UTC+2), a session whose last activity is 2026-08-08T00:30 local is attributed to **8 Aug** in "Most active days", not 7 Aug
- [ ] The activity heatmap and hourly chart place that session in the **00:00–01:00** cell, not 22:00–23:00
- [ ] `from`/`to` are inclusive of full local days — a "This month" preset at UTC−5 includes the whole local 1st and the whole local last day
- [ ] `fmt()` and `presetToRange` agree at both range edges in a zone west of UTC (currently off by one day)
- [ ] Clicking a heatmap cell selects the sessions in that **local** weekday+hour window, and the drill-down label carries no "UTC"
- [ ] A request with no `tz`, or an unparseable one, returns UTC-bucketed data rather than erroring — existing callers keep working
- [ ] `time.LoadLocation("Europe/Berlin")` resolves in a binary built and run in a container with no system zoneinfo
- [ ] All stored and transported timestamps remain UTC — no storage or API-shape change

**Out of Scope:**
- Changing how timestamps are stored or transported; UTC end to end below the presentation layer
- A user-selectable timezone override — browser zone is the requirement (see Context)
- Locale concerns beyond timezone: date format, number format, translations
- Non-analytics timestamps, which already render locally via `toLocale*`

## WHY

**Business Value:** The heatmap, hourly chart and "most active days" exist to answer one question — *when do I work?* — and they answer it in a timezone the user does not live in. For anyone not at UTC±0 the charts are wrong by their offset, which makes the flagship insight feature quietly untrustworthy.

**Problem Statement:** Session timestamps are parsed from JSONL `Z` values and never converted, so every bucketing helper in `internal/claudesessions/analytics.go` works in UTC — `bucketKey` (`:452`), `buildHeatmap` (`:380`), `buildHourlyActivity` (`:404`), `buildMostActiveDays` (`:359`) — filtered against `from`/`to` parsed as UTC midnights (`internal/api/claude_analytics.go:33`, `:49`). The UI accommodates the bug rather than fixing it: `GeneralUsagePage.tsx:326,392` tell the user "(UTC)", and `lib/drilldown.ts` builds windows with `getUTCDay()`/`getUTCHours()` and labels them `"Mondays 14:00–15:00 UTC"`.

Measured at `Europe/Berlin` (UTC+2, the machine this was investigated on): work at 00:30 local is attributed to the **previous day**, and the heatmap is shifted **two hours**, so a 9-to-5 day renders as 7am-to-3pm.

A second defect compounds it. `fmt()` (`analyticsShared.tsx:41`) takes `d.toISOString().slice(0, 10)` — a **UTC** date extracted from a **local** `Date` — while `presetToRange` (`:50`) builds "this month"/"last month" from **local** calendar arithmetic (`new Date(y, m, 1)`). The two halves disagree at the boundary, so users west of UTC get an off-by-one day on every preset. `formatDateLabel` (`:93`) then renders a UTC bucket key through `new Date(d + 'T00:00:00')`, which parses as local — a UTC bucket displayed with a local-parsed date.

**Expected Impact:** Every time-of-day and day-of-week insight becomes correct for the user's actual zone. Drill-down selects the right sessions. Date presets stop being off by a day at the edges. The "(UTC)" disclaimers — a workaround, not a fix — come out.

**Context:**
- Requirement stated directly by the user (2026-08-08): *"in the backend we can use UTC timezone... but in the frontend it should always use the browser timezone"* and *"all the storage and backend should store things in UTC"*. A user-selectable override was therefore **not** pursued; browser zone is the specification, not a default.
- Related: #190 touches `internal/claudesessions/analytics.go`, as do #180 and #186. No functional dependency in either direction — whichever lands second takes a small rebase in the bucketing helpers.
- Verified against `main` @ `b72f09b` on 2026-08-08: no `time/tzdata` import exists anywhere in the tree; `analytics.go` is unchanged since #14; non-analytics pages (`SessionJourneyPage.tsx:31`, `JobHistoriesPage.tsx`, `TasksPage.tsx`, `TaskForm.tsx`, `lib/utils.ts`) already use `toLocale*` and need no change.

## HOW

**Technical Approach:** Convert at the aggregation boundary only. The frontend tells the backend its zone; the backend resolves it once per request and applies `.In(loc)` before deriving any bucket key. Nothing about storage or transport changes — timestamps stay UTC on the wire, which keeps the API honest and the change reviewable.

**Implementation Details:**

### 1. Carry the location
```go
type AnalyticsParams struct {
    From, To time.Time
    Project  string
    Loc      *time.Location // nil → time.UTC
}
```
`handleGetClaudeAnalytics` (`internal/api/claude_analytics.go:18`) reads `tz`, resolves via `time.LoadLocation`, and falls back to `time.UTC` on empty or error — never a 400, so existing callers and bookmarked URLs keep working. Same treatment for the two date parses in `claude_session_insights.go` (`:61`, `:69`).

### 2. Bucket in that location
Every helper that derives a calendar field from `s.LastActivity` takes the location:
```go
func bucketKey(t time.Time, granularity string, loc *time.Location) string {
    t = t.In(loc)
    // ...unchanged
}
```
Applies to `bucketKey`/`bucketLabel` (`:452`), `buildHeatmap` (`:380`, `.Weekday()`/`.Hour()`), `buildHourlyActivity` (`:404`), `buildMostActiveDays` (`:359`, `Format("2006-01-02")`), and the `fillTimeSeries`/`buildCostOverTime` walk loops (`:463`, `:419`) which step by 24h from `from`.

`from`/`to` become local day boundaries: parse `YYYY-MM-DD` with `time.ParseInLocation` and build the inclusive end as local 23:59:59, replacing the current `time.Date(..., t.Location())` at `:33` which inherits UTC from `time.Parse`.

### 3. Embed tzdata
`import _ "time/tzdata"` in `main.go`. Nothing imports it today, so `time.LoadLocation` currently depends on the host's `/usr/share/zoneinfo` — fine on a dev Mac or Linux box, absent in a scratch/distroless container. Costs ~450KB in the binary, which is the right trade for a tool distributed as a single self-contained executable.

### 4. Frontend
`lib/api.ts` appends `tz=${Intl.DateTimeFormat().resolvedOptions().timeZone}` to the analytics and insights calls.

`lib/drilldown.ts`: `collectWindows` (`:67`) steps with `setHours` instead of `setUTCHours` (`:80`); `heatmapCellTarget` (`:86`) and `hourlyBarTarget` (`:102`) match on `getDay()`/`getHours()`; `parseRangeBounds` (`:47`) builds bounds from local midnight rather than `T00:00:00Z`; the `"UTC"` suffixes come out of both labels (`:96`, `:111`), as does the module doc-comment (`:13`–`:22`) which currently documents the UTC behaviour as intentional. Its test file `drilldown.test.ts` asserts UTC throughout and must be rewritten with a pinned `TZ` so the expectations are deterministic in CI.

`analyticsShared.tsx`: `fmt` (`:41`) builds from `getFullYear`/`getMonth`/`getDate` so it agrees with `presetToRange`'s local arithmetic; `formatDateLabel` (`:93`) parses bucket keys consistently with the backend's new output.

`GeneralUsagePage.tsx`: drop "(UTC)" from `:326` and `:392`.

### 5. DST
A local day can be 23 or 25 hours while the heatmap grid is fixed at 24 columns. Bucket by local wall-clock hour and accept that a spring-forward day has one empty cell and a fall-back day has one doubled — correct for a "when do I work?" chart and simpler than any alternative. The `fillTimeSeries`/`buildCostOverTime` loops step by a fixed 24h from `from`, which drifts across a DST boundary; step by calendar day in `loc` instead.

### Testing Strategy
- `internal/claudesessions/analytics_test.go` — table test with a fixed `Europe/Berlin` location: a 2026-08-08T00:30+02:00 session lands in the 8 Aug day bucket and the hour-0 heatmap cell; the same input under `time.UTC` lands in 7 Aug / hour 22 (pins both behaviours); a DST-transition day produces the expected 23- and 25-hour day handling
- `internal/api/` — `tz` absent, empty and garbage all yield UTC bucketing without error; a valid `tz` changes the bucketing
- `frontend/src/lib/drilldown.test.ts` — rewritten against local time with `TZ` pinned (e.g. `process.env.TZ = 'Europe/Berlin'`) so CI is deterministic regardless of runner zone
- `frontend` — `fmt()` and `presetToRange` agree at month boundaries in a zone west of UTC (the current off-by-one)
- Manual: `make build && ./agento web`, compare the heatmap against known working hours

- `docs/` — one line: storage and API are UTC; presentation and bucketing follow the browser
- No new dependencies (tzdata is stdlib), no env vars, no config

**Files to Create:** none

**Files to Modify:** `internal/claudesessions/analytics.go`, `internal/api/{claude_analytics,claude_session_insights}.go`, `main.go`, `frontend/src/lib/{api,drilldown}.ts`, `frontend/src/lib/drilldown.test.ts`, `frontend/src/pages/{analyticsShared.tsx,GeneralUsagePage.tsx}`, `internal/claudesessions/analytics_test.go`, `docs/`

**Risks:**
- *Every existing chart shifts on upgrade and looks like a regression* → it is the fix, but call it out in release notes; the "(UTC)" copy disappearing is the visible signal
- *`drilldown.test.ts` currently encodes UTC as correct* → it must be rewritten, not patched; pin `TZ` so a CI runner in a different zone does not produce flaky expectations
- *DST edge cases in fixed-24h stepping* → step by calendar day in the target location rather than adding 24h
- *Rebase against #180/#186 in `analytics.go`* → those change the cost loops, this changes the bucket keys; overlap is small but real
- *Binary size +~450KB from tzdata* → acceptable for a self-contained distribution and the only reliable option in minimal containers

**Complexity:** Medium

---

vibexp-agent:
Closes #190

## What

The heatmap, hourly chart and "most active days" exist to answer one question — *when do I work?* — and they answered it in a timezone the user does not live in. Every bucketing helper derived its calendar fields from a UTC instant, so for anyone off UTC±0 the charts were wrong by their offset. At `Europe/Berlin` a 9-to-5 day rendered as 7am-to-3pm, and work at 00:30 was filed under the previous day. The UI accommodated this by printing "(UTC)" rather than fixing it.

**Storage and transport stay UTC end to end.** Only aggregation and display move: the frontend sends its IANA zone as `tz`, and the backend applies `.In(loc)` before deriving any bucket key. A bare `YYYY-MM-DD` `from`/`to` becomes a local day boundary via `ParseInLocation`, which is what makes "this month" cover the user's month.

A missing or unrecognized `tz` falls back to UTC rather than erroring — analytics is a read-only dashboard, and refusing to render it over a bad zone string is worse than rendering what every caller got before the parameter existed.

## Three things beyond a literal `.In(loc)` sweep

1. **The two byte-identical stepping loops became one `walkBuckets`, and daily steps advance the calendar day instead of adding 24 hours.** A local day is 23 or 25 hours across a DST transition, so a fixed step drifts off the wall clock, emitting one day key twice and skipping another.
2. **`collectWindows` steps by a real hour** rather than `setHours(getHours() + 1)`, which can jump two hours in spring or stall in autumn. Walking absolute time visits each elapsed hour exactly once and reads the local fields off it — a repeated 02:00 correctly yields two windows, a skipped one yields none.
3. **tzdata is embedded** (`_ "time/tzdata"`). Without it `LoadLocation` depends on the host's `/usr/share/zoneinfo`, absent in a distroless image — every lookup would fail and silently degrade the whole dashboard back to UTC, which is the bug being fixed. ~450KB, the right trade for a self-contained binary. It also fixes the `current_time` MCP tool, which has the same dependency.

## A second, independent defect

`fmt()` serialised a locally-built `Date` through `toISOString`, so it disagreed with `presetToRange`'s local calendar arithmetic (`new Date(y, m, 1)`) and shifted every range edge by a day. Making it local **incidentally fixes** `InsightsPage.previousPeriod` too, which parsed `from + 'T00:00:00'` locally and then formatted through the UTC `fmt` — those halves now agree.

## Acceptance criteria

- [x] At Berlin (UTC+2), a session at 2026-08-08T00:30 local is attributed to **8 Aug**, not 7 Aug
- [x] The heatmap and hourly chart place it in **00:00–01:00**, not 22:00–23:00
- [x] `from`/`to` are inclusive of full local days
- [x] `fmt()` and `presetToRange` agree at both edges west of UTC
- [x] Heatmap-cell drill-down selects the local weekday+hour window; labels carry no "UTC"
- [x] No `tz`, or an unparseable one, returns UTC-bucketed data without erroring
- [x] `LoadLocation("Europe/Berlin")` resolves with no system zoneinfo
- [x] All stored and transported timestamps remain UTC — no storage or API-shape change

## Review round

One finding, applied in `9d82b66`: parsing bounds at local midnight made the `MAX_DRILLDOWN_DAYS` span DST-sensitive, so a range crossing a fall-back measured 180 days **+ 1 hour** and was rejected. Verified in Berlin — `2026-06-01 → 2026-11-27` was refused while the spring-forward equivalent measured 179.96 days and passed, which is why the existing test stayed green. The guard now counts calendar days, and the new case pins both directions.

Deferred to #213: the sessions-list custom range still parses bare dates as UTC midnight while the presets beside it are local. Different page, untouched here, and its drill-down path is timezone-neutral epoch milliseconds.

## Testing

`analytics.go` had **no tests at all**. It has 8 now, covering both bucketing zones, both DST directions, and that every builder actually receives the location — half-converted analytics, heatmap local but day list UTC, would be worse than uniformly wrong.

`drilldown.test.ts` asserted UTC as correct and is rewritten rather than patched. It and the new `dateRange.test.ts` pass under **UTC, Europe/Berlin, America/Los_Angeles, Pacific/Kiritimati (UTC+14), Pacific/Midway (UTC−11) and Australia/Sydney** — so CI needs no pinned `TZ`, which is more robust than the issue's suggestion.

`go build`, `go vet`, `golangci-lint` (v2.5.0, 0 issues), `go test -race ./...` — 18/18 green. Frontend: typecheck, ESLint (0 errors), Prettier, build clean.

## Deliberately untouched

`formatRateDate` (`lib/pricing.ts`) pins UTC, because a pricing rate is keyed to a **day** rather than an instant — it has a test asserting exactly that, and harmonising it would reintroduce the bug #189 fixed.

> **Unrelated pre-existing issue:** `src/lib/streamingBlocks.test.ts` has 7 failing tests on `main`. CI's frontend job runs typecheck/lint/format/build but not `npm test`.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -97,6 +97,8 @@ One line per package — list files in a directory to see current contents.
 - **`internal/daemon/`** — `agento service` backend: installs and manages Agento as a user-level background service (launchd on macOS, systemd user units on Linux), with embedded unit/plist templates and a `commandRunner` seam for tests.
 - **`internal/build/`** — Version variables injected via `-ldflags`.
 
+**Timezones**: UTC end to end in storage and transport; **aggregation and display follow the browser** (#190). The analytics and insights endpoints take a `tz` IANA name (the frontend sends `Intl.DateTimeFormat().resolvedOptions().timeZone`), and `AnalyticsParams.Loc` is applied via `.In(loc)` before any bucket key is derived — a day, hour or weekday is meaningless until you say whose it is. A bare `YYYY-MM-DD` `from`/`to` is a **local** day boundary (`time.ParseInLocation`), not a UTC one. Missing or unrecognized `tz` falls back to UTC rather than erroring, so pre-`tz` callers are unchanged. `main.go` imports `_ "time/tzdata"` because a distroless container has no `/usr/share/zoneinfo` and every `LoadLocation` would silently degrade the dashboard to UTC. Daily bucket walks (`walkBuckets`) step the **calendar day**, not 24 hours — a local day is 23 or 25 hours across a DST transition, and a fixed step duplicates one key while skipping another. One deliberate UTC exception on the frontend: `formatRateDate` (`lib/pricing.ts`), because a pricing rate is keyed to a day rather than an instant.
+
 **Import rule**: `config` ← `service` ← `api` (never reverse).
 
 ### Frontend
--- a/docs/development.md
+++ b/docs/development.md
@@ -123,3 +123,41 @@ To verify the release config locally (no publish):
 ```bash
 goreleaser release --snapshot --clean
 ```
+
+---
+
+## Timezones
+
+**Storage and transport are UTC, end to end.** Session timestamps are parsed
+from the JSONL `Z` values, cached as UTC, and serialized as UTC on every API
+response. Nothing below the presentation layer knows about the user's zone.
+
+**Aggregation and display follow the browser.** A day, an hour and a weekday
+only mean something once you say whose they are — so the analytics and insights
+endpoints take a `tz` query parameter (an IANA name such as `Europe/Berlin`),
+which the frontend fills in from
+`Intl.DateTimeFormat().resolvedOptions().timeZone`. The backend resolves it and
+applies `.In(loc)` before deriving any bucket key, and interprets a bare
+`YYYY-MM-DD` `from`/`to` as a local day boundary rather than a UTC one.
+
+A missing or unrecognized `tz` falls back to UTC rather than erroring: analytics
+is a read-only dashboard, and refusing to render it over a bad zone string is
+worse than rendering what every caller got before the parameter existed.
+
+Two consequences worth knowing:
+
+- **The zone database is embedded** (`_ "time/tzdata"` in `main.go`). Without
+  it `time.LoadLocation` depends on the host's `/usr/share/zoneinfo`, which a
+  distroless or scratch container does not have — every lookup would fail and
+  silently fall the whole dashboard back to UTC. Costs about 450KB.
+- **Daily bucket walks step the calendar day, not 24 hours.** A local day is 23
+  or 25 hours long across a DST transition, so a fixed step drifts off the wall
+  clock and duplicates one day key while skipping another. The heatmap grid
+  stays a fixed 24 columns: a spring-forward day has one empty cell and a
+  fall-back day has one doubled, which is the honest answer for a "when do I
+  work?" chart.
+
+Non-analytics timestamps already render locally through `toLocale*` helpers and
+need nothing special. One deliberate exception: `formatRateDate`
+(`frontend/src/lib/pricing.ts`) pins UTC, because a pricing rate is keyed to a
+day rather than an instant — see `docs/pricing.md`.
--- a/frontend/src/lib/api.ts
+++ b/frontend/src/lib/api.ts
@@ -710,12 +710,22 @@ export const monitoringApi = {
 
 // ── Analytics ─────────────────────────────────────────────────────────────────
 
+/**
+ * The browser's IANA timezone, sent so the backend buckets days, hours and
+ * weekdays the way the user experiences them. Timestamps stay UTC on the wire;
+ * only aggregation moves. An unset or unrecognised value means UTC server-side.
+ */
+export function browserTimezone(): string {
+  return Intl.DateTimeFormat().resolvedOptions().timeZone
+}
+
 export const analyticsApi = {
   get: (params?: { from?: string; to?: string; project?: string }): Promise<AnalyticsReport> => {
     const qs = new URLSearchParams()
     if (params?.from) qs.set('from', params.from)
     if (params?.to) qs.set('to', params.to)
     if (params?.project) qs.set('project', params.project)
+    qs.set('tz', browserTimezone())
     const query = qs.toString()
     const suffix = query ? `?${query}` : ''
     return request<AnalyticsReport>(`/claude-analytics${suffix}`)
@@ -735,6 +745,7 @@ export const insightsApi = {
     if (params?.ids && params.ids.length > 0) qs.set('ids', params.ids.join(','))
     if (params?.from) qs.set('from', params.from)
     if (params?.to) qs.set('to', params.to)
+    qs.set('tz', browserTimezone())
     const suffix = qs.toString() ? `?${qs.toString()}` : ''
     return request<InsightSummary>(`/claude-sessions/insights/summary${suffix}`)
   },
--- a/frontend/src/lib/dateRange.test.ts
+++ b/frontend/src/lib/dateRange.test.ts
@@ -1,0 +1,72 @@
+import { describe, it, expect } from 'vitest'
+import { fmt, subDays, presetToRange } from '@/pages/analyticsShared'
+
+// These helpers had no coverage at all, which is how `fmt` came to serialise a
+// locally-built Date through a UTC formatter. Every assertion here is in local
+// time; run the suite under a few TZ values to confirm it holds either side of
+// the meridian.
+describe('fmt', () => {
+  it('formats a local date, not its UTC equivalent', () => {
+    // Local midnight on the 1st. toISOString would render this as the previous
+    // month's last day for anyone east of UTC.
+    expect(fmt(new Date(2026, 7, 1))).toBe('2026-08-01')
+  })
+
+  it('formats late-evening local times as the same local day', () => {
+    // 23:30 local on the 8th is the 9th in UTC east of the meridian; the range
+    // the user picked is the local one.
+    expect(fmt(new Date(2026, 7, 8, 23, 30))).toBe('2026-08-08')
+  })
+
+  it('formats early-morning local times as the same local day', () => {
+    // 00:30 local is the previous day in UTC west of the meridian.
+    expect(fmt(new Date(2026, 7, 8, 0, 30))).toBe('2026-08-08')
+  })
+
+  it('zero-pads month and day', () => {
+    expect(fmt(new Date(2026, 0, 5))).toBe('2026-01-05')
+  })
+})
+
+describe('subDays', () => {
+  it('steps back across a month boundary', () => {
+    expect(fmt(subDays(new Date(2026, 7, 3), 7))).toBe('2026-07-27')
+  })
+
+  it('does not mutate its argument', () => {
+    const d = new Date(2026, 7, 3)
+    subDays(d, 7)
+    expect(fmt(d)).toBe('2026-08-03')
+  })
+})
+
+describe('presetToRange', () => {
+  // The bug: presetToRange builds its dates with local calendar arithmetic
+  // (new Date(y, m, 1)) and then serialised them through a UTC formatter, so
+  // the two halves disagreed at every boundary.
+  it('starts "this month" on the local 1st', () => {
+    const { from } = presetToRange('this-month')
+    expect(from.endsWith('-01')).toBe(true)
+    const now = new Date()
+    expect(from).toBe(fmt(new Date(now.getFullYear(), now.getMonth(), 1)))
+  })
+
+  it('ends "last month" on the local last day of that month', () => {
+    const { from, to } = presetToRange('last-month')
+    const now = new Date()
+    expect(from).toBe(fmt(new Date(now.getFullYear(), now.getMonth() - 1, 1)))
+    expect(to).toBe(fmt(new Date(now.getFullYear(), now.getMonth(), 0)))
+  })
+
+  it('ends the rolling presets on the local today', () => {
+    const today = fmt(new Date())
+    for (const preset of ['7d', '30d', '90d', 'all-time'] as const) {
+      expect(presetToRange(preset).to).toBe(today)
+    }
+  })
+
+  it('spans the requested number of days', () => {
+    const { from, to } = presetToRange('7d')
+    expect(from).toBe(fmt(subDays(new Date(to + 'T00:00:00'), 7)))
+  })
+})
--- a/frontend/src/lib/drilldown.test.ts
+++ b/frontend/src/lib/drilldown.test.ts
@@ -9,15 +9,29 @@ import {
   parseRangeBounds,
 } from './drilldown'
 
-// 2026-08-03 is a Monday, 2026-08-09 a Sunday (local time).
+// 2026-08-03 is a Monday, 2026-08-09 a Sunday.
+//
+// Every expectation below is expressed in LOCAL time, matching what the backend
+// now buckets in. Anchoring these with an explicit `Z` is what made the old
+// suite pass while the feature was wrong for everyone outside UTC, so the
+// assertions deliberately parse without one and read local Date fields.
 const FROM = '2026-08-03'
 const TO = '2026-08-09'
 
 describe('parseRangeBounds', () => {
-  it('spans the inclusive `to` date, anchored to UTC', () => {
+  it('spans the inclusive `to` date, anchored to local midnight', () => {
     const bounds = parseRangeBounds(FROM, TO)!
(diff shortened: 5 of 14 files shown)


vibexp-agent:
Deferred to #213 — the sessions-list custom range (`src/lib/timefilter.ts`) still parses bare dates as UTC midnight while the presets beside it are local. Different page, untouched by this PR, and the drill-down path it shares with analytics is timezone-neutral epoch milliseconds.

PR #222, opened 2026-08-09
test(frontend): repair stale streamingBlocks assertions and run vitest in CI (#215)

### LINKED ISSUE #215: test(frontend): CI never runs the vitest suite — 7 streamingBlocks tests have been failing on main unnoticed
Follow-up from PR for #199.

## Problem

Two coupled problems, found while adding the sessions-filter tests in #199.

**1. `npm run test` is not in CI.** The `Frontend` job in `.github/workflows/ci.yml` runs `typecheck`, `lint`, `format:check` and `build` — but never `npm run test`, even though `package.json` defines it and the repo has 8 vitest files. The frontend test suite is effectively dead weight: it can rot indefinitely with CI fully green.

**2. It already rotted.** On `main` today (`eca5070`), `npx vitest run` reports **7 failing tests, all in `frontend/src/lib/streamingBlocks.test.ts`**:

```
× applyThinkingDelta > creates a new thinking block when blocks are empty
× applyThinkingDelta > creates a new thinking block when the last block is text
× applyTextDelta     > creates a new text block when blocks are empty
× applyTextDelta     > creates a new text block when the last block is tool_use
× applyTextDelta     > preserves chronological order: text → tool → text
× applyTextDelta     > consecutive text deltas merge into a single text block
× applyTextDelta     > text deltas after a tool create a new separate text block
```

All seven are the same mismatch — `streamingBlocks.ts` gained a `_key` field on each block and the assertions were never updated:

```
AssertionError: expected [ { type: 'thinking', …(2) } ] to deeply equal [ { type: 'thinking', text: 'hello' } ]
+     "_key": "k1",
      "text": "hello",
      "type": "thinking",
```

The product code is fine; the assertions are stale. Nobody noticed because nothing runs them.

## Why this matters

#199 exists to put the sessions-list filters under test. Tests that CI never executes buy far less than they appear to — this repo just demonstrated the failure mode on itself.

## Direction

1. Update the 7 `streamingBlocks.test.ts` assertions to the current block shape (match on the meaningful fields, or use `objectContaining`, so a future internal key does not break them again).
2. Add a `Test` step to the `frontend` job, after `ESLint`: `run: npm run test`.

Order matters — wiring CI first would land `main` red.

## Out of scope

- Any change to `streamingBlocks.ts` itself; the implementation is correct.
- Frontend coverage thresholds or reporting.

---
## WHAT
**Summary:** Repair the 7 stale `streamingBlocks.test.ts` assertions, then wire `npm run test` into the Frontend CI job so the vitest suite can never rot unnoticed again.

**Scope:**
- Update the 7 failing assertions in `frontend/src/lib/streamingBlocks.test.ts` to tolerate the `_key` field the implementation now attaches to every block.
- Add a `Test` step running `npm run test` to the `frontend` job in `.github/workflows/ci.yml`, positioned after `ESLint`.
- Land them in that order (fix first, then wire) so `main` is never red.

**Affected Components:**
- `frontend/src/lib/streamingBlocks.test.ts` — assertion shape only; 7 tests across the `applyThinkingDelta`, `applyTextDelta` and `streaming block ordering (integration)` describes.
- `.github/workflows/ci.yml` — one new step in the `frontend` job (`working-directory: frontend` is already the job default, so the step needs no path of its own).

**Acceptance Criteria:**
- [ ] `cd frontend && npm run test` exits 0 with 8 test files and 123 tests passing (currently `7 failed | 116 passed`).
- [ ] `.github/workflows/ci.yml` `frontend` job contains a step running `npm run test`, and it appears after the `ESLint` step and before `Build`.
- [ ] The repaired assertions do not name `_key`: adding a further internal field to a block must not break them again (`expect.objectContaining`, or assert the meaningful fields individually).
- [ ] The assertions still fail if the meaningful behaviour regresses — verified by locally inverting one production behaviour (e.g. making consecutive text deltas create separate blocks) and observing a red test.
- [ ] A CI run on the PR shows the new `Test` step executing and passing.

**Out of Scope:**
- Any change to `frontend/src/lib/streamingBlocks.ts`. The implementation is correct; only the assertions are stale.
- Frontend coverage thresholds, coverage reporting, or SonarCloud ingestion of frontend coverage.
- The other 7 test files — they pass today and are untouched.
- Backend CI (the `backend` job already runs `go test`).

## WHY
**Business Value:** A test suite CI never executes is worse than no suite: it looks like coverage while providing none. This repo already paid the price — the streaming-block ordering tests, which guard the chat UI's most order-sensitive logic, have been red on `main` with CI fully green.

**Problem Statement:** The `frontend` job in `.github/workflows/ci.yml` (lines 65–97) runs `typecheck`, `lint`, `format:check` and `build` — never `npm run test`, despite `package.json:15` defining `"test": "vitest run"` and the repo carrying 8 vitest files. Consequently `streamingBlocks.ts` gained a `_key` field on every block without anyone updating the assertions, and 7 tests have been failing silently.

**Expected Impact:** Restores real enforcement for 123 frontend tests. Cost is one CI step (the local suite runs in ~0.7s, so the marginal CI time is dominated by the already-present `npm ci`). Prevents the whole class of "the tests exist but nobody runs them" defects for every future frontend test, including the sessions-filter tests #199 added.

**Context:**
- Follow-up from the PR for #199 (`Related: #199`) — found while adding the sessions-filter tests there, i.e. while adding tests to a suite CI does not run.
- Verified still failing on `main` at `c325de6` on 2026-08-09: `npx vitest run` → `Test Files 1 failed | 7 passed (8)`, `Tests 7 failed | 116 passed (123)`, every failure the same `_key` mismatch the issue body describes. No part of this issue was found to be outdated.

## HOW
**Technical Approach:** Two independent, small changes with a mandatory ordering. The assertion repair is mechanical — each failure is `toEqual` against an object literal that omits `_key`. Rather than adding `_key: 'kN'` to the expectations (which hardcodes the key generator's counter and re-breaks on the next internal field), assert with `expect.objectContaining` on the fields the test is actually about (`type`, `text`), matching the intent of each test. Only once the suite is green does the CI step land, since wiring it first would turn `main` red.

**Implementation Details:**
- `frontend/src/lib/streamingBlocks.test.ts` — 7 assertions:
  - `applyThinkingDelta` — "creates a new thinking block when blocks are empty", "creates a new thinking block when the last block is text".
  - `applyTextDelta` — "creates a new text block when blocks are empty", "creates a new text block when the last block is tool_use", "consecutive text deltas merge into a single text block" (`:202`), "text deltas after a tool create a new separate text block" (`:213`).
  - `streaming block ordering (integration)` — "preserves chronological order: text → tool → text".
  Replace each whole-array/whole-object `toEqual([{ type, text }])` with `toEqual([expect.objectContaining({ type, text })])`, keeping the existing `toHaveLength` assertions that guard block *count* (those are the real subject of the ordering tests and already pass).
- `.github/workflows/ci.yml` — insert after the `ESLint` step (currently lines 90–91):
  ```yaml
      - name: Test
        run: npm run test
  ```

**Testing Strategy:** The change is itself test code plus CI config, so verification is the run:
- `cd frontend && npm run test` locally → 123 passing.
- `npm run lint && npm run format:check` on the edited test file (Prettier and ESLint both cover `frontend/src/**`, and the existing CI steps will reject a mis-formatted test).
- Confirm the guard still bites by temporarily breaking `streamingBlocks.ts` and re-running.
- CI on the PR must show the new `Test` step green.

**Documentation / Configuration:** No docs change; `CLAUDE.md` already documents `npm run test` under the frontend commands. No new dependencies — `vitest` is already a devDependency and the suite runs today. No env vars.

**Files to Create / Modify:**
- Modify: `frontend/src/lib/streamingBlocks.test.ts`
- Modify: `.github/workflows/ci.yml`
- Create: none

**Risks:**
- *Loosening assertions hides a real regression* → use `objectContaining` only for the fields under test and keep the `toHaveLength` count assertions; verify red-on-break before merging.
- *The new CI step surfaces flakiness in the other 7 files* → they pass locally and deterministically today (no timers, no network); if one proves flaky in CI, fix it in this PR rather than skipping the step.
- *Landing the CI step first turns `main` red* → mitigated by ordering; if split across two PRs, the assertion PR must merge first.

**Complexity:** Low (2 files, well under 100 LOC, mechanical pattern).

---

vibexp-agent:
Closes #215

## What

Repairs the 7 stale `streamingBlocks.test.ts` assertions, then wires `npm run test` into the `Frontend` CI job so the vitest suite can no longer rot unnoticed.

- `frontend/src/lib/streamingBlocks.test.ts` — 7 assertions across `applyThinkingDelta`, `applyTextDelta` and `streaming block ordering (integration)` now use `expect.objectContaining({ type, text })` instead of whole-object `toEqual`. The existing `toHaveLength` count assertions are kept untouched.
- `.github/workflows/ci.yml` — new `Test` step in the `frontend` job, positioned after `ESLint` and before `Build`.

## Why

`streamingBlocks.ts` gained a `_key` field on every newly-created block (commit `682f4fa`, Sonar rule S6479) and the assertions were never updated. Nothing caught it, because the `frontend` job ran `typecheck`, `lint`, `format:check` and `build` — never `npm run test`. Seven tests guarding the chat UI's most order-sensitive logic have been red on `main` with CI fully green.

Matching on the fields each test is actually about (rather than adding `_key: 'kN'` to the expectations) avoids hardcoding the module-global key counter, which is order-coupled across tests and would re-break on the next internal field.

## How

Fix first, wire second — landing the CI step before the assertions would have turned `main` red.

## Acceptance criteria

- [x] `cd frontend && npm run test` exits 0 — **8 test files, 123 tests passing** (was `7 failed | 116 passed`).
- [x] `ci.yml` `frontend` job runs `npm run test`, after `ESLint` and before `Build` — confirmed in CI as step 7, between ESLint (6) and Build (9).
- [x] The repaired assertions do not name `_key`; a further internal field cannot break them again.
- [x] The assertions still bite: temporarily disabling the text-merge branch in `streamingBlocks.ts` turned 2 tests red (including the repaired `consecutive text deltas merge into a single text block`), and the source was reverted.
- [x] CI on this PR shows the new `Test` step executing and passing.

## Out of scope / deferred

- No change to `streamingBlocks.ts` — the implementation is correct.
- Deferred to #223: the vitest `include` glob is `src/**/*.test.ts`, so `.tsx` is not matched and a future component test would silently never run. Needs a DOM environment (the repo has no jsdom/testing-library today), so it is tracked separately.

## Notes

`pre-commit run` could not be used as the local gate on this machine — its node hook environment fails to install (`npm error EUNKNOWNCONFIG: --ignore-prepublish`). The equivalent commands (`typecheck`, `lint`, `test`, `format:check`, `build`) were run directly, plus `actionlint` on the modified workflow.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -90,6 +90,9 @@ jobs:
       - name: ESLint
         run: npm run lint
 
+      - name: Test
+        run: npm run test
+
       - name: Prettier check
         run: npm run format:check
 
--- a/frontend/src/lib/streamingBlocks.test.ts
+++ b/frontend/src/lib/streamingBlocks.test.ts
@@ -5,7 +5,7 @@ import type { MessageBlock, SDKContentBlock } from '@/types'
 describe('applyThinkingDelta', () => {
   it('creates a new thinking block when blocks are empty', () => {
     const result = applyThinkingDelta([], 'hello')
-    expect(result).toEqual([{ type: 'thinking', text: 'hello' }])
+    expect(result).toEqual([expect.objectContaining({ type: 'thinking', text: 'hello' })])
   })
 
   it('appends to the last thinking block', () => {
@@ -18,8 +18,8 @@ describe('applyThinkingDelta', () => {
     const blocks: MessageBlock[] = [{ type: 'text', text: 'some text' }]
     const result = applyThinkingDelta(blocks, 'thinking')
     expect(result).toEqual([
-      { type: 'text', text: 'some text' },
-      { type: 'thinking', text: 'thinking' },
+      expect.objectContaining({ type: 'text', text: 'some text' }),
+      expect.objectContaining({ type: 'thinking', text: 'thinking' }),
     ])
   })
 
@@ -34,7 +34,7 @@ describe('applyThinkingDelta', () => {
 describe('applyTextDelta', () => {
   it('creates a new text block when blocks are empty', () => {
     const result = applyTextDelta([], 'hello')
-    expect(result).toEqual([{ type: 'text', text: 'hello' }])
+    expect(result).toEqual([expect.objectContaining({ type: 'text', text: 'hello' })])
   })
 
   it('appends to the last text block', () => {
@@ -49,8 +49,13 @@ describe('applyTextDelta', () => {
     ]
     const result = applyTextDelta(blocks, 'output')
     expect(result).toEqual([
-      { type: 'tool_use', name: 'Bash', id: '1', input: { command: 'ls' } },
-      { type: 'text', text: 'output' },
+      expect.objectContaining({
+        type: 'tool_use',
+        name: 'Bash',
+        id: '1',
+        input: { command: 'ls' },
+      }),
+      expect.objectContaining({ type: 'text', text: 'output' }),
     ])
   })
 
@@ -132,9 +137,14 @@ describe('streaming block ordering (integration)', () => {
     blocks = applyTextDelta(blocks, 'Here is the content.')
 
     expect(blocks).toEqual([
-      { type: 'text', text: 'Let me read the file.' },
-      { type: 'tool_use', id: 'tool-1', name: 'Read', input: { file_path: '/tmp/a.txt' } },
-      { type: 'text', text: 'Here is the content.' },
+      expect.objectContaining({ type: 'text', text: 'Let me read the file.' }),
+      expect.objectContaining({
+        type: 'tool_use',
+        id: 'tool-1',
+        name: 'Read',
+        input: { file_path: '/tmp/a.txt' },
+      }),
+      expect.objectContaining({ type: 'text', text: 'Here is the content.' }),
     ])
 
     // Verify order: text, tool_use, text
@@ -199,7 +209,7 @@ describe('streaming block ordering (integration)', () => {
     blocks = applyTextDelta(blocks, ' ')
     blocks = applyTextDelta(blocks, 'world')
 
-    expect(blocks).toEqual([{ type: 'text', text: 'Hello world' }])
+    expect(blocks).toEqual([expect.objectContaining({ type: 'text', text: 'Hello world' })])
   })
 
   it('text deltas after a tool create a new separate text block', () => {
@@ -210,7 +220,7 @@ describe('streaming block ordering (integration)', () => {
 
     // Must be 3 separate blocks, not 2 (text must not merge across a tool_use)
     expect(blocks).toHaveLength(3)
-    expect(blocks[0]).toEqual({ type: 'text', text: 'before' })
-    expect(blocks[2]).toEqual({ type: 'text', text: 'after' })
+    expect(blocks[0]).toEqual(expect.objectContaining({ type: 'text', text: 'before' }))
+    expect(blocks[2]).toEqual(expect.objectContaining({ type: 'text', text: 'after' }))
   })
 })


vibexp-agent:
Review finding deferred to #223 — the vitest `include` glob (`src/**/*.test.ts`) does not match `.tsx`, so a future component test would silently never run. Out of #215's two-file scope, and it needs a DOM environment (no jsdom/testing-library in the repo today), so it is tracked separately.

PR #225, opened 2026-08-09
fix(claude-sessions): injected user events are not human turns (#197)

### LINKED ISSUE #197: fix(claude-sessions): injected string-content user events (system-reminder, command-message, task-notification) count as human turns
Follow-up from PR #195 (relates to #182).

## Problem

#182 made `message_count` mean conversational turns by excluding `tool_result` carrier events. A second class of non-human user event survives that filter: events whose content is a **plain JSON string** but whose text is machine-injected, not typed.

Sampling 120 local transcripts during the review of PR #195: of **510** string-content user events, **215 (~42%)** are wrappers such as `<command-message>`, `<local-command-stdout>`, `system-reminder` and `task-notification` blocks rather than anything a person wrote.

So turn counts remain inflated — far less than the ~10× before #182, but not zero.

## Why deferred

Both `message_count` and `session_insights.turn_count` share the `isUserTurnContent` / `isTurnStart` predicate, which is exactly the drift-proofing #182 set out to achieve. Filtering these would change `turn_count` and every metric derived from it (`StepsPerTurnAvg`, `AutonomyScore`, `TokensPerTurnAvg`), which #182 explicitly listed as **Out of Scope**. It needs its own decision about whether insight history should shift, and a `CurrentProcessorVersion` bump if so.

## Direction

- Decide whether injected events should be excluded from turns, from insights, or from both — they must stay consistent.
- If yes: extend the shared predicate in `internal/claudesessions/processor.go` to recognize the wrapper markers, bump both `CurrentScannerVersion` and `CurrentProcessorVersion`, and test each wrapper form.
- Be careful with false positives: a user can legitimately paste text containing these markers.

Found during the automated review of PR #195.

---
## WHAT
**Summary:** Stop counting machine-injected user events as human turns — extend the shared `isUserTurnContent` predicate to reject string content that is a Claude Code wrapper (`<task-notification>`, `<command-message>`, `<command-name>`, `<local-command-caveat>`, `<local-command-stdout>`, `<system-reminder>`), and bump both `CurrentScannerVersion` and `CurrentProcessorVersion` so `message_count` **and** `turn_count` shift together.

**Scope:**
- Extend `isUserTurnContent` (`internal/claudesessions/processor.go:259`) to reject wrapper-only string content, keeping the existing `tool_result` rejection.
- Bump `CurrentScannerVersion` (`scanner.go:49`) and `CurrentProcessorVersion` (`processor.go:24`) so cached sessions and stored insights are recomputed.
- Accept the resulting downward shift in `turn_count` and every metric derived from it: `StepsPerTurnAvg`, `AutonomyScore`, `TokensPerTurnAvg`.

**Affected Components:**
- `internal/claudesessions/processor.go` — `isUserTurnContent` gains a wrapper check; `isTurnStart` and the scanner's turn counter both inherit it unchanged, which is the point.
- `internal/claudesessions/scanner.go` — `CurrentScannerVersion` 6 → 7 (forces a re-read of every transcript; mtimes will not have changed).
- `internal/claudesessions/processor.go` — `CurrentProcessorVersion` 6 → 7 (forces insight reprocessing via the `insight_worker.go` rescan loop).

**Acceptance Criteria:**
- [ ] A user event whose content is a bare JSON string beginning with any of the six wrapper markers is not counted in `message_count` and does not start a turn in the insight pipeline.
- [ ] A user event whose content is ordinary prose is still counted — including prose that merely *contains* a marker further in (only a leading marker disqualifies).
- [ ] `event_count` is unchanged for every session: this filters turns, never raw events.
- [ ] `message_count` and `session_insights.turn_count` remain equal for the same transcript — the existing `TestScan_AgreesWithInsightPipeline` guard still passes.
- [ ] Both version constants are bumped, and existing cached sessions plus stored insights are recomputed without any file mtime changing.
- [ ] Measured on the local corpus, turn counts drop materially and in the expected direction (~55% of string-content user events are wrappers), with no session dropping to zero turns that has genuine human input.

**Out of Scope:**
- Any change to `event_count`, `isAssistantReply`, or the sidechain rules.
- Reclassifying the `tool_result` carrier events #182 already excluded.
- Backfilling or annotating historical insight values — they are recomputed, not preserved side by side.
- Filtering wrappers out of the message *content* shown in the session detail UI; this changes counting only.

## WHY
**Business Value:** Turn count is the denominator of the productivity story — `StepsPerTurnAvg`, `AutonomyScore` and `TokensPerTurnAvg` all divide by it. An inflated denominator makes every one of them understate how autonomously the agent ran, and the inflation is not uniform: sessions driven by slash commands, sub-agent delegation and skill invocations inject the most wrappers, so exactly the most autonomous sessions score as the least autonomous.

**Problem Statement:** #182 made `message_count` mean conversational turns by excluding `tool_result` carrier events. A second class survives that filter: events whose content is a plain JSON string that a machine injected rather than a person typing. `isUserTurnContent` only inspects *array* content for `tool_result` blocks — `parseContentBlocks` returns `nil` for anything not starting with `[`, so every string-content user event passes unconditionally.

**Expected Impact:** Turn counts fall by roughly the wrapper share of string-content user events; on the reference corpus that is ~55% of those events. Every derived insight moves accordingly — `StepsPerTurnAvg` and `TokensPerTurnAvg` up, `AutonomyScore` up. Historical figures change on the next scan. That is intended: they were measuring machine chatter as human input. One forced full re-scan (~18s on ~1,500 transcripts) plus one insight reprocessing pass; since #208 the scan runs in the background, so no request blocks on it.

**Context:**
- Follow-up from PR #195 (`Related: #195`), deferred from #182 (`Related: #182`), which listed insight-metric movement as explicitly out of scope and left the decision open.
- Verified still open on `main` at `c325de6` on 2026-08-09: `isUserTurnContent` (`processor.go:259`) rejects only `tool_result`, and both version constants sit at 6.
- **Re-measured on the current corpus (2026-08-09), superseding the body's figures.** The body cites 215/510 (~42%) across 120 transcripts sampled during the #195 review. Across the 60 most recently modified local transcripts: **138 of 253** non-sidechain string-content user events (~**55%**) are wrappers. The sample also found **two marker forms the body does not list** — `<command-name>` (20) and `<local-command-caveat>` (20) — alongside `<task-notification>` (53), `<command-message>` (26), `<local-command-stdout>` (16) and `<system-reminder>` (3). Both must be handled; the original four-marker list is incomplete.
- **Decision (2026-08-09):** exclude from **both** `message_count` and insights, via the one shared predicate. Filtering only `message_count` would reintroduce precisely the drift #182 existed to eliminate — two live definitions of "turn". Insight history shifting is accepted as the cost of correctness, and `CurrentProcessorVersion` is bumped to make it happen deterministically rather than leaving a mix of old and new values in the table.

## HOW
**Technical Approach:** One predicate, one place. `isUserTurnContent` already owns the "is this genuine human input" question for both the scanner's `message_count` and the insight pipeline's `isTurnStart`, so the change is additive inside it: after the existing `tool_result` scan over array content, handle the bare-string case that currently falls through. Match on a **leading** marker only, after trimming leading whitespace, so that prose which merely quotes a marker mid-text still counts — a leading-anchored check is the cheap defence against the false positive the body flags. Then bump both version constants so nothing is left half-recomputed.

**Implementation Details:**
- `internal/claudesessions/processor.go`:
  - Add a package-level `injectedTurnMarkers = []string{"<task-notification>", "<command-message>", "<command-name>", "<local-command-caveat>", "<local-command-stdout>", "<system-reminder>"}`, with a comment recording that these are Claude Code's own injection wrappers and that the list is empirical — a new Claude Code release can add one.
  - Add `isInjectedUserContent(raw json.RawMessage) bool`: return false unless `raw` is a JSON string (`raw[0] == '"'`); `json.Unmarshal` it into a `string`; `strings.TrimSpace`; report whether it `strings.HasPrefix` any marker. Anchoring on the prefix is deliberate and belongs in the comment.
  - In `isUserTurnContent`, keep the existing loop, then before the final `return true` add `return !isInjectedUserContent(content)`. Update the doc comment to describe both excluded classes.
  - `CurrentProcessorVersion` (`:24`) 6 → 7.
- `internal/claudesessions/scanner.go` — `CurrentScannerVersion` (`:49`) 6 → 7.
- No call-site changes: `isTurnStart` (`processor.go:288`), `turn_processor.go`, `depth_processor.go`, `rhythm_processor.go` and the scanner's counter all route through the predicate already.

**Testing Strategy:**
- `internal/claudesessions/processors_test.go` — a table test over `isUserTurnContent`: one case per marker; a marker with leading whitespace/newline (must be rejected); prose containing a marker mid-string (must be **counted**); ordinary prose; empty string; array content with `tool_result` (still rejected); array content without (still counted).
- `internal/claudesessions/counts_test.go` — `TestScan_StringContentUserEventIsATurn` (`:147`) currently asserts that *any* string-content user event is a turn; narrow it to plain prose and add the wrapper counterpart asserting it is not. `TestScan_MessageCountIsTurnsNotEvents` (`:80`) should gain a wrapper event and assert `event_count` still counts it while `message_count` does not.
- `TestScan_AgreesWithInsightPipeline` (`:222`) is the drift guard and must keep passing untouched — it is the reason the change lives in one predicate.
- `TestIncrementalScan_ScannerVersionRecomputesCounts` (`:285`) already covers version-bump recomputation; extend or mirror it for the processor version so stored insights are shown to reprocess.
- Sanity-check against the real corpus before merging: compare aggregate turn counts before and after, and confirm no session with genuine human input drops to zero turns.

**Documentation / Configuration:** Update the `internal/claudesessions` paragraph in `CLAUDE.md` where `message_count` and `isUserTurnContent` are described, to state that injected wrapper events are excluded alongside `tool_result` carriers and that the marker list is empirical. No new dependencies (`strings` and `encoding/json` are already imported), no env vars, no config.

**Files to Create / Modify:**
- Modify: `internal/claudesessions/processor.go`, `internal/claudesessions/scanner.go`, `internal/claudesessions/processors_test.go`, `internal/claudesessions/counts_test.go`, `CLAUDE.md`
- Create: none

**Risks:**
- *False positives — a user pastes text starting with one of these markers* → anchored prefix match after trimming, so only text that **begins** with a wrapper is rejected; a mid-text mention still counts, and this is covered by an explicit test. Residual risk accepted: a user deliberately opening a message with `<system-reminder>` loses one turn.
- *False negatives — a future Claude Code release adds a new wrapper* → the list is empirical and will drift; the mitigation is the comment plus the fact that a missed marker only leaves the count where it is today, never worse. Re-sampling the corpus is a periodic maintenance task, not a blocker.
- *Insight history moves under existing dashboards* → intended and decided; the `CurrentProcessorVersion` bump makes it a single deterministic recomputation rather than a slow drift across mixed values.
- *Only one of the two versions bumped* → leaves `message_count` and `turn_count` disagreeing, the exact failure #182 removed; the drift test catches it, and both bumps are acceptance criteria.

**Complexity:** Medium (5 files, well under 500 LOC, but it moves stored historical metrics and touches the shared predicate two subsystems depend on).

---

vibexp-agent:
Closes #197

## What

Stops counting machine-injected user events as human turns. `isUserTurnContent` — the single predicate behind both `message_count` and the insight pipeline's `isTurnStart` — now rejects string content that opens with one of Claude Code's own injection wrappers, alongside the `tool_result` carriers #182 already excluded.

Markers (`injectedTurnMarkers`): `<task-notification>`, `<command-message>`, `<command-name>`, `<local-command-caveat>`, `<local-command-stdout>`, `<system-reminder>`.

`CurrentScannerVersion` and `CurrentProcessorVersion` both go 6 → 7, so cached `message_count` and stored `turn_count` are recomputed together.

## Why

`parseContentBlocks` returns `nil` for anything not starting with `[`, so every string-content user event passed unconditionally. Turn count is the denominator of the productivity story — `StepsPerTurnAvg`, `AutonomyScore`, `TokensPerTurnAvg` all divide by it — and the inflation was not uniform: sessions driven by slash commands and sub-agent delegation inject the most wrappers, so the most autonomous sessions scored as the least autonomous.

## How

The match is **prefix-anchored after a trim, never a substring**. That is load-bearing, not defensive: the corpus already contains a genuine human prompt that quotes `<system-reminder>` mid-sentence, and a `Contains` test would stop counting it. Covered explicitly by `TestIsUserTurnContent`.

Both constants are bumped because the predicate is shared — bumping one would leave `message_count` and `turn_count` recomputed at different times and disagreeing, exactly the drift #182 existed to remove. `TestVersionConstants_BumpedTogetherForInjectedTurns` pins it.

## Measured impact (local corpus, 1,598 transcripts)

- Counted user turns: **4,733 → 2,962** (−1,771, **−37.4%**).
- **38 of 774** sessions drop to zero user turns. Every one is a pure slash-command session (`/plugin`, `/mcp`, `/reload-skills`) containing no prose of any shape — verified by construction, since only string content can be filtered. This satisfies the acceptance criterion "no session dropping to zero turns that has genuine human input".
- Derived metrics move accordingly: `StepsPerTurnAvg` and `TokensPerTurnAvg` up, `AutonomyScore` up. Intended — they were measuring machine chatter as human input.

## Review finding, fixed in this PR

Filtering the predicate also silenced the **preview**, which is gated on it and is the last fallback in `ResolveDisplayTitle`. All 38 wrapper-only sessions have no `custom-title` and no `ai-title`, so they would have rendered as **blank, unidentifiable rows** in the sessions list.

Preview now uses a deliberately weaker rule than the counter: a genuine turn always wins and replaces a wrapper already seeded, but a session that never contains one keeps the wrapper as its label. A noisy label beats a blank one. `TestScan_WrapperOnlySessionKeepsALabel`.

## Acceptance criteria

- [x] String content opening with any of the six markers is not counted and does not start a turn.
- [x] Ordinary prose still counts, including prose that mentions a marker mid-text.
- [x] `event_count` unchanged — the wrapper added to `countsFixture` moves it 14 → 15 while `message_count` stays 4.
- [x] `message_count` and `turn_count` stay equal — `TestScan_AgreesWithInsightPipeline` passes untouched, and its fixture now actually contains a wrapper, so the guard exercises this change.
- [x] Both constants bumped; recomputation happens with no file mtime changing (`TestIncrementalScan_ScannerVersionRecomputesCounts`).
- [x] Corpus measured; direction and magnitude as expected, no session with human input zeroed.

## Deferred (cross-linked)

- **#226** — array-content injections (`Base directory for this skill:` ×843, `[Request interrupted by user]` ×120) still count. #197 scoped itself to string content; this is ~963 more turns and needs its own version bump and false-positive analysis.
- **#227** — `journey.go` segments turns with its own weaker rule (`blocks[0].Type != "tool_result"`) that bypasses the shared predicate, so journey turns now disagree with `message_count`.

## Notes

`TestScan_StringContentUserEventIsATurn` is renamed to `TestScan_TypedStringContentUserEventIsATurn`: its fixture is a genuine prompt so it passes unchanged, but its old name and comment asserted a rule that is no longer the whole rule.


vibexp-agent:
Review findings deferred: **#226** (array-content injected user events — skill preambles and interrupt notices, ~963 more turns on the local corpus — still count; #197 scoped itself to string content) and **#227** (`journey.go` segments turns with its own weaker rule that bypasses the shared predicate, so journey turns now disagree with `message_count`).

Issue #218, opened 2026-08-09
feat(pricing): context-length rate tiers — Qwen long-context usage is under-reported by up to 300%


vibexp-agent:
Follow-up from #206 (relates to #187, #205). #206 explicitly defers this and asks for a separate issue.

## Problem

`Rate` (`internal/pricing/types.go`) has no context-length dimension, so a rate is a function of `(model, time)` only. Alibaba tiers by **input length**, so the catalog seeds base tiers and silently under-reports everything above them.

Verified against `https://www.alibabacloud.com/help/en/model-studio/model-pricing` (International/Singapore) on 2026-08-09:

| Model | Seeded base tier | Higher tier | Under-report |
|---|---|---|---|
| `qwen3.7-plus` | $0.40/$1.60 (0–256K) | $1.20/$4.80 (256K–1M) | up to **200%** |
| `qwen3.6-flash` | $0.25/$1.50 (0–256K) | $1.00/$4.00 (256K–1M) | up to **300%** |
| `qwen3.5-plus` | $0.40/$2.40 (0–256K) | $0.50/$3.00 (256K–1M) | 25% |
| `qwen3-max` | $1.20/$6.00 (≤32K) | $2.40/$12.00 (32–128K), $3.00/$15.00 (128–256K) | up to 150% |

`qwen3.7-max` is flat 0–1M and unaffected.

## Why it matters

Long context is exactly where cost concentrates, so the error is biased toward the expensive sessions rather than spread evenly. Unlike an unpriced model — which lands visibly in the unknown-pricing bucket and is disclosed as a floor — an under-tiered rate resolves *confidently* and reports a wrong number with nothing to signal it.

Currently no Qwen usage exists in the reference corpus, so nothing is mispriced today.

## Direction

Structural, not data — which is why #206 stayed a data change:

- `Rate` gains a context-tier dimension (e.g. an optional ascending `[]TierRate` keyed by an inclusive upper token bound, absent meaning flat).
- `Resolver.Resolve` takes the message's input token count alongside `model` and `spent_at`, or returns a rate object the accumulator can evaluate against a token count.
- The per-message cost accumulator passes the count it already has.
- `catalog.json` gains the higher tiers for the four models above.

Anthropic does not tier this way, so the flat path must stay the default and stay free — no behavior change for existing entries.

## Note

The tier boundary is on **input tokens**; confirm against the provider whether cached-read tokens count toward it before implementing, since that changes which tier a long cached conversation falls into.

---
[...]

PR #228, opened 2026-08-09
fix(claude-sessions): persist the sub-agent event_count (#196)

### LINKED ISSUE #196: fix(claude-sessions): sub-agent EventCount is computed but never persisted — no event_count column on claude_subagent_cache
Follow-up from PR #195 (relates to #182).

## Problem

#182 split `message_count` (conversational turns) from `event_count` (raw JSONL events) for sessions. Sub-agent summaries go through the same `readSummaryFile`/`processSummaryEvent` path, so `ClaudeSessionSummary.EventCount` **is** computed for every sub-agent transcript — but `claude_subagent_cache` has no `event_count` column, so `upsertSubagentRow` (`internal/claudesessions/scanner.go`) silently drops it.

Net effect: a sub-agent's `message_count` now correctly means turns, but the raw event total that sessions retain is unavailable for sub-agents. The asymmetry is invisible from the outside and easy to trip over when adding aggregate reporting.

## Why deferred

Out of scope for #182, whose stated scope was `claude_session_cache` only. It needs its own migration plus `ClaudeSubagent.EventCount`, the `upsertSubagentRow`/`loadSubagents` column plumbing, and a frontend type change — a coherent change of its own, not a line in the counting fix.

## Direction

- Migration: `ALTER TABLE claude_subagent_cache ADD COLUMN event_count INTEGER NOT NULL DEFAULT 0;`
- Add `EventCount` to `ClaudeSubagent` (`types.go`) and thread it through `upsertSubagentRow` + the sub-agent select.
- Bump `CurrentScannerVersion` so existing sub-agent rows backfill (mtimes will not have changed).
- `frontend/src/types.ts` `ClaudeSubagent`, and decide whether `ClaudeSessionDetailPage.tsx:302` ("N messages") should expose it.

Found during the automated review of PR #195.

---
## WHAT
**Summary:** Persist the sub-agent `EventCount` the scanner already computes — add an `event_count` column to `claude_subagent_cache` and thread it through the write, the read and the API type — so sub-agents retain the raw event total sessions have kept since #182.

**Scope:**
- Schema migration adding `event_count` to `claude_subagent_cache`.
- `ClaudeSubagent.EventCount` (Go) plus the `upsertSubagentRow` INSERT/UPDATE and the `ListSubagents` SELECT/Scan.
- `CurrentScannerVersion` bump, so existing sub-agent rows backfill without any file mtime changing.
- `ClaudeSubagent.event_count` in `frontend/src/types.ts`, mirroring the Go struct.

**Affected Components:**
- `internal/storage/sqlite.go` — append one migration to the `migrations` slice (`ALTER TABLE claude_subagent_cache ADD COLUMN event_count INTEGER NOT NULL DEFAULT 0;`).
- `internal/storage/sqlite_test.go:47` — the hardcoded expected `MAX(version)` moves from `20` to `21`.
- `internal/claudesessions/types.go:280` — `EventCount int \`json:"event_count"\`` on `ClaudeSubagent`, placed beside `MessageCount`.
- `internal/claudesessions/scanner.go` — `upsertSubagentRow` (~`:993`) column list, placeholders and `DO UPDATE SET`; `ListSubagents` (~`:1045`) SELECT and `rows.Scan`; `CurrentScannerVersion` (`:49`) 6 → 7; the `readSubagentSummary` doc comment (~`:1315`) that currently states the value is not persisted.
- `frontend/src/types.ts:621` — `event_count: number` on `ClaudeSubagent`.

**Acceptance Criteria:**
- [ ] `SELECT MAX(version) FROM schema_migrations` on a fresh DB returns 21, and `sqlite_test.go` asserts it.
- [ ] After a scan, `SELECT event_count FROM claude_subagent_cache` is non-zero for every sub-agent row that has any events, and `event_count >= message_count` holds for every row.
- [ ] Pre-existing sub-agent rows (written before this change, unchanged mtimes) carry a correct `event_count` after one scan, driven by the `CurrentScannerVersion` bump alone.
- [ ] `GET /api/claude-sessions/{id}` returns `event_count` on each entry of `subagents`, and `frontend/src/types.ts` declares it.
- [ ] Migrating an existing database with sub-agent rows succeeds and leaves every other column untouched.
- [ ] `npm run typecheck` passes with the new required field.

**Out of Scope:**
- Any UI change. `ClaudeSessionDetailPage.tsx:302` keeps showing `{subagent.message_count} messages`; exposing the raw event count in the UI is a separate decision (see Context).
- Aggregating sub-agent `event_count` into the parent session's `event_count`, or into any analytics figure. `ClaudeSessionSummary.EventCount` stays main-thread, exactly as `Usage` and `Cost` do.
- Any change to how `EventCount` is computed — `readSummaryFile` already produces it correctly.

## WHY
**Business Value:** Closes a silent asymmetry between sessions and sub-agents. The value already exists in memory on every scan and is thrown away at the write; the cost of keeping it is one integer column.

**Problem Statement:** #182 split `message_count` (conversational turns) from `event_count` (raw JSONL events) for sessions. Sub-agent transcripts run through the same `readSummaryFile` path, so `ClaudeSessionSummary.EventCount` **is** computed for each of them — but `claude_subagent_cache` has no such column, so `upsertSubagentRow` drops it. The gap is invisible from the outside: a sub-agent's `message_count` now means turns, while the raw total that gives it context is simply gone. The next person adding aggregate reporting over sub-agents finds the field missing with no explanation.

**Expected Impact:** Data completeness, not user-visible behaviour: no figure on any page changes. One `INTEGER NOT NULL DEFAULT 0` column on a table with one row per delegated sub-agent — negligible storage. One forced re-read of every transcript on the next scan (the `CurrentScannerVersion` bump), taking roughly the ~18s a full 1,500-transcript scan costs; since #208 that scan runs in the background, so no request blocks on it.

**Context:**
- Follow-up from PR #195 (`Related: #195`), deferred from #182 (`Related: #182`) whose stated scope was `claude_session_cache` only.
- Verified still open on `main` at `c325de6` on 2026-08-09: `claude_subagent_cache` has no `event_count` column (`internal/storage/sqlite.go:276` onward), and `scanner.go:1315` carries an explicit `// … so it is not persisted — see #196` comment. Nothing in the original body was outdated; the schema has since gained the #188 cost columns, so this migration appends after those.
- **Decision (2026-08-09):** persist and type it, but make **no UI change**. Surfacing a raw event count next to "N messages" is a product decision that does not belong inside a data-completeness fix; the column makes it available the moment someone wants it. The original body left this open ("decide whether `ClaudeSessionDetailPage.tsx:302` should expose it") — it is now decided as no.
- Depends on nothing; blocks nothing currently open.

## HOW
**Technical Approach:** Follow the exact shape of the #188 cost-column change, which added several columns to this same table: append a migration, extend the struct, extend the two SQL statements symmetrically, bump the scanner version so cached rows are re-read. The `NOT NULL DEFAULT 0` default means the migration itself needs no backfill — the version bump does it on the next scan, which is the established mechanism for "cached rows are incomplete but no file changed".

**Implementation Details:**
- `internal/storage/sqlite.go` — append to the `migrations` slice (`:23`), after the current final migration (`agent_breakdown`, `:467`):
  ```sql
  ALTER TABLE claude_subagent_cache ADD COLUMN event_count INTEGER NOT NULL DEFAULT 0;
  ```
  Migrations are append-only and each runs exactly once, tracked in `schema_migrations`.
- `internal/storage/sqlite_test.go:47` — `if version != 21`. The version is hardcoded, so every migration must update it or `TestNewSQLiteDB_MigrationVersion` fails.
- `internal/claudesessions/types.go` — add `EventCount int \`json:"event_count"\`` to `ClaudeSubagent` directly after `MessageCount`, with a comment noting it is the raw top-level event total (mirroring `ClaudeSessionSummary`).
- `internal/claudesessions/scanner.go`:
  - `upsertSubagentRow` — add `event_count` to the column list next to `message_count`, add one `?`, add `event_count = excluded.event_count` to `DO UPDATE SET`, and pass `s.EventCount` in the matching argument position.
  - `ListSubagents` — add `event_count` to the SELECT next to `message_count` and `&sa.EventCount` to `rows.Scan`, in the same position.
  - `CurrentScannerVersion` (`:49`) 6 → 7.
  - Update the `readSubagentSummary` comment (~`:1315`) — delete the "no column … not persisted — see #196" sentence and state that the value is persisted.
- `frontend/src/types.ts:621` — `event_count: number` after `message_count`.

**Testing Strategy:**
- `internal/storage/sqlite_test.go` — update the version assertion; the existing table-presence test already covers `claude_subagent_cache`.
- `internal/claudesessions/subagent_test.go` — extend the existing sub-agent scan test: write a transcript whose event total differs from its turn count (tool_result carrier events give exactly this), scan, and assert the persisted `EventCount` equals the raw line count while `MessageCount` stays lower.
- Round-trip: `upsertSubagentRow` then `ListSubagents` returns the same `EventCount` — the two statements must not drift.
- Backfill: write a row with `event_count = 0` and an unchanged mtime, run a scan with the bumped `CurrentScannerVersion`, assert the value is filled in. `internal/claudesessions/scanner_test.go` already has the version-bump-forces-reread pattern to follow.
- `cd frontend && npm run typecheck`.

**Documentation / Configuration:** Update the `internal/claudesessions` paragraph in `CLAUDE.md` where it describes `message_count` vs `event_count`, to say the split now holds for sub-agents too. No new dependencies, no env vars, no config.

**Files to Create / Modify:**
- Modify: `internal/storage/sqlite.go`, `internal/storage/sqlite_test.go`, `internal/claudesessions/types.go`, `internal/claudesessions/scanner.go`, `internal/claudesessions/subagent_test.go`, `frontend/src/types.ts`, `CLAUDE.md`
- Create: none

**Risks:**
- *Forgetting the hardcoded version in `sqlite_test.go`* → CI fails loudly rather than silently; called out explicitly above.
- *Argument-position drift in `upsertSubagentRow`* — the INSERT is a long positional list, so inserting a column mid-list can silently shift every later argument by one → add `event_count` immediately after `message_count` in **both** the column list and the argument list, and rely on the round-trip test to catch a mismatch (types differ enough that a shift would fail the scan or corrupt values visibly).
- *A forced full re-read on upgrade* → expected and already routine; since #208 the scan runs in the background, so only a cold cache waits.
- *Adding a required field to a TypeScript interface breaks a consumer* → `typecheck` in CI catches it; the only consumer is `ClaudeSessionDetailPage.tsx`, which reads named fields.

**Complexity:** Low (6 files, well under 100 LOC, follows an established migration + version-bump pattern).

---

vibexp-agent:
Closes #196

## What

Persists the sub-agent `event_count` the scanner already computed and then threw away, so sub-agents keep the turn/event split sessions have had since #182.

- Migration **22** — `ALTER TABLE claude_subagent_cache ADD COLUMN event_count INTEGER NOT NULL DEFAULT 0;`
- `ClaudeSubagent.EventCount` (Go) and `event_count` (TypeScript), beside `MessageCount`.
- `upsertSubagentRow` write and `ListSubagents` read.
- `CurrentScannerVersion` **7 → 8**, so existing rows backfill.

## Why

Sub-agent transcripts run through the same `readSummaryFile` path as parents, so `EventCount` was fully populated on every scan — and `upsertSubagentRow` dropped it for want of a column. A sub-agent's `message_count` meant turns while the raw total that gives it context was gone, and the gap was invisible from the outside: `scanner.go` even carried a `// so it is not persisted — see #196` comment.

No figure on any page changes. Per the issue's decision, the UI is untouched: `ClaudeSessionDetailPage` keeps showing `{subagent.message_count} messages`, and the column simply makes the value available.

## How

Follows the shape of #188's cost columns on the same table: append a migration, extend the struct, extend both SQL statements symmetrically, bump the scanner version. `NOT NULL DEFAULT 0` means the migration needs no backfill — the version bump does it on the next scan, which is the established mechanism for "cached rows are incomplete but no file changed". The value is deliberately **not** rolled up into the parent's `event_count`, which stays main-thread exactly as `Usage` and `Cost` do.

### Deviations from the HOW (both forced by earlier PRs in the same batch)

- Migration is **22**, not 21 — #218 took 21 (`model_pricing_tier`).
- `CurrentScannerVersion` goes **7 → 8**, not 6 → 7 — #197 took 7.
- Consequently `sqlite_test.go` asserts 22, and `TestMigration20_AppliesToExistingDatabaseWithRows` needed `ALTER TABLE claude_subagent_cache DROP COLUMN event_count` in its rollback, or migration 22 fails to re-apply with `duplicate column name`.

## Verification

The issue flags positional drift in the long 25-column upsert as the main risk. Checked mechanically rather than by eye: column list, placeholders and bind arguments all number 25 and align one-to-one, with `event_count ← s.EventCount` at position 11. The SELECT/Scan side is proven by the round-trip test returning 3/2 rather than shifted values.

## Acceptance criteria

- [x] Fresh DB reports `MAX(version) = 22`, asserted in `sqlite_test.go`.
- [x] After a scan `event_count` is non-zero for a sub-agent with events, and `event_count >= message_count` holds — `TestSubagent_EventCountPersists` (3 events, 2 turns).
- [x] Pre-existing rows backfill from the version bump alone, with no mtime change — `TestSubagent_EventCountBackfillsOnVersionBump`.
- [x] `GET /api/claude-sessions/{id}` returns `event_count` per sub-agent (scanned into the struct, which carries the json tag), and `frontend/src/types.ts` declares it.
- [x] Migrating an existing database with sub-agent rows succeeds and leaves other columns untouched — covered by the migration re-apply test.
- [x] `npm run typecheck` passes with the new required field.

## Deferred

**#229** — `TestMigration20_AppliesToExistingDatabaseWithRows` needs a hand-written inverse statement for every new migration, and a forgotten one fails *silently* (the test passes while no longer exercising the migration it is named after). This is the second PR in a row to extend it.


vibexp-agent:
Review finding deferred to **#229** — `TestMigration20_AppliesToExistingDatabaseWithRows` requires a hand-written inverse statement for every new migration, and a forgotten one fails silently (the test keeps passing while no longer exercising the migration it is named after). This PR is the second in a row to have to extend it.

PR #241, opened 2026-08-11
docs(readme): rebuild the README around install and conversion

vibexp-agent:
## What

A full rewrite of `README.md`, plus 15 regenerated screenshots under `docs/images/`.

## Why

The old README worked against itself for a first-time visitor:

- Six badges, then **`## Features`** as a wall of ten collapsed `<details>`. Nothing about the product was visible without clicking.
- **Installation sat at line 291**, below the entire feature list and the demo section.
- Every screenshot was hosted on `user-attachments` and several releases out of date: no Insights page, no cost-by-model, no per-session metrics, no model pricing catalog, no Data & Analytics settings.
- The pitch never said what the thing is in one line.

## What changed

**Structure now follows what a visitor needs, in order:** one-line positioning, a screenshot of the app working, the install command, then a feature tour where every claim sits next to a picture of it.

- Install moved above the feature list, with Homebrew and direct download side by side, and the Claude Code CLI requirement stated up front.
- The star ask appears twice: a callout near the top and a closing block.
- Feature sections are open by default and lead with the analytics surface, which is the part of Agento that has no equivalent elsewhere.
- Long-tail reference material (env vars, CLI reference, build from source, observability, notifications, Claude profiles, auto-update) stays collapsed at the bottom.
- Removed the separate `## Demo` section: its screenshots are now inline beside the features they demonstrate. The two chat videos are kept, collapsed.
- No em dashes in the prose.

**Screenshots.** All 15 are captured from the current UI at 1440px, palette-optimised, **864 KB for the whole set**. They are generated from a **synthetic session corpus**, so no real project names, filesystem paths, session titles or spend figures appear anywhere. The agents and scheduled tasks shown are fabricated examples.

## Accuracy

Every claim was checked against the code on `main`: pricing is per assistant message at its own model and timestamp, cost is attributed to the model that spent it including sub-agent work, durations are active time with a user-configurable idle threshold, the sessions list is paged and filtered in SQL, unknown models are reported as unknown, and integrations include WhatsApp.

## Notes for review

- The four `frontend-*` pre-commit hooks are skipped in this commit. They match `^frontend/src/**` only, and nothing outside `README.md` and `docs/images/` changed. They were skipped rather than run because the local hook environment cannot install here (`npm` rejects the `--ignore-prepublish` flag pre-commit passes).
- Worth a look on the rendered page rather than the diff: <https://github.com/shaharia-lab/agento/blob/docs/readme-conversion-refresh/README.md>
- The em-dash removal applies to README prose only. Screenshots still show the app's own UI copy, which contains em dashes in the insight cards. Say the word if you want those changed in the product too.


vibexp-agent:
### Follow-up: em dashes removed from the app itself

`eac67e1` extends this PR past the README into the product copy, so the screenshots and the UI agree.

**What changed.** Every em dash in text a user reads, rewritten per sentence rather than swapped mechanically:

- **React UI**: insight cards, chart subtitles, empty states, dialogs, tooltips and aria labels, toasts, permission-mode options, the onboarding wizard, the update banner.
- **CLI output**: `agento` root description, `agento update`, `agento service`.

**What was left alone, deliberately:**

- The bare `—` used as the **empty-value placeholder** in tables and stat cells (`if (!n) return '—'`). That is a typographic convention for "no value", not prose, and swapping it changes the visual grid for nothing.
- Comments and test descriptions, which no user sees.

**One reworded rather than repunctuated.** The session detail control read `Showing 12 of 40 events — show more`, which does not survive losing its dash. It is now `Show more (12 of 40 events)`.

**Screenshots recaptured.** Nine of the fourteen images show affected copy, so all of them were retaken from a build of this branch. The set is still 0.83 MB.

### Checks

Run by hand, since two hooks cannot run in this checkout:

| Check | Result |
|---|---|
| `npm run lint` | clean (14 pre-existing `react-refresh` warnings) |
| `npm run typecheck` | clean |
| `npm run test` | 193 passed, 14 files |
| `go test ./cmd/...` | pass |
| `make build` | pass |

`DataAnalyticsTab.test.tsx` asserted on one of the changed strings and was updated with it.

### Two pre-existing environment failures, for the record

Neither is caused by this branch, and **both reproduce on the unmodified base commit**:

1. The four `frontend-*` pre-commit hooks cannot install their node environment: `npm error Unknown cli flag: --ignore-prepublish`. Recent npm dropped the flag that pre-commit passes.
2. `golangci-lint` v2.5.0 panics: `file requires newer Go version go1.26 (application built with go1.25)`. The pinned binary predates the local go1.26.5 toolchain.

Both are worth fixing separately, since right now nobody gets a working local pre-commit run. CI runs both gates on this PR, which is the real check.

PR #243, opened 2026-08-11
feat(claude): index every Claude config dir, target one per run (#242)

### LINKED ISSUE #242: fix(claude): CLAUDE_CONFIG_DIR is unsupported — the settings file, the sessions scanner and the daemon all pin ~/.claude
A user reported that they run two Claude Code accounts side by side, switching
between them with `CLAUDE_CONFIG_DIR` (the documented Claude Code variable —
e.g. `CLAUDE_CONFIG_DIR="$HOME/.claude-personal" claude`). They asked for a way
to tell Agento which Claude config dir to use, having tried
`CLAUDE_CONFIG_DIR="$HOME/.claude-personal" agento web` without success, and
suggested it belongs in the settings view.

## Investigation

`CLAUDE_CONFIG_DIR` appears **nowhere** in the codebase — zero hits across Go,
TypeScript and docs. The installed Claude Code binary references it 32 times, so
the CLI itself honours it.

**The env var does reach the subprocess.** The SDK's `buildEnv` inherits
`os.Environ()` (`claude-agent-sdk-go@v0.3.1` `claude/process.go:522`). Verified
by placing a fake `claude` on `PATH` that dumps its argv and environment, then
running both `agento ask` and a real web chat with
`CLAUDE_CONFIG_DIR=$HOME/.claude-personal`: the subprocess received
`CLAUDE_CONFIG_DIR=/home/…/.claude-personal` in both cases. So account switching
for the agent run itself is *not* what is broken. Three other things are.

### A · Web chat force-feeds the default config dir's settings file

`internal/service/chat_service.go:318` calls
`config.LoadProfileFilePath(session.SettingsProfileID)`; with no profile selected
that returns `$HOME/.claude/settings.json` unconditionally
(`internal/config/profiles.go:72-75`), and `internal/agent/runner.go:187` turns
it into `--settings`. Captured from the fake-`claude` dump, with
`CLAUDE_CONFIG_DIR` pointing at `.claude-personal`:

```
  [--settings]
  [/home/…/.claude/settings.json]
```

So the *other* config dir's settings are applied to the run, and
`$CLAUDE_CONFIG_DIR/settings.json` never is. `agento ask` does not exhibit this —
it leaves `SettingsFilePath` empty and no `--settings` appears in its dump.

### B · The Claude Sessions surface is blind to an alternate config dir

`claudesessions.ClaudeHome()` (`internal/claudesessions/scanner.go:201`) and
`config.ClaudeSettingsDirPath()` (`internal/config/profiles.go:23`) both
hardcode `$HOME/.claude`. Consumers include the scanner
(`scanner.go:283,391,1789`), the todos loader (`scanner.go:2019`), the projects
list (`projects.go:108`) and the journey builder (`journey.go:190`). With the var
set, transcripts land in `~/.claude-personal/projects/` and never appear in
Sessions, Insights or Analytics. Symmetrically, settings profiles are written to
`~/.claude/settings_<slug>.json` and the default profile is synced into
`~/.claude/settings.json` — Agento writes into one config dir while running the
agent against another.

### C · The daemon drops the variable entirely

`internal/daemon/templates/agento.service.tmpl` and the launchd plist propagate
only `PATH`, `PORT`, `AGENTO_DATA_DIR` and `AGENTO_SKIP_UPDATE_CHECK`. Under
`agento service install` there is no way to pass `CLAUDE_CONFIG_DIR` at all —
the most likely reason the reporter saw no effect, since a terminal
`CLAUDE_CONFIG_DIR=… agento web` beside an installed service is a different
process from the one serving the UI.

---
## WHAT

**Summary:** Support multiple Claude Code config dirs — the sessions surface
indexes **all** of them, and each agent run targets **one**, chosen globally or
per agent.

**Scope:**
- A config-dir *set* for reading: `$HOME/.claude`, plus `CLAUDE_CONFIG_DIR` when
  set, plus any dirs added in Settings. The scanner walks the union; every
  cached row records which dir it came from.
- A config-dir *choice* for running: a global default (env-locked by
  `CLAUDE_CONFIG_DIR`, HTTP 409 per the existing convention) with an optional
  per-agent override, so a work agent and a personal agent can be live at once.
- Fix the `--settings` injection so it resolves inside the run's own config dir
  instead of always `$HOME/.claude/settings.json`.
- Propagate `CLAUDE_CONFIG_DIR` through the systemd unit and launchd plist.
- An account / config-dir filter on the sessions list, as a peer to the existing
  project filter.
- **Candidate discovery** so the union is not purely manual: Settings offers any
  sibling dir it finds beside `$HOME/.claude` that looks like a Claude config dir
  (matches `$HOME/.claude*` and contains a `projects/` subdirectory), which the
  user ticks to include. Suggested, never auto-enabled — discovery without
  guessing, and dirs living elsewhere are still added by hand.

**Affected Components:**
- `internal/config/claude_dirs.go` *(new)* — the set and the run-default resolver
- `internal/config/settings.go` — `ClaudeConfigDir` (run default) and
  `ClaudeConfigDirs` (extra dirs to index) + validation
- `internal/config/agent.go` — per-agent `ClaudeConfigDir` override
- `internal/config/profiles.go` — settings paths resolve within a given dir
- `internal/agent/runner.go` — `claude.WithEnv` for the run's dir; dir-aware
  `--settings`
- `internal/claudesessions/scanner.go` — `ClaudeHomes() []string`; the walk fans out
- `internal/claudesessions/session_query.go` — `config_dir` filter + facet
- `internal/daemon/templates/` — both service templates
- `frontend/` — Settings (dir list + run default), agent form (override),
  sessions filter

**Acceptance Criteria:**
- [ ] With two config dirs configured, one scan indexes transcripts from **both**;
      analytics totals, the cost-by-model breakdown and the leaderboards cover
      the union, and each session row reports the dir it came from.
- [ ] Filtering the sessions list by config dir returns exactly that dir's
      sessions, and the facet counts agree with the filtered set.
- [ ] An agent with `claude_config_dir: ~/.claude-personal` spawns `claude` with
      `CLAUDE_CONFIG_DIR=$HOME/.claude-personal` **and**
      `--settings $HOME/.claude-personal/settings.json` (or no `--settings` when
      absent), while an agent without the override uses the global default — both
      asserted by capturing the built `[]claude.Option`.
- [ ] `PUT /api/settings` setting a run default that differs from a set
      `CLAUDE_CONFIG_DIR` returns **409** and leaves the stored value unchanged.
- [ ] A configured dir that is missing or unreadable at scan time is skipped with
      a logged warning and **no cached rows are deleted** for it.
- [ ] `agento service install` writes a unit/plist carrying `CLAUDE_CONFIG_DIR`
      when it is set in the installing environment.
- [ ] A sibling dir such as `~/.claude-personal` containing `projects/` is
      offered as a candidate in Settings, is **not** indexed until confirmed, and
      a sibling without `projects/` is not offered at all.

**Out of Scope:**
- Merging or de-duplicating sessions that appear under two dirs — session ids
  are UUIDs, so they are genuinely distinct sessions.
- Any change to how Claude Code itself resolves credentials.

## WHY

**Business Value:** Running more than one Claude Code account is ordinary for
anyone separating work from personal usage, and Claude Code supports it out of
the box via `CLAUDE_CONFIG_DIR`. Agento is the only component in that setup that
cannot follow, which makes it unusable for one of the two accounts.

**Problem Statement:** The config dir is hardcoded in three independent places,
so switching accounts the documented way yields a silently inconsistent system
rather than an error: the agent authenticates as one account while being handed
the *other* account's settings file (A); its transcripts vanish from Sessions,
Insights and Analytics because the scanner looks in the wrong tree (B); and under
`agento service` the variable is not passed at all, so the switch appears to do
nothing (C). Nothing logs a warning — the sessions list simply looks empty.

**Expected Impact:** Multi-account users can use Agento at all. Critically, the
analytics surface stops reporting per-account subtotals as if they were totals:
"how much have I spent", the cost-by-model breakdown and the leaderboards become
correct across every account on the machine. No behaviour change for the
single-account default, which resolves to `$HOME/.claude` — today's hardcoded
value is the resolver's fallback.

**Context:**
- **Reading is a set, running is a choice** — the two subsystems have genuinely
  different needs, and this issue treats them separately on purpose. Analytics
  is retrospective, so restricting it to one dir at a time would hide half the
  history behind a global switch. Execution targets one account per run, so it
  takes a choice with a per-agent override.
- A **single-resolved-dir** design was refined first and rejected on inspection.
  It required purge-and-rescan semantics — detecting config-dir drift in
  `claude_cache_metadata` and deleting rows from the previous dir — which was the
  riskiest part of the plan and exists *only* because of the single-dir framing.
  The union has no drift to detect and nothing to purge.
- Two schema facts make the union cheap, both verified:
  `claude_session_cache` is `PRIMARY KEY (session_id, project_path)` with
  `file_path` already absolute (`internal/storage/sqlite.go:110`), so rows from
  two dirs cannot collide; and `diffDiskAndCache`
  (`internal/claudesessions/scanner.go:768`) keys on the absolute file path, so
  insert/update/delete detection is already dir-agnostic once the walk fans out.
- The env var already reaches the subprocess correctly
  (`claude-agent-sdk-go@v0.3.1` `claude/process.go:522`) — this is not a
  propagation bug, and `buildEnv` needs no change.
- Related: #174 (`agento service` install), #239 (Data & Analytics settings tab).

## HOW

**Technical Approach:** Introduce two resolvers in `internal/config` — the
package every layer may import, the same placement rationale as the idle-gap
bounds. `ClaudeConfigDirs() []string` is the set the scanner walks;
`ClaudeRunConfigDir(agentCfg) string` is the dir one run targets. The scanner's
three `filepath.Join(ClaudeHome(), "projects")` sites become a fan-out over the
set, and each cached row gains the dir it came from so the list can filter on it.
The run side uses the SDK's existing `WithEnv` to set `CLAUDE_CONFIG_DIR` for the
subprocess, which `buildEnv` already honours over the inherited parent value.

**Implementation Details:**
- `internal/config/claude_dirs.go` *(new)* — `DiscoverCandidateDirs() []string`
  (glob `$HOME/.claude*`, keep entries that are directories containing
  `projects/`, minus those already configured), `ClaudeConfigDirs() []string`
  (default ∪ env ∪ configured, deduped on `filepath.Clean` of the absolute path,
  default always first) and `ClaudeRunConfigDir(*AgentConfig) string`
  (per-agent override → global setting → env → `$HOME/.claude`), plus the
  process-wide snapshot on the `dataSettings` model, since the readers have no
  settings dependency and must agree.
- `internal/config/settings.go` — `ClaudeConfigDir string` (run default) and
  `ClaudeConfigDirs []string` (extra dirs to index, JSON column exactly as
  `HiddenProjects`); validation expands `~`, requires an existing directory,
  rejects a file.
- `internal/config/agent.go` — `ClaudeConfigDir string \`yaml:"claude_config_dir"
  json:"claude_config_dir"\`` on `AgentConfig`; empty means "use the global
  default".
- `internal/config/profiles.go` — `ClaudeSettingsDirPath` /
  `ClaudeSettingsJSONPath` / `LoadProfileFilePath` take the dir as a parameter
  instead of deriving `$HOME/.claude`. This is the fix for bug A, since
  `LoadProfileFilePath("")` is what injects the wrong `--settings` today.
- `internal/service/chat_service.go:318` and `internal/scheduler/executor.go:224`
  — pass the run's dir when resolving the settings path; skip `--settings`
  entirely when the resolved file does not exist.
- `internal/agent/runner.go` — in `appendSettingsOpts`, add
  `claude.WithEnv(map[string]string{"CLAUDE_CONFIG_DIR": dir})` whenever the run's
  dir differs from the process default.
- `internal/claudesessions/scanner.go` — replace `ClaudeHome() string` with
  `ClaudeHomes() []string`; fan out the walks at `scanner.go:283,391,1789`, the
  todos loader (`scanner.go:2019`), `projects.go:108` and `journey.go:190`.
  A dir that fails to walk is skipped and its cached rows are excluded from the
  diff's delete set rather than treated as absent.
- `internal/claudesessions/session_query.go` — `config_dir` clause in
  `buildFilter` and a `config_dir` facet; mirror any per-session definition
  change in `frontend/src/lib/sessionMetrics.ts` and the shared
  `testdata/session_metric_vectors.json` if the filter touches a derived figure.
- **Migration** — `config_dir` on `claude_session_cache` and
  `claude_subagent_cache`, plus the two settings columns. Backfill `config_dir`
  to `$HOME/.claude`, not empty, for the same reason migration 25 defaults
  `idle_threshold_ms` to 600000: it is not a guess, and an empty default would
  force every install to re-read a corpus it already has. Bump the hardcoded
  migration count in `sqlite_test.go`.
- `internal/api/settings.go` — 409 `EnvLockedError` when the incoming run default
  differs from a set `CLAUDE_CONFIG_DIR`; on a change to the dir *set*, call
  `EnsureScan()` as `applyDataSettings` does, so a newly added dir is indexed
  without a restart.
- `internal/daemon/templates/agento.service.tmpl` + the plist — emit
  `CLAUDE_CONFIG_DIR` when the installing environment has it; extend the template
  data struct and `systemdEnv` / `xmlEscape` the value.
- Frontend — Settings → Claude: the dir list (add/remove, env-provided entries
  read-only) and the run default; agent create/edit: the optional override;
  sessions list: the config-dir filter beside the project filter. Mirror the new
  fields in `frontend/src/types.ts` and `frontend/src/lib/api.ts`.

**Testing Strategy:**
- Unit: `ClaudeConfigDirs` dedup/ordering and `ClaudeRunConfigDir` precedence
  (per-agent > setting > env > home); validation rejections.
- Unit: `buildSDKOptions` emits the right `CLAUDE_CONFIG_DIR` and the right
  `--settings` for an overriding agent and a non-overriding one — the direct
  regression tests for bug A.
- Integration: two `synthcorpus` trees under different dirs; one scan indexes
  both, totals equal the sum, and `config_dir` attribution is correct.
- Integration: a configured dir made unreadable mid-run deletes nothing.
- API: 409 on an env-locked run-default update; 200 + rescan when unlocked.
- `internal/daemon` render tests assert the variable in both templates.

**Documentation / Configuration:** `CLAUDE.md` — the Environment section (add
`CLAUDE_CONFIG_DIR`), the `internal/config` and `internal/claudesessions` notes
(`ClaudeHome()` is gone; the scanner walks a set), and the Agent Configuration
section (the new per-agent field). `docs/` — a short multi-account guide. No new
dependencies.

**Files to Create:**
- `internal/config/claude_dirs.go`, `internal/config/claude_dirs_test.go`

**Files to Modify:**
- `internal/config/settings.go`, `internal/config/profiles.go`,
  `internal/config/agent.go`
- `internal/agent/runner.go`, `internal/service/chat_service.go`,
  `internal/scheduler/executor.go`
- `internal/claudesessions/scanner.go`, `projects.go`, `journey.go`,
  `session_query.go`
- `internal/storage/sqlite.go`, `internal/storage/sqlite_settings_store.go`,
  `internal/storage/sqlite_test.go`
- `internal/api/settings.go`
- `internal/daemon/templates/agento.service.tmpl`,
  `internal/daemon/templates/*.plist*`, `internal/daemon/render.go`
- `frontend/src/components/ClaudeSettingsTab.tsx`,
  `frontend/src/pages/AgentCreatePage.tsx`, `AgentEditPage.tsx`,
  `ClaudeSessionsPage.tsx`, `frontend/src/types.ts`, `frontend/src/lib/api.ts`
- `CLAUDE.md`

**Risks:**
- *An unreadable or unmounted dir wipes its cached rows.* Its files vanish from
  the on-disk set and `diffDiskAndCache` would mark every row deleted. → Skip a
  dir that fails to walk and exclude its rows from the delete set; covered by a
  dedicated test. This is the single most important guard in the change.
- *Adding a large second dir makes the next scan slow.* → The scan is already
  parallel-read/batched-write and asynchronous (`ListPage` never waits), and the
  status endpoint reports `files_done`/`files_total`, so it surfaces as progress
  rather than an empty list.
- *Aggregate figures shift the moment a second dir is added.* Expected, but it
  will look like a regression. → Note it in the release notes and show the
  config-dir dimension in the UI so the jump is attributable.
- *Per-agent override diverges from the settings profile it is paired with.* →
  Skip `--settings` when the resolved file is absent in that dir rather than
  passing a path that does not exist.

**Complexity:** High — three subsystems, a schema migration and a new filter
dimension. Lower risk than the single-dir alternative despite comparable size,
because nothing is purged and the diff logic is already dir-agnostic.

---
**Refined by:** AI Issue Refiner · **Date:** 2026-08-12 · **Status:** Ready

vibexp-agent:
Closes #242

Claude Code keeps credentials, projects and settings under `CLAUDE_CONFIG_DIR`, and running two accounts side by side by pointing it at a second directory is ordinary. Agento hardcoded `$HOME/.claude` in three independent places, so switching accounts the documented way produced a silently inconsistent system — the agent authenticated as one account while being handed the *other* account's settings file, its transcripts vanished from Sessions and Analytics, and under `agento service` the variable was dropped entirely.

## What

**Reading is a set, running is a choice.** The two subsystems have genuinely different needs and are treated separately.

- **The scanner walks every configured config dir into one corpus**, so "how much have I spent", the cost-by-model breakdown and the leaderboards cover every account on the machine rather than whichever is selected. Each row records the dir it came from, and the sessions list gains an account filter (shown only when the corpus actually spans more than one).
- **Agent runs target one dir** — a global default with an optional per-agent override, so a work agent and a personal agent can be live at once. Resolved in `appendSettingsOpts`, which is what makes chat, scheduled tasks, Telegram triggers and the CLI all honour it; two of those populate no settings fields at all.
- **`--settings` resolves inside the run's own dir** and is omitted when that file does not exist, instead of always passing `$HOME/.claude/settings.json`. This is the original defect.
- **Both service templates propagate `CLAUDE_CONFIG_DIR`**, which the daemon dropped.
- **Candidate discovery** so the union is not purely manual: Settings offers a `$HOME/.claude*` sibling containing `projects/`, suggested and never auto-enabled.

## Why these decisions

**Union, not one-dir-at-a-time.** Analytics is retrospective. Restricting it to a single dir would hide half the history behind a global switch, and a single-resolved-dir design was refined first and rejected: it required purge-and-rescan semantics that the union does not need at all.

**Two hazards the union creates**, both load-bearing and both tested:

1. **A dir that cannot be listed must not have its rows deleted.** "No file on disk" and "we could not look" are indistinguishable to the diff, and an unmounted drive would otherwise wipe an account's corpus — taking `custom_title` and `is_favorite`, the two user-owned columns a rescan deliberately preserves, with it. Protection is per *project*, not per config dir, so one unreadable directory does not freeze deletion for the whole corpus.
2. **A session present in two dirs is one session.** The ordinary way to set up a second account is to copy the first dir, which duplicates every session id under the same project path. Since `claude_session_cache` is keyed on `(session_id, project_path)` while `file_path` is only a non-unique index, indexing both would double that corpus's tokens and cost *and* leave the losing path oscillating between insert and delete, re-firing `EventSessionDiscovered` every scan.

## Deviations from the issue's HOW

Both are cases of the codebase winning over the written plan:

| Issue said | Reality | Result |
| --- | --- | --- |
| `PUT /api/settings` returns **409** `EnvLockedError` | Neither exists in the settings path — that is the *monitoring* manager's convention. `UserSettings` uses a `locked` map and `Update` errors map to **400** | Followed the real convention; a test pins the status |
| Setting lives in **Settings → Claude** | `ClaudeSettingsTab` edits `settings.json` via profiles and never touches `UserSettings` | Run default → General tab (beside `default_working_dir`, which already carries the env-lock pattern); indexed dirs → Data & Analytics (beside hidden projects) |

## Review

Four independent reviewers over two rounds, across all eight dimensions. Five real defects were found and fixed, none of which CI would have caught:

- `CurrentScannerVersion` was not bumped, so `config_dir` would have stayed empty on every existing row — the account filter matching none of an upgraded corpus.
- Validation ran against the env-*pinned* value, so a `CLAUDE_CONFIG_DIR` naming a not-yet-created directory made **every** settings save fail, on a field the UI renders read-only.
- The analytics memo key omitted the config-dir set, so removing a dir kept serving its cost and tokens — removal deliberately triggers no rescan, so `last_scanned` never moves.
- A per-agent dir was accepted unvalidated; a relative one is `stat`-ed against the server's working directory but resolved by the subprocess against its own, so a repo-resident `settings.json` — which carries `hooks` and `env` — could reach `--settings`. Absoluteness is now enforced at every resolver, after round 2 found the first fix was defeated by its own fallback.
- A genuine path traversal: `GetSessionJourney` validated its session id but `GetSessionDetail` never did, and this refactor made them share one helper. CodeQL flagged it independently.

Two gaps were proven by **mutation**, not inspection: deleting `addConfigDirScope` or `ApplyClaudeDirs` passed the entire suite. Both now have tests that fail without them, re-verified by re-running the same mutation. One new test was itself caught this way — it rewound the scanner version *relative* to the constant, so it survived its own fix being reverted.

## Compatibility

Migration 27 adds `config_dir` defaulting to empty, which readers interpret as the default dir, so no install has to re-read anything to keep the figures it already had. The `CurrentScannerVersion` bump to 13 then re-stamps every row on the next scan — a one-time full re-read, which is exactly what that constant exists for.

## Deferred

- #244 — `agento ask` ignores the stored global default (fixing it means giving a CLI path a DB dependency it does not have)
- #245 — a claim shift re-fires `EventSessionDiscovered`; bounded to spurious reprocessing, no data loss

## Note on the dismissed CodeQL alert

`ValidateClaudeConfigDir` stats a directory the operator named, which is the feature — Agento must let a user point it at their own config dir, including one outside `$HOME`. Containing it to `$HOME` would remove that capability while closing nothing: `CLAUDE_CONFIG_DIR` bypasses this validator entirely (it is env-locked, never validated), and `internal/api/filesystem.go` already performs a strictly stronger `os.ReadDir` on a user-supplied path. Agento is a single-user desktop app with intentionally no authentication, so there is no privilege boundary here — the same caller can create a bypass-permission agent and run arbitrary Bash.

The genuine exposure surfaced while reviewing this PR is the browser as an attack path — a cross-origin simple `POST` needs no preflight and is executed, DNS rebinding defeats CORS regardless of a loopback bind, and the server currently binds `0.0.0.0`. That is filed as **#246** and is unrelated to this change.


vibexp-agent:
Deferred to #244 and #245 — both filed from this PR's review and cross-linked back to #242.

PR #253, opened 2026-08-13
fix(claude-sessions): reload the session list when the account changes

### LINKED ISSUE #252: fix(claude-sessions): account switcher sets the value but never reloads the sessions list
## Summary

On the **Claude Sessions** page, the account (Claude config dir) switcher in the toolbar updates its displayed value when you pick a different account, but the sessions list underneath is **not reloaded**. It keeps showing the previous account's sessions until *some other* filter is changed.

## Steps to reproduce

Requires a corpus spanning more than one Claude config dir, since the account switcher only renders when `configDirs.length > 1`.

1. Open **Claude Sessions**.
2. Open the account dropdown (e.g. `~/.claude-personal`) and select a different account (e.g. `~/.claude`).
3. Observe: the dropdown now shows the new account, but the list, the toolbar counter (`N sessions · … tokens · $…`) and the day groups are unchanged.
4. Now change any *other* filter (e.g. the project dropdown). The list reloads — and only now reflects the account picked in step 2.

Image

## Expected

Selecting an account immediately reloads the list (and the facet totals) scoped to that account.

## Actual

The value is set but no refetch happens. The account change is only picked up when another filter change happens to trigger a reload.

## Root cause

In `frontend/src/pages/ClaudeSessionsPage.tsx`, the `filters` object is assembled in a `useMemo` that reads `filterConfigDir`:

```tsx
const filters = useMemo<SessionFilters>(() => {
  return {
    ...NO_FILTERS,
    ...advanced,
    project: filterProject,
    configDir: filterConfigDir, // read here
    ...
  }
}, [
  advanced,
  filterProject,          // present
  // filterConfigDir      <-- MISSING
  debouncedSearch,
  filterFavorites,
  timePreset,
  customFrom,
  customTo,
  drilldownWindows,
])
```

`filterConfigDir` is used in the body but missing from the dependency array. `useSessionPages(filters, sort)` keys its entire fetch off the `filters` object identity (`filters` → `queryKey` → `reload` → the fetch effect), so when only the account changes, `filters` does not recompute, its identity is unchanged, and nothing refetches. Changing the project (which *is* a dependency) recomputes `filters`, which then reads the already-updated `filterConfigDir` — which is exactly why "change account, then change project" works.

This slipped through CI because the ESLint config extends `react-hooks`'s recommended rules, where `react-hooks/exhaustive-deps` is a **warning**, and the lint script is a bare `eslint .` (no `--max-warnings 0`), so the warning never fails the build.

## Proposed fix

Add `filterConfigDir` to the `filters` `useMemo` dependency array, plus a regression test that switches the account and asserts the list refetches scoped to the new `config_dir`.

amims71:
## Summary

The account (Claude config dir) switcher on the **Claude Sessions** page set its value but never reloaded the list — you had to change another filter (e.g. the project) before the account took effect. Fixes the dropdown reported in #252.

## Root cause

`frontend/src/pages/ClaudeSessionsPage.tsx` assembles the `filters` object in a `useMemo` that reads `filterConfigDir` but omitted it from the dependency array. `useSessionPages(filters, sort)` keys its whole fetch off the `filters` object identity (`filters` → `queryKey` → `reload` → the fetch effect), so an account-only change never recomputed `filters` and nothing refetched. Changing the project — which *is* a dependency — recomputed `filters`, picking up the already-set account, which is exactly why "change account, then change project" worked.

Not caught by CI because `react-hooks/exhaustive-deps` is a **warning** in the recommended ruleset and the lint script is a bare `eslint .` (no `--max-warnings 0`).

## Fix

- Add `filterConfigDir` to the `filters` `useMemo` dependency array (one line).
- Add `ClaudeSessionsPage.test.tsx`: a regression test that renders the page, switches the account, and asserts the list refetches scoped to the new `config_dir`. It **fails before the fix and passes after**.

## Testing

- `npm run test` — 15 files, 198 tests pass (including the new one)
- `npm run typecheck` — clean
- `npm run lint` — 0 errors (only pre-existing `react-refresh` warnings; the `exhaustive-deps` warning on this file is now gone)
- `npx prettier --check` on both changed files — clean

## Docs

No documentation change needed — this is a frontend behavioural fix with no API or architecture change.

Closes #252

https://claude.ai/code/session_011DEe9NxrnSq48rC9PYYSXd

### DIFF
--- a/frontend/src/pages/ClaudeSessionsPage.test.tsx
+++ b/frontend/src/pages/ClaudeSessionsPage.test.tsx
@@ -1,0 +1,129 @@
+import { describe, it, expect, vi, beforeAll, beforeEach } from 'vitest'
+import { render, screen, waitFor } from '@testing-library/react'
+import userEvent from '@testing-library/user-event'
+import { MemoryRouter } from 'react-router-dom'
+import ClaudeSessionsPage from './ClaudeSessionsPage'
+import { claudeSessionsApi } from '@/lib/api'
+import type { ClaudeSessionPage, ClaudeSessionFacets, ClaudeSessionStatus } from '@/types'
+
+vi.mock('@/lib/api', () => ({
+  claudeSessionsApi: {
+    list: vi.fn(),
+    facets: vi.fn(),
+    projects: vi.fn(),
+    status: vi.fn(),
+    refresh: vi.fn(),
+    toggleFavorite: vi.fn(),
+  },
+}))
+
+// Two indexed accounts, so the account switcher renders at all — it only
+// appears when the corpus spans more than one Claude config dir.
+const ACCOUNT_A = '/home/me/.claude'
+const ACCOUNT_B = '/home/me/.claude-personal'
+
+const emptyPage: ClaudeSessionPage = { items: [], next_cursor: '', has_more: false }
+
+const facets = (overrides: Partial<ClaudeSessionFacets> = {}): ClaudeSessionFacets => ({
+  total: 0,
+  total_tokens: 0,
+  total_cost_usd: 0,
+  token_p90: 0,
+  models: [],
+  permission_modes: [],
+  config_dirs: [ACCOUNT_A, ACCOUNT_B],
+  has_favorites: false,
+  has_prs: false,
+  ...overrides,
+})
+
+const idleStatus: ClaudeSessionStatus = {
+  costs_stale: false,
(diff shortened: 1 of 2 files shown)


vibexp-agent (approved):
Thanks @amims71 — clean root-cause analysis and the regression test genuinely fails without the one-line fix. 🙏

PR #255, opened 2026-08-13
feat(claude-sessions): add a Refresh button to the session detail page

### LINKED ISSUE #254: feat(claude-sessions): add a Refresh button to the session detail page
## Summary

The **Claude session detail page** has no way to pull fresh data from Claude. To update the figures for an active session you have to leave the page, go back to the sessions list, click its **Refresh**, and navigate back into the session. The list already has a Refresh button that does the right thing; the detail page should have the same.

## Why a plain browser reload isn't enough

`GET /claude-sessions/{id}` re-reads the transcript **live**, so the input/output/cache tokens, duration and messages update on any reload. But two figures come from the **cached scan**, not the live re-read:

- **Cost** (`GetSummary(id).Cost` / `SubagentCost`) — stored per session by the scanner.
- The **delegated sub-agent** tokens/cost ("Includes … from N delegated sub-agents") — from `ListSubagents(id)`.

So for an active session those stay stale until a scan runs. The list's Refresh handles this because it triggers the server-side rescan (`POST /claude-sessions/refresh` → invalidate cache + `EnsureScan`, which re-reads Claude's JSONL), then reloads. The detail page needs the same, not just a page refresh.

## Proposal

Add a **Refresh** button to the session detail page header (alongside Journey / Resume), mirroring the list's behaviour exactly:

1. `POST /claude-sessions/refresh` — kick off the background rescan that re-reads Claude's transcripts.
2. Re-fetch the detail (`GET /claude-sessions/{id}`), so the freshly-scanned Cost & sub-agent figures show alongside the live token counts.

The `RefreshCw` icon spins while in flight; the button is disabled during the request.

## Scope

Frontend-only — both endpoints already exist. No API, type, or docs changes.

amims71:
## Summary

Adds a **Refresh** button to the Claude session detail page, so you can pull fresh data from Claude without leaving for the sessions list and coming back. It mirrors the list's Refresh exactly — it is not a page reload. Implements #254.

## Why not just reload the page

`GET /claude-sessions/{id}` re-reads the transcript **live**, so input/output/cache tokens, duration and messages refresh on any reload. But **Cost** and the **delegated sub-agent** tokens/cost come from the **cached scan** (`GetSummary(id)` / `ListSubagents(id)`), so for an active session those stay stale until a scan runs. The list's Refresh handles this by triggering the server rescan; the detail page now does the same.

## Behaviour

`RefreshCw` icon + label in the header action row (⭐ · **Refresh** · Journey · Resume), styled like the Journey button:

1. `POST /claude-sessions/refresh` → invalidate the cache + `EnsureScan` (re-reads Claude's JSONL).
2. Brief pause, then re-fetch the detail (`GET /claude-sessions/{id}`).

Icon spins while in flight; button disabled during the request. This is a deliberate 1:1 match of the list page's `handleRefresh`.

## Scope

Frontend-only — both endpoints already exist. No API, type, or docs changes.

## Testing

- Added `ClaudeSessionDetailPage.test.tsx` (TDD, red first): renders the page, clicks Refresh, asserts `refresh()` fires and the detail re-fetches.
- `npm run test` — 16 files, **199 tests pass**.
- `npm run typecheck` / `npm run lint` (0 errors) / `prettier --check` — clean.
- Verified live in a local build on the real corpus: clicking Refresh returned `202` from the rescan endpoint and the header's Cost + Insights updated to fresh values (not just the live token counts).

Closes #254

https://claude.ai/code/session_011DEe9NxrnSq48rC9PYYSXd

### DIFF
--- a/frontend/src/pages/ClaudeSessionDetailPage.test.tsx
+++ b/frontend/src/pages/ClaudeSessionDetailPage.test.tsx
@@ -1,0 +1,130 @@
+import { describe, it, expect, vi, beforeAll, beforeEach } from 'vitest'
+import { render, screen, waitFor } from '@testing-library/react'
+import userEvent from '@testing-library/user-event'
+import { MemoryRouter, Routes, Route } from 'react-router-dom'
+import ClaudeSessionDetailPage from './ClaudeSessionDetailPage'
+import { claudeSessionsApi, insightsApi } from '@/lib/api'
+import type { ClaudeSessionDetail } from '@/types'
+
+vi.mock('@/lib/api', () => ({
+  claudeSessionsApi: {
+    get: vi.fn(),
+    refresh: vi.fn(),
+    toggleFavorite: vi.fn(),
+    updateTitle: vi.fn(),
+    continue: vi.fn(),
+  },
+  insightsApi: {
+    getSession: vi.fn(),
+  },
+}))
+
+const SESSION_ID = 'sess-1'
+
+const zeroUsage = {
+  input_tokens: 0,
+  output_tokens: 0,
+  cache_read_tokens: 0,
+  cache_creation_tokens: 0,
+  cache_creation_5m_tokens: 0,
+  cache_creation_1h_tokens: 0,
+}
+const zeroCost = {
+  input_usd: 0,
+  output_usd: 0,
+  cache_read_usd: 0,
+  cache_write_usd: 0,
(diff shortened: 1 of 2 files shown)


vibexp-agent (approved):
Thanks @amims71 — a faithful 1:1 mirror of the list's Refresh, frontend-only, with a test to match. 🙏

PR #281, opened 2026-08-14
feat(desktop): port the Claude Agent SDK to Rust

### LINKED ISSUE #269: Port: the Claude Agent SDK to Rust
## Scope

Reimplement `github.com/shaharia-lab/claude-agent-sdk-go` in Rust — **our own OSS project**, checked out at `~/Projects/claude-agent-sdk-go` (it carries its own `CLAUDE.md` and `GAP_ANALYSIS.md`). Read that as the spec rather than inferring the protocol from agento's call sites.

**Scale: ~20 non-test Go files, ~4,590 lines** — smaller than the analytics endpoint. Its "highest risk" reputation is about failure *mode*, not volume: a protocol mistake shows up as a chat that silently hangs, not as a failing diff.

It is **not an API client** but a subprocess protocol client: spawns the `claude` CLI, speaks stream-json over stdin/stdout plus a control protocol. No model inference to reimplement.

## Surface agento uses

- session lifecycle: `NewSession`, `Query`, `Stream`, `Result`
- options: thinking modes, allowed tools, permission modes, setting sources
- control protocol: initialize handshake, permission-callback round trips, interrupt
- **`StartInProcessMCPServer`** — every integration is one of these

## Use the prior art

The protocol details that cost real time are already written down in this project's learning logs: the initialize-handshake reorder (**PR #47**), `sdkMcpServers` having to be an array of strings (**#44**), `terminal_reason` (**#50**). A port from the Go source alone would rediscover all three. That repo also proved the test technique: a **scripted fake CLI** for deterministic spawn ordering — so this is testable with no real `claude` binary and no API key.

## Parity bar differs

There is no JSON response to diff. Hold the **SSE stream** identical: raw CLI JSON lines passed through verbatim, plus two synthetic events (`user_input_required`, `permission_request`).

## Not blocked — parallel-safe, and the single best task to run alongside everything else (different repo, zero file overlap).
---
**Verify:**
```bash
```

vibexp-agent:
Closes #269.

Targets `desktop`; `main` untouched.

## What

`github.com/shaharia-lab/claude-agent-sdk-go` reimplemented in Rust as `desktop/src-tauri/src/claude/`. It is **not an API client** but a subprocess protocol client: it spawns the `claude` CLI, speaks stream-json over stdio, and multiplexes a control protocol down the same pipes. There is no model inference to reimplement and no API key — the CLI's own sign-in is the credential.

Ported surface, all of what Agento uses across its 21 call sites:

- session lifecycle — `query`, `run`, `Session`, `Stream`
- every option, across both channels it splits over (CLI flags, and the initialize message)
- the control protocol — handshake, permission and hook round trips, interrupt, and the `set_*`/mcp requests
- the wire types and the line parser
- `start_in_process_mcp_server` — what all seven integrations are built on

## Why now, ahead of anything that calls it

Nothing routes to it yet: the runner, chat service and SSE handler are still Go. It lands first because **phases 4 and 5 both need it** — every integration is one of its MCP servers, so phase 4 cannot move before it does — and because its correctness is testable in isolation, against a scripted CLI, in a way it stops being once a live route depends on it.

## How it was verified

There is no JSON response to diff, so `parity-instance.sh` has nothing to say here. The bar is the **SSE stream**: raw CLI lines passed through verbatim plus Agento's two synthetic events, which is why `Event::raw` keeps every message's bytes.

The tests are the technique the Go SDK proved — a **scripted fake CLI** (`tests/claude_sdk.rs`), a Python program that logs every stdin line and replies to order. That is the only way to check things that are properties of a *sequence* rather than of a function. **106 tests: 90 unit, 16 driving the real spawn path.**

Local gates, all green: `cargo fmt --check`, `cargo clippy --all-targets -- -D warnings`, `cargo test`, `cargo build --release`.

## Four protocol facts carried over rather than re-discovered

Each is silent when wrong, which is why they are in the code as comments and in `desktop/CLAUDE.md`:

- **The handshake order.** Reader task live → register the request id → write `initialize` → *block* on the acknowledgement → only then the first user message. A `control_response` cannot be routed before something reads stdout, and MCP servers, agents and hooks are configured during the ack. Wrong ordering races; it does not fail.
- **`sdkMcpServers` is never sent.** The CLI accepts only an array of strings there, and a rejection fails the *entire* initialize — silently taking hooks, agents, the system prompt and the output format with it.
- **`control_response` routes on the nested `response.request_id`** and hands back the *innermost* payload. No top-level fallback; an absent inner payload is a success with no data.
- **Every inbound `control_request` must be answered** or the CLI hangs, including ones we only acknowledge. `can_use_tool` with no handler is answered with an **error, never an allow** — fail closed.

## Four places a mechanical port would have been wrong

Documented at each site:

| | |
|---|---|
| `lenient.rs` | Go keeps the fields that decoded and reports the error; serde aborts the struct. Reproduced, because a CLI that reshapes one field must not blank a message. |
| `Stream` / `StreamControl` | Go's single `*Stream` is safe from any goroutine; reading a channel in Rust needs `&mut self`, so the control half is a separate clonable handle. |
| async callbacks | Go blocks a goroutine inline on the reader. Blocking a runtime worker would be a bug, so handlers return futures awaited in the same position — same semantics, no pinned thread. |
| handle lifetimes | Go cancels via `context.Context`. Dropping a `Stream` shuts the subprocess down, so a cancelled chat cannot orphan one. |

## One scope call worth reviewing

`start_in_process_mcp_server` takes an `McpService` trait rather than an `rmcp` server. Go hands its MCP SDK's `*mcp.Server` to that SDK's streamable-HTTP handler; the equivalent here would pin `rmcp`'s API to satisfy **no caller**, since nothing has an MCP server to host until phase 4. The seam sits exactly where Go's does — the SDK owns the listener, the caller owns the tools — and an `rmcp` adapter is one impl away. Say so if you would rather take the dependency now.

## Dependencies added

`tokio` gains `process`/`io-util`/`io-std`/`sync` (the sidecar goes through tauri-plugin-shell, which only launches declared sidecars; the agent CLI is an arbitrary executable on `PATH`), plus `uuid` for request ids and `libc` on unix for the SIGTERM step of shutdown.

### DIFF
--- a/desktop/CLAUDE.md
+++ b/desktop/CLAUDE.md
@@ -110,6 +110,15 @@ src-tauri/src/
   sidecar.rs     spawn + health-wait + shutdown of the Go server
   proxy.rs       axum reverse proxy; route_is_native() is the porting switch
   menu.rs        native menu → menu://action events
+  claude/        the Claude Agent SDK, ported from Go (phase 5's foundation)
+    process.rs   spawn, the control protocol, and the initialize handshake
+    client.rs    Stream (events) + StreamControl (interrupt, set_model, …)
+    session.rs   persistent multi-turn conversations on one subprocess
+    options.rs   every option, and which of the two channels it travels on
+    messages.rs  the wire types and parse_line
+    permissions.rs / hooks.rs   the two callback round trips
+    mcp.rs       loopback HTTP host for in-process MCP servers
+    lenient.rs   Go's partial-decode semantics, which serde does not have
   native/        ported endpoints (phase 2+)
     mod.rs       endpoint registry, mode switch, response shaping
     gojson.rs    Go-compatible JSON encoder — read this before porting anything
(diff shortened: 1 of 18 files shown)


vibexp-agent:
Deferred to #282 — the choice of Rust MCP server implementation (`rmcp` vs the `McpService` trait this PR introduces) and the SDK's typed-tool layer (`NewTool`/`ToolServer`/`WithTools`), which depends on it. Both are phase-4 decisions: nothing has an MCP server to host until the integrations move.

Issue #289, opened 2026-08-15
Port: Claude session scan status and refresh (blocked on the scanner owning the write path)


vibexp-agent:
Follow-up from PR #290 (relates to #271).

## Scope

`GET /api/claude-sessions/status` and `POST /api/claude-sessions/refresh`, the two endpoints #271 named that could not move with the rest of it.

## Why they were deferred rather than ported

Both are about the **scan**, not about reading cached rows:

- `/status` reports `costs_stale`, `scan_in_progress`, `files_done`, `files_total` and `last_scanned_at`. The middle three are in-memory state of the scanner running inside the Go sidecar. Rust cannot observe them, and a native answer of `false`/`0` would be *actively wrong* while a Go scan runs — it would blank the "Scanning ~/.claude… 412 / 1,373 transcripts" the sessions list shows during a first run, which is the whole reason that endpoint carries progress at all.
- `/refresh` invalidates the cache metadata and starts a rescan. That is a write.

Rust deliberately does not own scanning: `native/db.rs` opens the database **read-only** precisely so two processes never write one SQLite file, and `desktop/CLAUDE.md` records that wiring the ported scanner in — and with it retiring `native/sessions::freshness_probe` — belongs to phase 3.

Porting them earlier would mean either lying about the scan or having two writers, and #271's own trap note ("porting `/status` and `/refresh` retires `freshness_probe`") is exactly the phase-3 dependency.

## Acceptance

- Rust owns the scan: `db.rs` is read-write, the ported scanner in `native/scanner/` applies its diff, and the Go sidecar no longer scans.
- `freshness_probe` and `PROBE_PATH` are deleted, along with the `Answer.probe` plumbing in `native/mod.rs` and `proxy.rs` if nothing else uses it.
- Both endpoints are native and diff byte-identical, including a `/status` observed **while a scan is in progress** — an idle instance answers `false`/`0`/`0` and diffs clean while proving nothing.

## Blocked by

Phase 3 (storage + writes). Do not start this before the scanner writes.

vibexp-agent:
Landed on `desktop` in #303 (squashed). `Closes #289` does not fire for PRs targeting a non-default branch, so closing by hand.

**All three acceptance criteria met:**
- Rust owns the scan — `native/scan.rs` orchestrates the scanner from #270, `db.rs` is read-write, and the sidecar runs with `AGENTO_SCANNER=off` (checked in `Cache.EnsureScan`, the single place a scan starts on the Go side).
[...]

PR #290, opened 2026-08-15
feat(desktop): port the session detail and project reads to Rust (#271)

### LINKED ISSUE #271: Port: session detail, status, refresh and projects
## Scope

`GET /api/claude-sessions/{id}`, `/status`, `/refresh`, `/projects`.

`GetSessionDetail` **re-reads the JSONL transcript** — this is scanner work, not a SQLite read, which is why it waits.

Porting `/status` and `/refresh` also **retires `native/sessions::freshness_probe`**: the probe exists only because reading is what keeps the corpus fresh on the Go side, and it is sidecar-dependent.

## Traps

- `/claude-sessions/projects` currently returns `decoded_path` identical to `encoded_name` — an upstream bug; match Go, then fix in `main`.
- The project filter means different things on `/claude-analytics` (`decoded_path`) and `/claude-sessions` (`project_path`).

## BLOCKED BY: the scanner port.
---
**Verify:**
```bash
```

**Blocked by:** #270

vibexp-agent:
Closes #271

## What

`GET /api/claude-sessions/{id}` and `GET /api/claude-sessions/projects` are answered by Rust.

## Scope: two of the four, and why

The issue names four endpoints. Two are ported here; **`/status` and `/refresh` are not, and that is a decision rather than an omission** — deferred to #289 and cross-linked.

Both are about the *scan*, not about reading cached rows. `/status` reports `scan_in_progress`, `files_done` and `files_total`, which are in-memory state of the scanner running inside the Go sidecar; a native answer of `false`/`0` would be **actively wrong** while a Go scan runs and would blank the "Scanning ~/.claude… 412 / 1,373 transcripts" the list shows on a first run. `/refresh` invalidates the cache and starts a scan — a write. Rust deliberately does not own scanning: `native/db.rs` opens the database read-only precisely so two processes never write one SQLite file, and `desktop/CLAUDE.md` already records that wiring the scanner in (and with it retiring `freshness_probe`, which the issue's trap note calls out) belongs to phase 3.

## How

**`sessions/detail.rs`** re-reads the session's own transcript, so the detail is correct even for a session the scanner has not reached. Four things still come from the cache, each for its own reason: `custom_title` and `is_favorite` are the only columns the user typed; the native/AI titles are read **only as a fallback**, because reading them unconditionally would blank a title Claude Code has just written; `cost` is accumulated per assistant message during a scan (#188) and a re-read has no per-message pricing context; sub-agents live in sibling transcripts.

Two asymmetries in `readSessionDetail` are reproduced deliberately — a sidechain **user** event is skipped while a sidechain **assistant** event is not, and the detail's `preview` uses a weaker rule than the scanner's turn-aware one, so the list and the detail can legitimately disagree.

**`sessions/projects.rs`** derives the list from `walk_all_disk_files` — the *same* walk a scan publishes from — rather than from Go's cold-start `walkProjects` fallback, which counts distinct session ids and includes empty project directories. Go serves the published list, so matching the publisher is the faithful choice.

## Two shared pieces fell out

- `gojson` now hosts `compact`/`compact_raw` (they were private in `chats.rs`; a `tool_use` `input` needs the same verbatim carry) and the new `null_is_zero_value`.
- `transcript::Message` carries its content **twice** — as a `Value` for the turn predicates, and as the original bytes so a `tool_use` `input` reaches the wire with its stored key order and number spelling. One decode produces both, so they cannot describe different content.

## The bug the live diff caught

Adding `uuid`/`parentUuid` to the transcript decoder **silently dropped the first user message of every conversation**: `parentUuid` is `null` on exactly the event that starts one, and `serde` rejects `null` for a `String` where Go's `json.Unmarshal` leaves `""`. A rejected field fails its struct, a failed struct drops its event, and a dropped event is simply absent.

The diff caught it at byte 112 of a 700 KB response — the session's `preview` was the wrong user message. **No unit test would have**, because every hand-written fixture had the field present. Every `#[serde(default)]` scalar in the transcript decoder now goes through `null_is_zero_value`; genuinely unparseable input still fails, as it does in Go. Written up in `desktop/CLAUDE.md`.

Also fixed: `find_session_file` now sorts the project directories, because `os.ReadDir` sorts and `std::fs::read_dir` does not, and that order decides which directory wins a session id present in two.

## Verification

- 333 unit tests green, clippy clean at `-D warnings`, `cargo fmt` clean, `go test ./desktop/parity/` green.
- `tests/parity_session_detail.rs` — live diff against a Go server built from this checkout. Sessions are picked **by property** rather than by position (sub-agents, a stored cost, linked PRs, a custom title, the most recent, the busiest) and the suite asserts it found enough distinct ones to be worth running: 4 details, 702 KB / 670 KB / **2.9 MB** / 497 KB, all identical. Both `/projects` cases identical at 8145 B.
- **The whole parity corpus re-run**, because the shared transcript decoder changed: `parity_scanner` (926 + 920 recomputed rows), `parity_sessions`, `parity_insights`, `parity_analytics` — all green.

### DIFF
--- a/desktop/CLAUDE.md
+++ b/desktop/CLAUDE.md
@@ -144,7 +144,10 @@ src-tauri/src/
     agents.rs    GET /api/agents and /api/agents/{slug}
     chats.rs     GET /api/chats and /api/chats/{id}; compact() is Go's, byte for byte
     tasks.rs     GET /api/tasks, /api/job-history and the three reads between them
-    sessions/    GET /api/claude-sessions and /facets; corpus.rs loads the lot
+    sessions/    GET /api/claude-sessions, /facets, /projects and /{id}
+      detail.rs  one session re-read from its transcript, patched from the cache
+      projects.rs the project picker's list, derived from the same walk a scan is
+      corpus.rs  loads the lot
     analytics/   GET /api/claude-analytics
       buckets.rs Go's time.Date/AddDate and the bucket walks, in the request's tz
       params.rs  from/to/project/tz, and the granularity the window picks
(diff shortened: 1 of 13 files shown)


vibexp-agent:
Deferred to #289: `GET /api/claude-sessions/status` and `POST /api/claude-sessions/refresh` are blocked on Rust owning the scan (phase 3), for the reasons in the PR description.

Issue #278, opened 2026-08-14
Delete the Go sidecar


vibexp-agent:
## Scope

The cut-over. Only after every other sub-issue is closed.

- Drop `externalBin` from `tauri.conf.json`; delete `sidecar.rs`, `scripts/build-sidecar.sh` and the CI step that builds it.
- **Turn every fallback `Err` into a real HTTP response.** While the sidecar exists, a native handler error forwards to Go; with no sidecar there is nothing to fall back to. Several handlers deliberately return `Err` today for cases Go answers with a 404 or 400 (missing agent, cursor/sort mismatch) — each needs its own status and body.
- **Wire the handlers that were written but deliberately left unclaimed.** Two exist, both following the `migrate::apply` precedent from #274 — written, unit-tested against literals captured from Go, and switched on at the cut-over rather than before it. `PUT /api/settings` is the one to check first: `desktop/parity/write_routes.json` still carries it as `deferred` under owner **#305, which is closed**, so nothing open tracks it but this bullet. Its blocker is the sidecar's own snapshot — `notificationServiceImpl.UpdateSettings` is a read-modify-write over `settingsMgr.Get()` that persists the whole `user_settings` row, so one unrelated `PUT /api/notifications/settings` silently reverts a native settings write — and that blocker dies with the sidecar. Re-run the write-routes test after wiring; it fails on a stale classification, which is what will catch anything else in this category.
- **Decide `/health` and `/metrics`, which the write-routes audit does not cover.** That file is writes-only by design (*"GET routes are deliberately out of scope"*), so nothing has checked the read surface against the seam. These two are the gap: `proxy::is_api_path` recognises both, but **no module claims either** — `monitoring::claims` matches only `/api/monitoring` and `/api/monitoring/test` — so both forward to Go today and simply stop answering when it goes.
  - `/metrics` should be **dropped deliberately, not left to 404**. Telemetry is already a declined feature here (`monitoring::DECLINED`: *"the desktop build exports no telemetry; run the agento server if you need OpenTelemetry or Prometheus"*), and the PUT/POST that would configure it are claimed purely so they answer 501 rather than reaching a sidecar that is going away. `/metrics` is the same decision and needs the same treatment. It is also **user-facing**: `frontend/src/components/MonitoringTab.tsx:258` tells the reader "Metrics are available at `/metrics`", which becomes a dead link in the desktop build — the text needs to go or be conditioned on the build.
[...]

Issue #277, opened 2026-08-14
Port: integration writes and OAuth flows


vibexp-agent:
## Scope

Integration CRUD, OAuth2 consent round trips, webhook registration, and the Telegram trigger dispatcher (`internal/trigger/`).

## Traps

- OAuth consent must open in the **user's real browser**, never the app webview.
- `POST /webhooks/telegram/{id}` is mounted at the **root**, not under `/api`, arrives with a foreign `Host`, and is authenticated by its own secret token — the `/api` guards deliberately do not apply.

## BLOCKED BY: the Claude Agent SDK port, and storage + writes.
---
**Branch:** target `desktop`, never `main`. One PR, merged before the next task starts.
**Bar:** byte-identical JSON. Any field-name, key-order, escaping or float-spelling drift is a regression.
**Verify:**
```bash
cd desktop && eval "$(./scripts/parity-instance.sh start)"
(cd src-tauri && cargo test --test live_parity -- --ignored --nocapture)
./scripts/parity-instance.sh stop
```
**Read first:** VibeXP artifact `agento-desktop-rust-port-handover-2026-08-14`, then `desktop/CLAUDE.md`.
Never diff against the server on :8990 (stale binary). Seed the scratch instance if the endpoint has no local data, or an empty response diffs clean and proves nothing.

**Blocked by:** #269, #274

_Unblocked issues in this epic can be worked in parallel._

vibexp-agent:
## Scope correction before implementing

Revalidating the HOW against the code: **most of this issue is blocked by #282, which it does not list as a blocker.** The stated blockers (#269, #274) are both merged, so it looked ready. Working the unblocked subset and stating the rest explicitly rather than half-porting it.

The port's standing rule is that a route moves only when Rust reproduces **every** effect it has. Applying that:

### Portable now — no effect Rust cannot reproduce
| Route | Why it is safe |
|---|---|
| `POST /api/integrations` | `integrationService.Create` is validation → uuid → timestamps → `store.Save`. **No registry call** — verified in full. |
| `POST /api/integrations/{id}/triggers` | Pure row write. |
| `PUT /api/integrations/{id}/triggers/{rid}` | Pure row write. |
| `DELETE /api/integrations/{id}/triggers/{rid}` | Pure row write. The dispatcher calls `ListRules` **per request** (`dispatcher.go:168`), so there is no cache to invalidate — this is what makes rule writes portable while integration writes are not. |

### Must keep forwarding, each for a stated reason
[...]

Issue #276, opened 2026-08-14
Port: chat execution and SSE streaming


vibexp-agent:
## Scope

`internal/agent/`, `internal/service/chat_service.go`, `internal/api/livesessions.go`. The hardest single route.

## Traps

- **Chat SSE is a POST response** — `EventSource` cannot be used.
- **There is no terminal event**, and `result` can arrive more than once in one request (an `AskUserQuestion` keeps the stream open past it). End the turn on stream close, never on `result`.
- **Mid-stream failures arrive as `result` with `is_error: true`**, not as an `error` event — HTTP 200 is already committed.
- **`AskUserQuestion` is answered by *denying* the tool** with the user's text in `Message`. Not a bug.
- A turn that produced no final text persists nothing — not even the user message.
- Per-session mutex serialises concurrent SSE sends.
- `--settings` resolves inside the run's Claude config dir and is omitted when absent (#242).

## BLOCKED BY: the Claude Agent SDK port, and storage + writes.
---
**Branch:** target `desktop`, never `main`. One PR, merged before the next task starts.
**Bar:** byte-identical JSON. Any field-name, key-order, escaping or float-spelling drift is a regression.
**Verify:**
```bash
cd desktop && eval "$(./scripts/parity-instance.sh start)"
(cd src-tauri && cargo test --test live_parity -- --ignored --nocapture)
./scripts/parity-instance.sh stop
```
**Read first:** VibeXP artifact `agento-desktop-rust-port-handover-2026-08-14`, then `desktop/CLAUDE.md`.
Never diff against the server on :8990 (stale binary). Seed the scratch instance if the endpoint has no local data, or an empty response diffs clean and proves nothing.

**Blocked by:** #269, #274

_Unblocked issues in this epic can be worked in parallel._

vibexp-agent:
**Spec captured, not started.** Analysis from the #274 batch, so the next session can begin cold rather than re-deriving it. Full version in the VibeXP handover `agento-desktop-rust-port-handover-2026-08-15b`.

**The four routes move together or not at all.** `/messages`, `/input`, `/permission` and `/stop` share the process-local `liveSessions` registry. Porting `/messages` alone leaves `/stop` in Go unable to see the session Rust started — the stop button silently does nothing. There is no safe partial, which is why this was not started rather than half-landed.

[...]

Issue #275, opened 2026-08-14
Port: the scheduler (cron execution + job history writes)


vibexp-agent:
## Scope

`internal/scheduler/` — cron/interval scheduling and job execution with history.

## Traps

- Two Agento processes sharing a data dir share one scheduler: a task fires twice and the Telegram webhook is re-registered under whichever instance registered last. During the transition, only one process may schedule.

## BLOCKED BY: storage + writes.
---
**Branch:** target `desktop`, never `main`. One PR, merged before the next task starts.
**Bar:** byte-identical JSON. Any field-name, key-order, escaping or float-spelling drift is a regression.
**Verify:**
```bash
cd desktop && eval "$(./scripts/parity-instance.sh start)"
(cd src-tauri && cargo test --test live_parity -- --ignored --nocapture)
./scripts/parity-instance.sh stop
```
**Read first:** VibeXP artifact `agento-desktop-rust-port-handover-2026-08-14`, then `desktop/CLAUDE.md`.
Never diff against the server on :8990 (stale binary). Seed the scratch instance if the endpoint has no local data, or an empty response diffs clean and proves nothing.

**Blocked by:** nothing. #274 is closed, and the tool-capability refusal in `native/chat/runner.rs::build_options` that really gated the ownership flip was lifted by #310 (local tools) and #311–#317 (integration MCP servers), all merged. A scheduled task's agent now builds options natively unless it names an `mcps.yaml` server or an unhostable type (whatsapp, dropped) — decide what a scheduled run does in that case, since forwarding is not available to a fire that no Go scheduler is listening for.

_Unblocked issues in this epic can be worked in parallel._

vibexp-agent:
**Ordering finding from #274's merge — this issue is blocked by #276, not just by #274.**

Two structural facts, both from reading `internal/scheduler/`:

**1. The executor needs the agent runner, which is #276's.** `executeTask` (`internal/scheduler/executor.go:20`) loads the task and hands it to `internal/agent` — the runner that #276 owns. A ported scheduler that fires on time and cannot run anything is not a port of this issue; it is half of one. So #276 has to land first, and the "blocked by #274" in the issue body understates it.

[...]

PR #293, opened 2026-08-15
feat(desktop): give Rust the write path — storage, migrations and the pure-row writes (#274)

### LINKED ISSUE #274: Port: storage layer, migrations and all write endpoints
## Scope

Roughly **50 write routes** (POST/PUT/PATCH/DELETE) plus the 27 SQLite migrations.

**This is the point of no return.** Once Rust runs migrations, two servers can no longer share one database file, so this lands close to the cut-over — and it is **one agent's task**, not several. Parallelising writes means several people redefining migration ordering at once.

`native/db.rs` stops being read-only here.

## Traps

- **Every migration must also bump the hardcoded version in `sqlite_test.go`.**
- `PUT /integrations/{id}` destroys credentials on a scrubbed round-trip — an upstream bug the desktop UI already works around by requiring re-entry. **Do not "simplify" that away.**
- `AgentRequest` has no `permission_mode`, so agents can never persist it; the service also rejects `plan`/`dontAsk`. Match Go, then fix upstream.
- Validation errors are **422**, conflicts **409** — not 400.
- Pricing add-versus-correct are deliberately separate endpoints, not one upsert.

## BLOCKED BY: the read ports (analytics, insights, chats, tasks, settings, notifications, fs, session detail).
---
**Verify:**
```bash
```

**Blocked by:** #262, #263, #264, #265, #266, #267, #268, #271

vibexp-agent:
Closes #274

Phase 3's foundation, plus the write routes Rust can fully own today. +2684/−98.

## Storage

`db.rs` gains `open_read_write`. Its pragmas are set **per connection**, because that is what they are: WAL persists in the file, `busy_timeout`/`foreign_keys`/`synchronous` do not. Go only sets them once because `SetMaxOpenConns(1)` means it has exactly one connection. Missing `foreign_keys=ON` is the quiet failure — cascades stop firing and a deleted chat keeps its messages, with nothing to say so. Pinned by a test.

Neither handle passes `SQLITE_OPEN_CREATE`: a missing database means the app has never run, and conjuring an empty one would let a write appear to succeed against a schema that does not exist.

## Migrations — and the one deliberate departure from the issue

All 27 are ported, **but Rust verifies the schema version rather than applying it.**

Go's `applyMigrations` reads the current version *outside* the transaction that applies the next one (`internal/storage/sqlite.go`). Two processes starting together therefore both decide to apply the same version, and the loser's DDL fails with `table already exists` — which is not a conflict it retries but an error that fails `NewSQLiteDB` and takes its whole startup with it. So while the sidecar is bundled, exactly one process may migrate, and it is Go: the one that also creates the database, seeds the pricing catalog and runs the legacy filesystem import.

`migrate::apply` is written and tested and unused. #278 turns it on in the same change that stops the sidecar migrating — one commit, no window where both do it. It also re-reads the version *inside* the transaction, which is what makes two writers safe where Go's is not.

They are also **not transcribed**: `desktop/parity/migrations_vectors.json` is generated from Go, asserted against the `migrations` slice by a new test in `internal/storage`, and embedded by `migrate.rs` with `include_str!`. Adding migration 28 without regenerating fails Go's own suite. The file also outlives the Go server — same reason `desktop/parity/` keeps its other vectors.

## The seam

`Request` carries a body; `Answer` carries a status, and a `None` body for the bare `w.WriteHeader(204)` Go's deletes use — an empty JSON document under a JSON content type is a *different* response.

Two safety rules, both tested:

- **`Mode::Diff` never executes a write.** Shadow mode runs Go *and* Rust and compares them. For a read that is the point; for a mutation it applies it twice. `native::may_serve` makes this blanket over the method rather than a per-endpoint flag someone forgets on the one that mutates.
- **A write must fail before it mutates.** `Err` still means "forward to Go", so a handler that half-applied something and then errored would have it applied twice. Every handler validates, checks the schema version, and does the whole mutation in one transaction.

## Which routes moved, and why not the other 41

**A route moves only when Rust can reproduce every effect it has.** Most of the 50 write routes do something besides writing a row, and that something belongs to a sibling issue in this batch:

| Effect | Routes | Owner |
|---|---|---|
| Scheduler registration | tasks create/update/delete/pause/resume | #275 |
| Subprocess + live-session channels | chats messages/input/permission/stop | #276 |
| MCP reload, OAuth, Telegram calls | integration + trigger + webhook writes | #277 |
| Scan invalidation | `/claude-sessions/refresh` | #289 |

Porting `POST /api/tasks` without the scheduler would store a task that never fires — worse than not porting it. Also left with Go: `PUT /api/settings` (mutates process-global snapshots), the Claude-settings profiles (filesystem plus Go-side caching), and the pricing writes (in-memory cache invalidation).

So this PR moves the writes that are only rows:

- **Agents** — `POST`, `PUT`, `DELETE`
- **Chats** — `POST`, `PATCH /{id}`, `DELETE /{id}`, bulk `DELETE`
- **Job history** — `DELETE`, `DELETE /{id}`

Each module's `claims` test records what stays and why.

## Go behaviours reproduced rather than improved

- **Deleting a missing agent or chat is a 500, not a 404.** The store returns a plain `fmt.Errorf`, so `httpErr` never reaches its `NotFoundError` arm. This port does not reproduce 500s — it detects the case, writes nothing, and forwards so Go answers for itself. Job-history's delete *is* a real 404, because its service checks first; the difference is now covered by tests.
- **`permission_mode` still cannot be set through the agent API**, because `AgentRequest` has no such field. Recorded as an upstream bug; "fixing" it here would make the two implementations disagree.
- Validation order differs between agent create and update — update looks the agent up *first*, so a `PUT` to a missing agent with an empty name is a 404, not a 422.

## A divergence found while testing

**serde deserializes a struct from a JSON array**, positionally. So `POST /api/agents` with a body of `["My Agent"]` would have *created an agent* on a request Go answers with a 400. `writes::decode_body` now requires an object — and treats a `null` body the way Go does, as a documented no-op leaving the zero value, which reaches the handler and fails validation with a 422 rather than a 400.

## Verification

```
npm run build
./scripts/build-sidecar.sh
cargo fmt --check
cargo clippy --all-targets -- -D warnings
cargo test                 # 407 + 17 + 7 passed, 0 failed
cargo build --release
go test ./internal/storage/
golangci-lint run (v2.5.0, CI pin)   # 0 issues
```

Every ported write has unit tests over a database built by the **real migrations** rather than a hand-written fixture table — which is where a column default drifts from production's. `tests/parity_writes.rs` pins Go's real answers (status *and* bytes) for the same cases; it is `#[ignore]`d like every parity suite, and unlike the others it **mutates**, so it refuses to run without `AGENTO_LIVE_URL`.

Review found and this PR fixes three defects, two of them wire divergences: a missing chat answered 400 where Go answers 404, the PATCH response carried the row's old `updated_at` where Go carries the write's, and three handlers could return `Err` after `commit()` — which forwards to Go and re-applies the write. Deferred to #294, #295, #296.

Note: `tests/claude_sdk.rs::run_returns_the_final_result_and_surfaces_an_error_one` flaked once under full-suite parallel load and passes consistently in isolation and on re-runs. It is a scripted-fake-CLI timing test in a file this PR does not touch.


vibexp-agent:
Deferred to #294, #295 and #296 — the three `ticket` findings from review:

- **#294** — native route matching uses the undecoded path, so a percent-encoded id diverges from Go. Harmless while every claimed route was a read (a miss forwarded); the writes make it an invented 404.
- **#295** — a `null` inside a JSON array is a 400 here and `""` to Go. `null_is_zero_value` covers scalars only.
- **#296** — `POST /api/fs/mkdir` fits no deferral category and should be ported; the deferred-route table should be exhaustive rather than by-category.

Issue #282, opened 2026-08-14
Decide the Rust MCP server implementation, and port the SDK's typed-tool layer onto it


vibexp-agent:
Follow-up from PR #281 (relates to #269).

## What

Two decisions the SDK port deliberately left open, which phase 4 (integrations) has to settle together.

**1. Which MCP server implementation.** `start_in_process_mcp_server` (`desktop/src-tauri/src/claude/mcp.rs`) currently takes an `McpService` trait — "handle one JSON-RPC message" — and owns only the transport: bind loopback, POST → handle → respond, 202 for notifications. The Go SDK instead takes an `*mcp.Server` from `modelcontextprotocol/go-sdk` and hands it to that SDK's streamable-HTTP handler.

The Rust equivalent is `rmcp` (the official Rust SDK, 3.1.2, features `server` + `transport-streamable-http-server`). It was not taken in #281 because **nothing has an MCP server to host yet** — Agento's seven integrations plus the local-tools server are all phase 4 — so it would have added a large dependency to satisfy no caller and pinned an API that phase 4 should choose. The seam sits where Go's does, so an `rmcp` adapter is one `impl McpService` away.

**2. The typed-tool layer.** `claude/tool.go` — `ToolDef`, `NewTool[In, Out]`, `ToolServer`, `WithTools` — is not ported. It is the ergonomic layer for defining a tool as a plain function, and it depends on reflection-derived JSON schema (`google/jsonschema-go` in Go, `schemars` in Rust), which is exactly the part `rmcp` would supply.

Agento does not use it today: every integration builds an `mcp.Server` directly and passes it to `StartInProcessMCPServer`, so #281's scope — the surface Agento uses — did not include it. It is still a genuine gap against "reimplement the SDK".

## Why

Deferring is right, but only until phase 4 starts. Porting the integrations is what forces the choice, and making it implicitly (by whichever integration lands first) is how two ways to host a tool end up in the tree.

## Acceptance

- A decision recorded in `desktop/CLAUDE.md`: `rmcp` or the current trait, with the reasoning.
- If `rmcp`: `start_in_process_mcp_server` accepts an `rmcp` service, the `McpService` trait either wraps it or goes away, and the existing loopback/notification/shutdown tests still hold.
- `NewTool` / `ToolServer` / `WithTools` ported onto whatever wins, or an explicit note saying they are not needed and why.

vibexp-agent:
Decided and shipped — merged in #325 (`cdf4ca5`).

[...]

Issue #272, opened 2026-08-14
Port: integration reads


vibexp-agent:
## Scope

`GET /api/integrations` and siblings, for the **six** remaining integrations: google, github, slack, jira, confluence, telegram.

## Traps

- **Secrets are stored in plaintext** in `integrations.credentials` / `.auth`. `GET` scrubs them — the Rust version must scrub identically and must never echo them back.
- Each integration is an in-process MCP server, which is why this waits on the SDK.

## BLOCKED BY: the Claude Agent SDK port.
---
**Branch:** target `desktop`, never `main`. One PR, merged before the next task starts.
**Bar:** byte-identical JSON. Any field-name, key-order, escaping or float-spelling drift is a regression.
**Verify:**
```bash
cd desktop && eval "$(./scripts/parity-instance.sh start)"
(cd src-tauri && cargo test --test live_parity -- --ignored --nocapture)
./scripts/parity-instance.sh stop
```
**Read first:** VibeXP artifact `agento-desktop-rust-port-handover-2026-08-14`, then `desktop/CLAUDE.md`.
Never diff against the server on :8990 (stale binary). Seed the scratch instance if the endpoint has no local data, or an empty response diffs clean and proves nothing.

**Blocked by:** #269

_Unblocked issues in this epic can be worked in parallel._

vibexp-agent:
Completed in #291, merged to `desktop`. (GitHub does not auto-close on a merge to a non-default branch, so this stayed open.)

Issue #246, opened 2026-08-12
security: close the browser as an attack path — Content-Type enforcement, Host validation, loopback-by-default bind


vibexp-agent:
Agento deliberately ships without authentication — it is a single-user desktop app. That is a defensible design, but it only holds if the browser cannot be used as a way in. Today it can.

These findings are pre-existing and unrelated to any one feature; they were surfaced while reviewing #243.

## The problem

**The server binds all interfaces.** `internal/server/server.go:95` uses `Addr: fmt.Sprintf(":%d", port)`, which is `0.0.0.0` — anyone on the same network reaches an unauthenticated API.

**Loopback alone would not fix it.** A page you visit runs *in your browser*, which reaches `127.0.0.1` normally. Production CORS is same-origin (`corsMiddleware`, `server.go:141`), which correctly blocks a cross-origin `PUT` — that triggers a preflight and fails. But a `POST` with `Content-Type: text/plain` is a [simple request](https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS#simple_requests): **no preflight, so it is sent and executed.** Only the response is unreadable, which an attacker does not need. Handlers call `json.NewDecoder(r.Body).Decode()` without checking the content type, so the body is parsed as JSON regardless.

Reachable that way today:

| Endpoint | Effect |
| --- | --- |
| `POST /api/agents` | create an agent with `permission_mode: bypass` |
| `POST /api/chats` + `POST /api/chats/{id}/messages` | run it — arbitrary `Bash` |
| `POST /api/fs/mkdir` | `os.MkdirAll(clean, 0750)` on any path (`internal/api/filesystem.go:94`) |

So visiting a web page is enough to get code execution on the machine.

**There is no `Host` header validation**, so DNS rebinding defeats CORS entirely: an attacker domain that rebinds to `127.0.0.1` is *same-origin* as far as the browser is concerned. This is the standard attack against local web servers and is unaffected by binding to loopback.

## What to do, in order

1. **Require `Content-Type: application/json` on state-changing methods** (POST/PUT/PATCH/DELETE). This is the fix that closes the drive-by chain, because that content type forces a preflight that same-origin CORS then blocks. **Costs nothing:** `frontend/src/lib/api.ts:53` already sets it on every request. The multipart upload endpoint needs an explicit exemption.
[...]

PR #300, opened 2026-08-15
feat(desktop): port the integration writes that own no live state (#277)

### LINKED ISSUE #277: Port: integration writes and OAuth flows

vibexp-agent:
Closes #277

Revalidating #277's HOW against the code found that most of its scope is blocked by **#282**, which the issue does not list as a blocker — its stated blockers (#269, #274) are both merged, so it looked ready. Rather than half-port it, this moves the subset that is genuinely unblocked and states the rest. Scope finding written up on the issue: shaharia-lab/agento#277 (comment).

## What moved

| Route | Why it is safe |
|---|---|
| `POST /api/integrations` | `integrationService.Create` is validation → uuid → timestamps → `store.Save`. **No registry call** — verified against the whole function, which matters because `Create` and `Update` look alike and differ exactly here. |
| `POST /{id}/triggers` | Pure row write. |
| `PUT /{id}/triggers/{rid}` | Pure row write. |
| `DELETE /{id}/triggers/{rid}` | Pure row write. The dispatcher calls `ListRules` **per inbound message** (`dispatcher.go:168`), so there is no cached rule set for a native write to leave stale — this is what makes rule writes portable while integration writes are not. |

## What deliberately did not

- `PUT /{id}` → `registry.Reload(id)`; `DELETE /{id}` → `registry.Stop(id)`. These restart the **live in-process MCP server**. Rust has none until #282, so a native write would persist the row and leave the integration running on stale config — still using a token the user just revoked, with a 200 saying it worked.
- OAuth (`auth/start|status|validate`): in-memory flow state, must open the user's **real browser**, and writes credentials Go's MCP servers consume. All three move together or not at all — the #276 live-registry lesson.
- Webhooks: Telegram network calls, and the secret authenticates Go's **root-mounted** inbound handler.
- `POST /webhooks/telegram/{id}` + `internal/trigger/`: the dispatcher runs an agent using the integration's credentials, which this port deliberately never reads (#272).

## The credential validators are where the parity risk is

Seven types, each with its own required fields and its own **exact** 422 wording, all of which ship to the client. Two of Go's failure paths embed a string with no Rust spelling — `encoding/json`'s unmarshal error (it names Go types) and `url.Parse`'s — so both **forward** rather than guess. Everything else is reproduced:

- **Absent and `null` credentials fail differently.** Go tests `len(cfg.Credentials) == 0`, true only when the key is *absent*; a literal `null` is four bytes, decodes to the zero value, and reports the missing *field*. `gojson::captured_raw` keeps them distinguishable, and decoding through `Option<T>` stops serde erroring where Go does not — wire-format rule 4, which bit once here and was caught by its own test.
- **Jira rewrites what it stores**: trims trailing slashes and re-marshals the struct, so unknown keys the caller sent are dropped and the three known ones come back in declaration order. **Confluence, sharing the credential type, does not** — and is HTTPS-only where Jira takes either.

## A create is the first native path that handles a secret

Unavoidably: the caller supplies it. It travels as a borrowed `RawValue` from decode to `INSERT` — never through `serde_json::Value`, which would reorder keys and respell numbers — and the response is still built from `ScrubbedIntegration`, which has no field to leak it through. A test asserts the stored bytes are verbatim and the response contains no `bot_token`.

`WriteError` gains `Forbidden` for the 403 the rule routes answer when a rule belongs to another integration — checked **before** the body is decoded, so a malformed payload there is 403 and not 400. Both are asserted.

## Verification

- 459 lib + 9 `chat_turn` + 17 `claude_sdk` + 7 `scanner_roundtrip` = **492 pass**; `clippy --all-targets -D warnings` and `fmt --check` clean; release build succeeds.
- 8 new write tests run against a **really migrated** database, not a hand-rolled schema.
- The two pre-existing tests asserting these routes were *not* native were updated — that is the check that would catch an accidental claim, so it is worth noting they had to change.
- `tests/parity_writes.rs` gains a case pinning Go's own answers for all of the above, including the absent-vs-`null` pair. As ever it is `#[ignore]`d and needs a scratch instance, so **CI does not run it**.

### DIFF
--- a/desktop/CLAUDE.md
+++ b/desktop/CLAUDE.md
@@ -146,7 +146,11 @@ src-tauri/src/
     version.rs   GET /api/version and /version/update-check (dev builds only)
     notifications.rs GET /api/notifications/settings (password masked) and /log
     integrations.rs GET /api/integrations, /{id}, /available-tools, /{id}/triggers —
-                 credentials are never selected and auth is a bool made in SQL
+                 credentials are never selected and auth is a bool made in SQL;
+                 plus POST /api/integrations and the trigger-rule writes (#277).
+                 PUT/DELETE /{id} stay with Go: they reload/stop the live MCP server
+    integration_credentials.rs the seven per-type validators, and the two failures
+                 whose Go error text is not reproducible (both forward)
     fs.rs        GET /api/fs — the working-dir picker's listing (Unix; forwards on Windows)
     gopath.rs    Go's filepath.Clean/Dir/Join, pinned to vectors generated from Go
     query.rs     one query parameter, read the way r.URL.Query().Get reads it
--- a/desktop/src-tauri/src/native/integration_credentials.rs
+++ b/desktop/src-tauri/src/native/integration_credentials.rs
@@ -111,8 +111,17 @@ where
     serde_json::from_str::<Option<T>>(raw.get())
         .map(Option::unwrap_or_default)
         .map_err(|e| {
+            // **The error is deliberately not included.** serde_json quotes the
+            // offending value, so a caller who sends
+            // `"credentials": "ghp_realtoken"` would have that token echoed into
+            // the message — which the seam logs on every forward. Go's own error
+            // names the type and never the value. The position is enough to
+            // debug with and carries nothing secret.
             WriteError::Fallback(format!(
-                "{kind} credentials do not decode ({e}); Go's json error text is not reproducible here"
+                "{kind} credentials do not decode at line {} column {}; \
+                 Go's json error text is not reproducible here, so this forwards",
+                e.line(),
+                e.column()
             ))
         })
 }
@@ -370,6 +379,12 @@ fn split_url(raw: &str) -> Result<(String, String), WriteError> {
         }
     }
 
+    // A leading `:` is `getScheme`'s own error (`missing protocol scheme`),
+    // not an empty scheme, so it must forward rather than answer.
+    if raw.starts_with(':') {
+        return Err(forward());
+    }
+
     // `url.Parse` takes the scheme only when what precedes the first ':' is a
     // valid scheme; otherwise the whole string is an opaque path and Scheme is
     // "". A bare `example.atlassian.net` therefore has no scheme and no host,
@@ -392,12 +407,26 @@ fn split_url(raw: &str) -> Result<(String, String), WriteError> {
         return Ok((scheme, String::new()));
     };
     let authority = authority.split(['/', '?', '#']).next().unwrap_or_default();
-    // Userinfo is not part of Host.
-    let host = match authority.rsplit_once('@') {
-        Some((_, host)) => host,
-        None => authority,
-    };
-    Ok((scheme, host.to_string()))
+
+    // Everything below is a shape `parseHost` has a rule for that this port
+    // does not reimplement — userinfo validation, IPv6 bracket matching,
+    // percent-escapes (which a host may not contain at all), and port syntax.
+    // **Forwarding is the only safe answer for these**, because the failure
+    // mode of guessing is not a wrong message but a wrong *acceptance*: Rust
+    // would create an integration whose `site_url` Go's own parser can never
+    // read, and being native the request never reaches Go to be refused.
+    if authority.contains(['@', '[', ']', '%']) {
+        return Err(forward());
+    }
+    // A `:` in the authority introduces a port, which Go requires to be digits
+    // — and to appear at most once.
+    if let Some((host, port)) = authority.split_once(':') {
+        if !port.chars().all(|c| c.is_ascii_digit()) {
+            return Err(forward());
+        }
+        return Ok((scheme, host.to_string()));
+    }
+    Ok((scheme, authority.to_string()))
 }
 
 /// A Go-compatible JSON string literal.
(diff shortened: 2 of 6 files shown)


vibexp-agent:
## Round 2 — all six applied (`388833d`)

All six were valid, and #1 is the one that matters: my PR body and module doc both claimed credentials travel "never through `serde_json::Value`", and that was **false** — `writes::decode_body` shape-checks through a `Value` and then deserialized *from it*, so `captured_raw` captured a re-serialization. Corrected the claim as well as the code.

| # | Fix |
|---|---|
| 1 🔴 | `decode_body` now deserializes from the **original bytes**; the `Value` stays a shape check only. Pinned at that level (`writes.rs`) rather than only at the call site that noticed. |
| 2 🔴 | `split_url` forwards on any authority containing `@`, `[`, `]` or `%`, and on a non-numeric or repeated port. A numeric port still resolves natively. |
| 3 🟠 | `json_string` gains `\b`/`\f` arms ahead of the `< 0x20` catch-all. |
| 4 🟠 | Both fixtures widened to a multi-key blob with a decimal and interior whitespace. |
| 5 🟡 | The forward reason carries **line and column only** — serde's message quotes the offending value, so a caller sending `"credentials": "ghp_realtoken"` had it written to the log on every forward. |
| 6 🟡 | A leading `:` forwards (`missing protocol scheme`) instead of answering with `got ""`. |

**Reverted #1 to confirm #4 now bites:** the widened assertion fails with `{"bot_token":"t","rate":1.5,"zebra":"z"}` against the expected `{"zebra":"z", "bot_token":"t","rate":1.50}` — sorted, respelled and whitespace-stripped, exactly as you measured.

New tests: byte preservation in `decode_body`; all seven over-accepted URL shapes forwarding for both atlassian types; a numeric port still resolving; the `\b`/`\f` table; and that a bad blob's error never contains the credential.

**Gate:** 464 lib + 9 `chat_turn` + 17 `claude_sdk` + 7 `scanner_roundtrip` = 497 pass; clippy `-D warnings` and fmt clean; release build succeeds.

On the observations: the missing OTel spans/logs for natively-served writes and the `guards.go` 415-vs-doc drift are both real and both pre-date this diff — I'll file them rather than widen this PR. `Closes #277` never registers on this repo because the PR targets `desktop` rather than the default branch, so these issues are closed by hand at merge.

vibexp-agent:
Filed both observations rather than widening this PR:

- **#301** — natively-served requests emit no OTel spans and no application log lines. Systemic since #274, not introduced here. Framed as a decision to make (reproduce the spans in Rust, or document that the desktop build is not observable) rather than a bug, since telemetry is listed under *Do not port* — but the current half-and-half state is the most misleading of the options.
- **#302** — this one turned out to be sharper than doc drift. The root `CLAUDE.md` says "a body-less request of any method passes"; `guards.go:114-120` says the exact opposite **and explains why** — a cross-origin `POST` with no body and no `Content-Type` is itself a CORS-simple request, so exempting it reopens the hole #246 closed. It also describes multipart as generally admitted when it is gated to `/api/uploads`. Since `CLAUDE.md` is loaded as authoritative instruction, an agent trusting it would read a 415 on a body-less `DELETE` as a bug and "fix" it in exactly the wrong direction. Filed with the replacement wording.

Good catch on both — neither was in the diff.

vibexp-agent:
## Round 4 — `8b8e18a`

**#1 and #2 confirmed done by re-review** (ASCII enumeration: 0 over-accepts, 0 host mismatches; `u.Host` identical across every accepted case). **#3 was rejected, correctly.**

My duplicate-key fallback was wrong in exactly the direction this review has been circling — it over-accepted. `encoding/json` keeps the last occurrence but **type-checks every one**, and the collapsed `serde_json::Value` has already discarded the occurrences Go would have judged. So `{"count":"notanumber","count":1}` is a 400 to Go and decoded cleanly here; on `POST /api/integrations` that is a row created which Go would have refused. It also lost byte-verbatim capture on that path, re-breaking finding #1 for those bodies.

That makes forwarding the *only* correct option rather than merely the cautious one: neither local path can reproduce the answer. `decode_body` now forwards duplicate-key bodies, and Go accepts the well-typed ones and 400s the ill-typed ones. A non-duplicate malformed body is still a local `InvalidBody` — this does not turn every decode failure into a sidecar round trip.

Added a test asserting ordinary bodies still take the **strict** path, so a future change that routes normal traffic through the fallback fails loudly rather than silently re-breaking the storage guarantee.

**Gate:** 470 lib + 9 `chat_turn` + 17 `claude_sdk` + 7 `scanner_roundtrip` = 503 pass; clippy `-D warnings` and fmt clean; release build succeeds. CI green on `25bb485` (all four) and re-running.

Noting for the record that three of my four fix rounds on this PR introduced a new divergence that review then caught — the `Value` round trip, then the shape-based URL guard, then this. Each was a case of reproducing what Go *looked like* it did instead of what it does.

PR #336, opened 2026-08-16
fix(desktop): a null inside a JSON array is a zero value, not a type error (#295)

### LINKED ISSUE #295: desktop: a null inside a JSON array diverges from Go's zero-value semantics
Follow-up from PR #293 (relates to #274).

`gojson::null_is_zero_value` is applied to **scalars** only. Inside an array it is not:

- `{"ids":[null]}` is a serde type error → `InvalidBody` → **400**. Go unmarshals it to `[""]` and proceeds.
- `{"capabilities":{"built_in":[null]}}` is the same shape.

Reachable only by a hand-crafted request, which is why it was recorded rather than fixed in #293 — but that PR makes the null-fidelity argument in three places (`writes.rs`, `desktop/CLAUDE.md`) and this is the one shape it does not cover.

**Fix:** a deserializer for `Vec<String>` that maps an element `null` to `""`, applied to `BulkDeleteRequest.ids` and `Capabilities`' three lists.

**Files:** `desktop/src-tauri/src/native/chats.rs`, `desktop/src-tauri/src/native/tasks.rs`, `desktop/src-tauri/src/native/agents.rs`, `desktop/src-tauri/src/native/gojson.rs`.

vibexp-agent:
Closes #295

## What

Two types in `gojson.rs` — `GoList<T>` and `GoMap<V>` — decode a `null` **inside** a JSON container as the zero value, the way `encoding/json` does. Applied to `BulkDeleteRequest.ids` (`chats.rs`, `tasks.rs`), all four containers in `Capabilities`/`McpCapability`, and the claimed integration writes (`ServiceConfig.tools`, `CreateIntegrationRequest.services`, the trigger rule's two filter lists).

## Why

`gojson::null_is_zero_value` is attached to fields, and `serde` only consults it where it is attached — so it covered `{"ids":null}` and stopped there. Measured against `encoding/json`:

| body | Go | Rust before |
|---|---|---|
| `{"ids":[null]}` | `[""]`, no error | type error → `InvalidBody` → **400** |
| `{"ids":["a",null,"b"]}` | `["a","","b"]`, no error | 400 |
| `{"capabilities":{"built_in":[null]}}` | `[""]`, no error | 400 |
| `{"capabilities":{"mcp":{"s":null}}}` | `{"s":{"tools":null}}`, no error | 400 |
| `{"services":{"slack":null}}` | zero `ServiceConfig`, no error | 400 |
| `{"ids":[1]}` | error | error ✔ |

Reachable only by a hand-crafted request, which is why #293 recorded it rather than fixing it — but that PR makes the null-fidelity argument in three places, and these are the shapes it does not cover. While every claimed route was a read this degraded to a fallback; on the writes #274 and #277 claimed it is a rejection of a request Go applies.

## Why types, and not `deserialize_with` functions

This is the substantive thing the PR learned, and it was caught in review after the first version shipped it.

`serde`'s derive makes a field carrying `deserialize_with` **required** — the `missing_field` path that lets a bare `Option` default to `None` is not generated — so every call site had to add `#[serde(default)]`. That attribute also feeds the derive's `visit_seq` arm, which rejects a short array **only for fields with no default**. Measured on the first version:

| body | before the PR | first version | Go |
|---|---|---|---|
| `{"capabilities":[]}` | 400 | **201, agent created** | 400 |
| `{"capabilities":{"mcp":{"s":[]}}}` | 400 | **201, agent created** | 400 |

A fix for a `null` would have shipped a widened **over-accept** — the one direction this port must not move in, because `Err` means forward and nothing errors when Rust accepts what Go refuses. A type needs no attribute at all, so `Capabilities` and `McpCapability` now carry none and are exactly as strict about their own shape as before. Pinned by `a_container_default_would_have_widened_the_struct_from_array_over_accept`.

## Scope note

The issue lists the array case. The **map-value** case (`{"mcp":{"s":null}}`) is the identical divergence one line away in the same struct, and the integration writes carry the same containers — fixing three of `Capabilities`' four would leave it unfaithful, and the same argument applies verbatim to `integrations.rs`. Both are included.

## How

- `GoList<T>` / `GoMap<V>` with hand-written `Deserialize`, plus `Deref` and `From` so call sites read unchanged.
- Both **keep the outer `Option`**, because that is the nil-versus-empty distinction Go puts on the wire: `null` stays `None`, `[]` stays `Some(vec![])`.
- A newtype **serializes as its inner value**, so every read is byte-identical. Pinned by `the_types_serialize_as_their_inner_value`.
- `ServiceConfig` gains a `Default` derive (`GoMap` requires it); its zero is Go's zero.

## Acceptance

- `gojson`'s tests pin each type against answers measured from `encoding/json`, including the nil-versus-empty rows, the wrongly-typed rows that must still fail, and the two over-accept boundaries.
- Five call-site tests — `chats`, `tasks`, `agents` (through `create()` and the stored row) and two in `integrations` — assert the end-to-end effect, and **all seven behaviour tests go red when `GoList`/`GoMap` are reduced to pass-throughs** (verified).
- The read-side guard pins that a stored `[]` must not read as an *empty* allowlist.
- Local gates green: `cargo fmt --check`, `cargo clippy --all-targets -- -D warnings`, `cargo test` (664 lib), `cargo build --release`, `npm run build`, `go test ./desktop/...`, `go vet ./desktop/...`, golangci-lint at the CI-pinned v2.5.0.

## Deferred, recorded rather than fixed

`serde` builds a nested struct from a **full-length** JSON array, positionally, so `{"capabilities":[["Read"],null,null]}` is accepted here and Go answers 400. That is `writes::decode_body`'s #274 quirk one level deeper — `decode_body` guards the body, nothing checks a value *inside* it. It predates this change and is an **over-accept**, so it is its own issue: **#337**. Pinned by `a_full_length_positional_array_is_a_known_over_accept` and noted in `desktop/CLAUDE.md`.

## Review

Automated review returned **REQUEST CHANGES** on the first version, for the over-accept above; all findings were applied and the rework re-verified.

### DIFF
--- a/desktop/CLAUDE.md
+++ b/desktop/CLAUDE.md
@@ -451,6 +451,26 @@ conversation**, because `parentUuid` is `null` on exactly the event that starts
 one. The live diff caught it; no unit test would have, since the fixtures were
 all written by hand with the field present.
 
+**`serde` only consults the rule where it is attached, so a container needs its
+own** (#295). `null_is_zero_value` covers the *field*: `{"ids":null}` was
+already `None` while `{"ids":[null]}` stayed a type error, and Go answers `[""]`
+to the second with no error at all. `null_elements_are_zero_values` is that one
+level down, `null_map_values_are_zero_values` is the same for a `null` object
+*value* (`{"mcp":{"s":null}}` is the zero struct to Go), and both keep the outer
+`Option` so the nil-versus-empty distinction is untouched. They are on
+`BulkDeleteRequest.ids` in `chats.rs` and `tasks.rs` and on all three of
+`Capabilities`' lists plus its MCP map. On a *read* this class of bug degrades
+to a fallback; on the writes #274 claimed it is a **400 or 422 for a request Go
+applies**.
+
+One thing those helpers deliberately do **not** fix, pinned in
+`gojson.rs::a_nested_struct_from_an_array_is_a_known_over_accept`: serde builds
+a struct from a JSON *array*, positionally, so `{"mcp":{"s":[]}}` is a zero
+`McpCapability` here and a 400 to Go. `writes::decode_body` guards that shape at
+the body level (#274) and nothing checks it for a value *inside* the body. It is
+an over-accept — the direction that creates rows Go refuses — and it predates
+#295.
+
 Genuinely unparseable input still fails, which is also what Go does — so the
 null case and the malformed case need separate tests.
 
--- a/desktop/CLAUDE.md
+++ b/desktop/CLAUDE.md
@@ -454,22 +454,35 @@ all written by hand with the field present.
 **`serde` only consults the rule where it is attached, so a container needs its
 own** (#295). `null_is_zero_value` covers the *field*: `{"ids":null}` was
 already `None` while `{"ids":[null]}` stayed a type error, and Go answers `[""]`
-to the second with no error at all. `null_elements_are_zero_values` is that one
-level down, `null_map_values_are_zero_values` is the same for a `null` object
-*value* (`{"mcp":{"s":null}}` is the zero struct to Go), and both keep the outer
-`Option` so the nil-versus-empty distinction is untouched. They are on
-`BulkDeleteRequest.ids` in `chats.rs` and `tasks.rs` and on all three of
-`Capabilities`' lists plus its MCP map. On a *read* this class of bug degrades
-to a fallback; on the writes #274 claimed it is a **400 or 422 for a request Go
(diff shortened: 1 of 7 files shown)


vibexp-agent:
Deferred finding from this PR filed as #337 — a nested struct still deserializes from a JSON array positionally, so `{"mcp":{"s":[]}}` creates a row Go refuses. Pinned here by `a_nested_struct_from_an_array_is_a_known_over_accept`.

PR #339, opened 2026-08-16
refactor(desktop): resolve the fallback model lazily, and dedup error_json (#299)

### LINKED ISSUE #299: desktop: chat turn opens a needless second DB connection, and error_json is duplicated
Follow-up from PR #297 (relates to #276). Both diff-scoped cleanups, neither behavioural.

- **`turn::run` calls `default_model(&db_path)` on every turn where `chat_sessions.model` is empty**, opening a second read-only SQLite connection and loading the settings row — including for agent chats, where the value is then discarded because `agent.model` wins. Go only computes it in the no-agent branch (`chat_service.go`, `resolveAgentConfig`). Make it lazy.

- **`error_json` is duplicated verbatim** in `native/chat/turn.rs` and `native/chat/mod.rs`, nested `ErrorBody` struct included. One copy, in `chat/mod.rs`.

vibexp-agent:
Closes #299

Both items from the issue, plus a parity divergence the first one turned up.

## 1 — The fallback model is resolved lazily

`turn::run` called `default_model(&db_path)` on every turn where `chat_sessions.model` is empty, opening a second read-only SQLite connection and loading the settings row — including for agent chats, where the value was then discarded. `RunSpec.no_agent_model` is a `Box<dyn Fn() -> String + Send + Sync>` now (renamed: `Options::fallback_model` already means the CLI's `--fallback-model` in the same function), called at most once and only in the branch that can use it.

### …and it is not purely a cleanup

The issue calls both items non-behavioural. The first one is not, and doing it properly is what surfaced why.

Go's `resolveAgentConfig` (`internal/service/chat_service.go:279`) branches on whether the chat **names an agent**, not on whether that agent has a model — it returns the agent's config outright — and `internal/agent/runner.go:250` then sets a model only when `agentCfg.Model != ""`. Verified the whole chain: `resolveAgentConfig` → `agent.StartSession` → that `if`.

| chat | Go | Rust before | Rust now |
|---|---|---|---|
| agent with a model | the agent's | the agent's | the agent's |
| **agent with an empty model** | **no model option at all** | session model, else the user's default | **no model option at all** |
| no agent, session model set | the session's | the session's | the session's |
| no agent, no session model | the user's default | the user's default | the user's default |

So the port treated an agent's empty model as a request for a default Go would never have given it — a different model from the one Go picks, silently. `build_options` matches on the **agent** rather than on its model now, which is the laziness and the fix in one line.

## 2 — `error_json` is one copy

It and its nested `ErrorBody` were duplicated verbatim in `native/chat/turn.rs` and `native/chat/mod.rs`. `turn.rs` uses the one in `chat/mod.rs`; a private item in a parent module is already visible to its children, so no visibility changed.

## Acceptance

- Three tests in `runner.rs` **count the closure's calls** rather than only checking the resulting model — a test that asserted the model alone would pass with the work still being done on every turn. An agent chat resolves it **0** times, a no-agent chat exactly **1**.
- The empty-agent-model case asserts `opts.model == DEFAULT_MODEL`, i.e. `with_model` was never called. The test models are deliberately distinct from `DEFAULT_MODEL`, because `Options::default()` already carries it and the existing `agent_with` helper's model is *exactly* that string — so a test reusing it could not tell "we set the model" from "we never called `with_model`".
- Local gates green: `cargo fmt --check`, `cargo clippy --all-targets -- -D warnings`, `cargo test`, `cargo build --release`, `npm run build`, `go test ./desktop/...`, golangci-lint at the CI-pinned v2.5.0.

## 3 — The user's default is *resolved*, not read raw

Found in review, and it is the last row of the table above. `default_model` read `load_stored` — the raw `SELECT` — where Go reads `settingsMgr.Get()`. `SettingsManager.load` fills `"sonnet"` when nothing is stored, *before* `applyEnvOverrides` applies `AGENTO_DEFAULT_MODEL` / `ANTHROPIC_DEFAULT_SONNET_MODEL`. Consequences for a chat with no agent and no session model:

| | Go | Rust before |
|---|---|---|
| nothing ever saved | `sonnet` | `claude-sonnet-4-6` (the SDK default, a different string) |
| `AGENTO_DEFAULT_MODEL` exported | honoured | **ignored** |

`settings::resolve` is the documented mirror of `Get()` and every other caller in the port already goes through it. Pre-existing — the eager version called the same function — but this PR is the audit of exactly this resolution.

## What this PR does *not* remove

`build_options` still calls `resolve_agent_config_dir` unconditionally, which falls through to `stored_config_dir()` and opens its own read-only connection to load the same `user_settings` row unless the agent carries an absolute override. So the per-turn settings read is **narrowed, not eliminated** — what went away is the redundant second one. Filed as **#340**, which also notes that two consumers of one row with different fields is the shape that drifts: it is how one of them ended up on `load_stored` and the other on `resolve`.

## Review

**APPROVE with one follow-up.** The reviewer traced the Go chain independently at every link and confirmed the parity claim, then found the `load_stored` divergence above (fixed here), that the closure-selection branch had no coverage (now extracted as `no_agent_model_for` and tested on all three arms), the `fallback_model` name collision (renamed), and that "no model option at all" needed a clause saying the SDK default is what reaches the CLI (added).

Worth recording: an empty agent model is not reachable through any supported write path — create, update, the YAML import and the column default all fill `claude-sonnet-4-6` — so the parity fix is cheap insurance rather than a live bug. The `load_stored` one is reachable today by anyone who has not saved settings.

### DIFF
--- a/desktop/CLAUDE.md
+++ b/desktop/CLAUDE.md
@@ -1034,16 +1034,30 @@ issue. That gap is #298; do not read the list below as fully covered.
   the session row is still written: `updated_at`, the token totals, and on a
   first message a title derived from a message that was never stored.
 
-**The model a turn runs is the agent's or the fallback's — never both** (#299).
-`resolveAgentConfig` branches on whether the chat **names an agent**, not on
-whether that agent has a model: it returns the agent's config outright, and
+**The model a turn runs is the agent's or the no-agent branch's — never both**
+(#299). `resolveAgentConfig` branches on whether the chat **names an agent**, not
+on whether that agent has a model: it returns the agent's config outright, and
 `runner.go` then sets a model only when `agentCfg.Model != ""`. So an agent with
-an empty model runs with **no model option at all**, and the session's model and
-the user's default are read only in the no-agent branch. `RunSpec.fallback_model`
-is a closure for that reason — resolving it eagerly opened a second read-only
-SQLite connection and loaded the settings row on every turn of every agent chat,
-to throw the answer away, *and* treated an agent's empty model as a request for a
-default it would never have got from Go.
+an empty model gets **no model option from Agento at all** — the SDK's own
+default (`claude-sonnet-4-6`, the same constant in both SDKs) is what reaches the
+CLI — and the session's model and the user's default are read only in the
+no-agent branch. `RunSpec.no_agent_model` is a closure for that reason: resolving
+it eagerly loaded the settings row on every turn of every agent chat to throw the
+answer away, *and* treated an agent's empty model as a request for a default Go
+would never have given it.
+
+It is named for the branch and not for the fallback because
+`Options::fallback_model` already means something else in the same function — the
+CLI's `--fallback-model`, for when the *primary* model is unavailable.
+
+**Read the user's default through `settings::resolve`, never `load_stored`.** Go
+reads `settingsMgr.Get()`, and `SettingsManager.load` fills `"sonnet"` when
+nothing is stored *before* `applyEnvOverrides` applies `AGENTO_DEFAULT_MODEL` /
+`ANTHROPIC_DEFAULT_SONNET_MODEL`. The raw `SELECT` has neither: a user who had
+never saved settings ran on the SDK default rather than `sonnet` — two different
+strings — and one who exported `AGENTO_DEFAULT_MODEL` had it silently ignored.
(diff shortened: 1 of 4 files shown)


vibexp-agent:
Follow-up filed as #340: `build_options` still opens a settings connection per turn to resolve the Claude config dir, so this PR narrows the per-turn read rather than removing it.

PR #341, opened 2026-08-16
test(desktop): cover the permission round trip, and compact embedded RawValues (#298)

### LINKED ISSUE #298: desktop: the chat permission round trip has no test, and embedded RawValues skip Go's compact/escape
Follow-up from PR #297 (relates to #276). Two findings from that review, both `ticket`.

## 1. The permission path is untested

`tests/chat_turn.rs` drives `AskUserQuestion` as an assistant `tool_use` block, so it exercises `extract_ask_user_question` → the post-result continuation and nothing else. Untested, in code whose comments call these headline rules:

- **"`AskUserQuestion` is answered by *denying* the tool with the user's text as the message"** — `turn.rs`. `desktop/CLAUDE.md` lists it among rules "all pinned by `tests/chat_turn.rs`". It is not.
- `wrap_permission_handler` (`runner.rs`) — the `AskUserQuestion` allowlist bypass, the deny-**without**-a-prompt for a tool outside the list, and the unwrapped-when-empty case. Zero tests.
- `Notify::Permission` → the `permission_request` frame, and the drop-on-full at `NOTIFY_CAPACITY`.
- The third disconnect wait (the `can_use_tool` race added in #297). **Correction (2026-08-15):** this bullet originally said the other two waits were covered by `a_disconnect_while_a_prompt_is_pending_releases_the_chat`. They were not — that test dropped the body before any frame was written, so it tore down through the pre-existing failed-send path and passed with the entire disconnect fix reverted. It has since been rewritten to read frames until the prompt arrives before disconnecting, and a sibling added for the loop arm; both now fail when reverted. The `can_use_tool` wait remains the uncovered one.
- The multi-turn persistence claim: blocks and token totals accumulate across turns while `assistant_text` is reset. The AskUserQuestion test asserts frame names only and never reads the database.

**Do:** extend the fake CLI to issue a `can_use_tool` control request — the harness already handles them for `initialize` — and assert the `permission_request` frame, the deny-with-the-user's-text written back on stdin, and that an out-of-allowlist tool produces **no** frame. Add a DB assertion to the multi-turn test. Then correct the coverage claim in `desktop/CLAUDE.md`.

## 2. Embedded `RawValue`s bypass Go's compact + HTML-escape

`encoding/json` runs `compact(…, escapeHTML=true)` over a `Marshaler`'s output, so a `json.RawMessage` nested inside a marshalled value is whitespace-stripped **and** has `<`, `>`, `&`, U+2028/9 escaped. serde_json's `RawValue` writes its bytes as-is and `gojson::GoFormatter` never sees them.

- **SSE frames** (`user_input_required`, `permission_request`): Go ships `{"question":"a \u0026 b"}`, Rust ships `{"question":"a & b"}`.
- **The stored `blocks` column**: Rust writes uncompacted, unescaped bytes. Masked on read — `chats::decode_blocks` applies `compact_raw` — so the API response still matches, but the two implementations' database bytes differ for the same input.

`gojson::compact_raw` already exists and is already used on the read side.

**Do:** apply it to `input` in `append_assistant_blocks` (or in `append_message` before encoding) and to both synthetic-frame payloads. Also correct the comment in `persist.rs` that says Go stores verbatim and compacts on emit — Go compacts **on store**.

vibexp-agent:
Closes #298

Both halves of the issue.

## 1 — The permission path had no coverage at all

`tests/chat_turn.rs` drove `AskUserQuestion` as an assistant `tool_use` block, which reaches `extract_ask_user_question` and the post-result continuation — and **not the permission handler**. `desktop/CLAUDE.md` listed the deny-with-the-user's-text rule among those "pinned by `tests/chat_turn.rs`". It was not.

The fake CLI issues a real `can_use_tool` control request now, and **logs every stdin line**, because the entire observable effect of that round trip on the CLI side is a `control_response` — frame assertions cannot see it.

| test | asserts |
|---|---|
| `an_ask_user_question_is_answered_by_denying_the_tool_with_the_users_text` | the `user_input_required` frame, and that the answer goes back as `{"behavior":"deny","message":"<the user's text>"}` |
| `an_ordinary_tool_prompts_for_permission_and_the_answer_is_written_back` | the `permission_request` frame with its input, and that allow **and** deny both reach the CLI — deny with Go's own `Permission denied by user` |
| `a_tool_outside_the_agents_allowlist_is_denied_without_a_prompt` | **no** frame, and the deny carrying `wrap_permission_handler`'s message. The absence is the assertion |
| `ask_user_question_bypasses_the_allowlist` | the same agent, which names only `Read`, still prompts for `AskUserQuestion` |

**All four confirmed to fail under the matching mutation** — turning the deny into an allow, dropping the `AskUserQuestion ||` from the allowlist check, and returning the inner handler unwrapped. The first is the one that matters: reverting `PermissionResult::deny(answer)` to an allow leaves **every frame unchanged**, which is precisely why the old frame-name assertions could never have caught it.

The multi-turn test now reads the database: one user message plus the **last** turn's text (not both concatenated), both `tool_use` blocks, and `3+5`/`4+6` tokens — the accumulation-vs-reset claim the frame names could not reach. The fake CLI reports per-turn usage so that assertion is real rather than `(0, 0)`.

## 2 — Embedded `RawValue`s skipped Go's compact + HTML-escape

`encoding/json` runs `compact(…, escapeHTML=true)` over a `Marshaler`'s output, so a nested `json.RawMessage` is whitespace-stripped and has `<`, `>`, `&` and U+2028/9 escaped — **while keeping its key order and number spelling**. `serde_json` writes a `RawValue`'s bytes as-is through `write_raw_fragment`, which `GoFormatter` never sees. Measured against Go, marshalling `json.RawMessage(`{ "z" : 1.50 , "a" : "a < b & c" }`)`:

```
Go:   {"z":1.50,"a":"a < b & c"}
Rust: { "z" : 1.50 , "a" : "a < b & c" }
```

- **The synthetic SSE frames.** `gojson::serialize_compacted` is attached to the *field* rather than applied at each construction site, so a future one cannot forget it — the type says how it is emitted.
- **The stored `blocks` column.** Go compacts **on store**, not on emit; this file and `persist.rs` both said the opposite. Writing the SDK's bytes verbatim left the two implementations' databases different for the same input, **masked on read** because `chats::decode_blocks` compacts what it loads — which is exactly why nothing noticed.

Both directions matter and pull against each other: `compact` strips whitespace and escapes without doing what a `serde_json::Value` round trip would do to key order and `1.50`.

## Acceptance

- Every expected byte string was measured from `encoding/json`, not reasoned about.
- The `chat_turn` suite was run **25×** with no failures and no hangs — the check this file's own history calls for, since a hang rather than a failure is its characteristic failure mode.
- `desktop/CLAUDE.md`'s coverage claim is corrected: it now says what the old test actually reached and what closed the gap.
- Local gates green: `cargo fmt --check`, `cargo clippy --all-targets -- -D warnings`, `cargo test`, `cargo build --release`, `npm run build`, `go test ./desktop/...`, golangci-lint at the CI-pinned v2.5.0.

### DIFF
--- a/desktop/CLAUDE.md
+++ b/desktop/CLAUDE.md
@@ -1015,9 +1015,20 @@ for that chat** and forward otherwise. Go then answers — correctly, because it
 is the side that has the session.
 
 Five rules that are silent when broken, all pinned by `tests/chat_turn.rs` —
-**except** the deny-with-the-user's-text half of `AskUserQuestion`, which is
-reached only through a `can_use_tool` control request the fake CLI does not yet
-issue. That gap is #298; do not read the list below as fully covered.
+including, since #298, the deny-with-the-user's-text half of `AskUserQuestion`.
+That one was the gap the claim used to paper over: the suite drove
+`AskUserQuestion` as an assistant `tool_use` block, which reaches
+`extract_ask_user_question` and the post-result continuation and **not the
+permission handler at all**. The fake CLI issues a real `can_use_tool` control
+request now, and the assertion is on what the SDK wrote *back* — the whole
+observable effect of that round trip is a `control_response`, so the fake CLI
+logs every stdin line and the tests read it. Reverting the deny to an allow
+leaves every frame unchanged and fails only there.
+
+Covered with it: `wrap_permission_handler`'s allowlist (a tool the agent does not
+name is denied **without a prompt** — the absence is the assertion), the
+`AskUserQuestion` bypass of that allowlist, and the `permission_request` frame
+with its allow and deny answers.
 
 - **`result` is not terminal.** With an `AskUserQuestion` pending the same
   subprocess carries on, so one HTTP request spans several turns and several
@@ -1058,6 +1069,24 @@ never saved settings ran on the SDK default rather than `sonnet` — two differe
 strings — and one who exported `AGENTO_DEFAULT_MODEL` had it silently ignored.
 `resolve` is the documented mirror of `Get()`, and every other caller in the port
 already goes through it.
+
+**An embedded raw value is compacted and HTML-escaped on the way out, and Go
+does it on the way *in*** (#298). `encoding/json` runs
+`compact(…, escapeHTML=true)` over a `Marshaler`'s output, so a nested
+`json.RawMessage` is whitespace-stripped and has `<`, `>`, `&` and U+2028/9
+escaped — while keeping its key order and number spelling. `serde_json` writes a
+`RawValue`'s bytes as-is through `write_raw_fragment`, which `GoFormatter` never
+sees. Two places that mattered:
+
+- the **synthetic SSE frames**: Go ships `{"question":"a \u0026 b"}` where Rust
+  shipped `{"question":"a & b"}`. `gojson::serialize_compacted` is attached to the
+  field rather than applied at each construction site, so a future one cannot
+  forget it;
+- the **stored `blocks` column**: Go compacts **on store**, not on emit — this
+  file and `persist.rs` both used to say the opposite — so writing the SDK's bytes
+  verbatim left the two implementations' databases different for the same input.
+  It was masked on read, because `chats::decode_blocks` compacts what it loads,
+  which is exactly why nothing noticed.
 
 **The `tool_use` input must never round-trip through a `serde_json::Value`.**
 The first version of `append_assistant_blocks` did, and turned `{"z":1.50,"a":1}`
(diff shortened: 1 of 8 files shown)


vibexp-agent:
Review follow-ups applied in `f3a44b0`. The one finding left as a follow-up is filed as **#342**: the SDK's `can_use_tool` arm echoes `updatedInput` through a `serde_json::Value`, so the CLI runs a re-sorted, re-spelled tool input. It is outside this diff (it predates it, from the SDK port) and it is the *last* site violating the port's own "`serde_json::Value` is a parity smell" rule — and the only one whose blast radius is what a tool actually executes rather than what the UI renders.

Issue #317, opened 2026-08-16
Port: the Confluence integration MCP server


vibexp-agent:
## Scope

`internal/integrations/confluence/` — Agento's Confluence integration, an in-process MCP server (444 lines of Go).

One of six sibling issues, one per backend. **They are independent of each other** — all six are blocked only by #282, and can be worked in parallel by six agents once the MCP host is chosen.

## Acceptance

- Every tool the Go server exposes, under the **same tool name and the same input schema** — the names travel in each agent's `capabilities.mcp` allowlist and in every stored `tool_use` block, so a rename silently breaks existing agents and rewrites history.
- Tool *results* are what the model reads and what gets persisted; the parity bar is the result payload, not just the schema.
- Credentials come from the stored integration row and never appear in a response or a log line.
- See the Jira note: same credential struct, different validation. The smallest of the six.

**Blocked by:** #282

---
**Branch:** target `desktop`, **never `main`**. One PR per issue, merged before the next starts.
**Bar:** byte-identical JSON. Any field-name, key-order, escaping or float-spelling drift is a regression.
**Verify:**
```bash
cd desktop && eval "$(./scripts/parity-instance.sh start)"
(cd src-tauri && cargo test --test live_parity -- --ignored --nocapture)
./scripts/parity-instance.sh stop
```
**Read first:** `desktop/CLAUDE.md`, then the seam rules in `src-tauri/src/native/mod.rs` and `native/writes.rs`.
Never diff against the server on :8990 (it is a stale binary). Seed the scratch instance if the endpoint has no local data, or an empty response diffs clean and proves nothing.

vibexp-agent:
Ported in #356, merged to `desktop` as c00555a7c.

`native/integrations/confluence/` — six tools in one service group, over a site URL, email and API token read from the row. `confluence` joined `registry::HOSTED_TYPES`, so the sidecar now runs with `AGENTO_INTEGRATIONS=off:github,confluence` and this shell is the only process hosting it.

Five surfaces pinned by `desktop/parity/confluence_vectors.json`, taken from the running Go server over its real MCP transport: the hosted tool set, each schema, the request each tool built, the result text, and `ValidateSiteURL` per input.

[...]

Issue #313, opened 2026-08-16
Port: the Google integration MCP server


vibexp-agent:
## Scope

`internal/integrations/google/` — Agento's Google (Gmail, Calendar, Drive) integration, an in-process MCP server (763 lines of Go).

One of six sibling issues, one per backend. **They are independent of each other** — all six are blocked only by #282, and can be worked in parallel by six agents once the MCP host is chosen.

## Acceptance

- Every tool the Go server exposes, under the **same tool name and the same input schema** — the names travel in each agent's `capabilities.mcp` allowlist and in every stored `tool_use` block, so a rename silently breaks existing agents and rewrites history.
- Tool *results* are what the model reads and what gets persisted; the parity bar is the result payload, not just the schema.
- Credentials come from the stored integration row and never appear in a response or a log line.
- OAuth2 access tokens are minted from the stored refresh token; the refresh path is shared with #OAUTH and should not be duplicated here.

**Blocked by:** #282

---
**Branch:** target `desktop`, **never `main`**. One PR per issue, merged before the next starts.
**Bar:** byte-identical JSON. Any field-name, key-order, escaping or float-spelling drift is a regression.
**Verify:**
```bash
cd desktop && eval "$(./scripts/parity-instance.sh start)"
(cd src-tauri && cargo test --test live_parity -- --ignored --nocapture)
./scripts/parity-instance.sh stop
```
**Read first:** `desktop/CLAUDE.md`, then the seam rules in `src-tauri/src/native/mod.rs` and `native/writes.rs`.
Never diff against the server on :8990 (it is a stale binary). Seed the scratch instance if the endpoint has no local data, or an empty response diffs clean and proves nothing.

vibexp-agent:
Done in #360 (the port) and #361 (the hosting flip).

**#360 — the port, measured rather than read.** Eight tools across calendar, gmail and drive over a self-refreshing OAuth2 token. Unlike the other five integrations Google's requests are built by generated client libraries, so `desktop/parity/google_vectors.json` is not a confirmation of the port — it is the only written specification of what that code generator emits.

**#361 — the flip.** `google` joins `HOSTED_TYPES`, so all six integration types are now hosted by the Rust shell and the sidecar is told to drop every one of them. Only `whatsapp` remains Go's, and it is dropped rather than deferred.

Seven things were found that reading could not have produced, every one by adding a vector and letting real Go answer:

[...]

Issue #311, opened 2026-08-16
Port: the integration registry lifecycle, and PUT/DELETE /api/integrations/{id}


vibexp-agent:
## Scope

`internal/integrations/registry.go` — the Start/Stop/Reload lifecycle — plus the two routes that drive it.

| Route | Effect |
|---|---|
| `PUT /api/integrations/{id}` | row write **+ `registry.Reload(id)`** |
| `DELETE /api/integrations/{id}` | row write **+ `registry.Stop(id)`** |

## Why it was deferred

Quoting PR #300, which ported the integration writes that own no live state:

> These restart the **live in-process MCP server**. Rust has none until #282, so a native write would persist the row and leave the integration running on stale config — still using a token the user just revoked, with a 200 saying it worked.

This is the #276 live-registry lesson in a second place: process state has exactly one owner, and a write that moves the row without moving the process is worse than not porting the route.

## Acceptance

- A Rust registry with the same Start/Stop/Reload semantics, hosting whatever #282 chose.
- `PUT`/`DELETE` native, each reloading/stopping the **Rust** server — and only once no integration is still hosted by Go, or the two registries will disagree about which is running.
- `GET /api/integrations` and `/available-tools` (already native, #272) keep answering identically.

## Traps

- Reload is not restart-if-changed: Go reloads unconditionally on `PUT`. Reproduce the timing, not just the end state — a tool call in flight during a reload has an observable outcome.
- Credentials are read here for the first time. #272 deliberately never read them and #277 stored one without ever decoding it; this is where they get used, so the scrubbing rules apply on every path out.

**Blocked by:** #282

---
**Branch:** target `desktop`, **never `main`**. One PR per issue, merged before the next starts.
**Bar:** byte-identical JSON. Any field-name, key-order, escaping or float-spelling drift is a regression.
**Verify:**
```bash
cd desktop && eval "$(./scripts/parity-instance.sh start)"
(cd src-tauri && cargo test --test live_parity -- --ignored --nocapture)
./scripts/parity-instance.sh stop
```
**Read first:** `desktop/CLAUDE.md`, then the seam rules in `src-tauri/src/native/mod.rs` and `native/writes.rs`.
Never diff against the server on :8990 (it is a stale binary). Seed the scratch instance if the endpoint has no local data, or an empty response diffs clean and proves nothing.

Issue #310, opened 2026-08-16
Port: the local in-process tools MCP server (internal/tools)


vibexp-agent:
## Scope

`internal/tools/` — Agento's **local in-process MCP server** (`registry.go`, `current_time.go`), the one an agent's `capabilities.local` allowlist points at.

## Why it blocks things

`native/chat/runner.rs:74` refuses the turn outright:

> `"agent uses local or MCP tools, which are not ported yet (#277/#282)"`

So today every chat with a local tool or any integration tool forwards to the Go sidecar. This is the smaller half of that refusal and the one with no external dependencies — no OAuth, no network, no credentials.

## Acceptance

- The local server is hosted on whatever #282 picks, registered the way Go registers it, and reachable from a native chat turn.
- `runner.rs`'s refusal narrows to *integration* tools only.
- A chat turn using a local tool runs natively end to end, with the tool-call blocks stored byte-identically to Go's.

## Traps

- The typed-tool layer (`ToolDef`/`NewTool`/`ToolServer`) is exactly what #282 defers a decision on. This issue is its first real caller — if #282 chose `rmcp`, this is where `schemars` earns its place.
- Tool names on the wire are part of the agent's allowlist and of every stored `tool_use` block. Renaming one silently breaks existing agents.

**Blocked by:** #282

---
**Branch:** target `desktop`, **never `main`**. One PR per issue, merged before the next starts.
**Bar:** byte-identical JSON. Any field-name, key-order, escaping or float-spelling drift is a regression.
**Verify:**
```bash
cd desktop && eval "$(./scripts/parity-instance.sh start)"
(cd src-tauri && cargo test --test live_parity -- --ignored --nocapture)
./scripts/parity-instance.sh stop
```
**Read first:** `desktop/CLAUDE.md`, then the seam rules in `src-tauri/src/native/mod.rs` and `native/writes.rs`.
Never diff against the server on :8990 (it is a stale binary). Seed the scratch instance if the endpoint has no local data, or an empty response diffs clean and proves nothing.

vibexp-agent:
Done — PR #343 merged into `desktop`.

**Branch:** `feature/issue-310-port-local-in-process-tools-mcp`
**PR:** https://github.com/shaharia-lab/agento/pull/343

- `internal/tools` — the local in-process MCP server — is hosted natively on the `rmcp` typed-tool layer #282 chose, registered under `local-tools` with the tool resolving as `mcp__local-tools__current_time`, exactly as Go registers it.
[...]

PR #346, opened 2026-08-16
feat(desktop): port the GitHub integration MCP server (#312)

### LINKED ISSUE #312: Port: the GitHub integration MCP server
## Scope

`internal/integrations/github/` — Agento's GitHub integration, an in-process MCP server (1281 lines of Go).

One of six sibling issues, one per backend. **They are independent of each other** — all six are blocked only by #282, and can be worked in parallel by six agents once the MCP host is chosen.

## Acceptance

- Every tool the Go server exposes, under the **same tool name and the same input schema** — the names travel in each agent's `capabilities.mcp` allowlist and in every stored `tool_use` block, so a rename silently breaks existing agents and rewrites history.
- Tool *results* are what the model reads and what gets persisted; the parity bar is the result payload, not just the schema.
- Credentials come from the stored integration row and never appear in a response or a log line.
- The largest of the six. Token auth only, so it has no OAuth dependency — a good first one to take.

**Blocked by:** #282

---
**Branch:** target `desktop`, **never `main`**. One PR per issue, merged before the next starts.
**Verify:**
```bash
```
**Read first:** `desktop/CLAUDE.md`, then the seam rules in `src-tauri/src/native/mod.rs` and `native/writes.rs`.
Never diff against the server on :8990 (it is a stale binary). Seed the scratch instance if the endpoint has no local data, or an empty response diffs clean and proves nothing.

vibexp-agent:
## Summary

- Ports `internal/integrations/github` — Agento's largest in-process MCP server, **20 tools across five service groups** — onto the `rmcp` typed-tool layer #282 settled. This is that layer's second caller and its first with credentials and a network.
- The parity bar is the whole payload, not just the schema: the advertised surface, **the request each tool builds**, and every result and error sentence are pinned to vectors taken from the **live Go server** — a real MCP client over the real HTTP transport, against a scripted fake GitHub that records what each tool sent.
- Lands the **reflector divergence map** #310 asked for before the integration ports start, so #313–#317 inherit it rather than each rediscovering where `schemars` and `google/jsonschema-go` disagree.
- Scope stops at the server. The registry lifecycle and the `PUT`/`DELETE` routes are **#311**; `runner.rs`'s MCP refusal is deliberately untouched.

## The finding

`jsonschema:"required,Repository owner"` **is not a directive.** `google/jsonschema-go` reads the whole tag as the property description and marks a field optional only on `omitempty`/`omitzero`. No struct in the integration has either — so every one of the 20 tools advertises descriptions that literally begin `required,`, and **every field of every tool is required**.

That is what the Go server has always sent the model, so it is what the port reproduces. It also collapses the `Option<T>` problem for #312 entirely: all fields are plain `String`/`i64`/`bool`. The vectors record it, so a Go-side fix has to move both languages at once and cannot be done silently.

## Changes

**The port** — `desktop/src-tauri/src/native/integrations/github/`
- `mod.rs` — server name (`github-<id>`), `build_allowed_set` / `service_enabled`, registration order, `start_github_mcp_server`. Keeps Go's `len(allowed) > 0` guard: an empty allowed set hosts everything.
- `client.rs` — the redirectable API base, the two 15s clients (one no-redirect), `call` / `call_raw` / `get_redirect_url`, the three read caps, `split_csv`, the per-tool paging clamp, cancellation.
- `body.rs` — `json.Marshal`-over-a-Go-map request bodies (key-sorted), and `encoding/json`'s unmarshal error vocabulary for `trigger_workflow`'s inputs.
- `{repos,issues,pulls,actions,releases}.rs` — the 20 tools, one module per service group.
- Response bytes are passed through **verbatim** — no `serde_json::Value` round trip — so a body with unsorted keys, an integer past 2^53 and `1.50` reaches the model exactly as Go sent it. The vectors assert on precisely such a body.

**Shared machinery**
- `native/gourl.rs` — `path_escape` (`encodePathSegment`) and `query_escape` (`encodeQueryComponent`, space → `+`) plus `gourl::Values`, whose `Encode` sorts keys as Go's does. `form_urlencoded` was **not** usable: it escapes `~` where Go does not and keeps `*` where Go does not. Both are now in `gourl_vectors.json`.
- `claude/tool.rs` — `strip_schema_dialect` becomes `normalize_go_schema`, dropping three keys the Go reflector can never emit: `$schema`, `format` (schemars' `int64`) and `default`. The walk is structure-aware rather than a key sweep — all three are legal property names, and there is a test for a field called `format`. The `Arc` is still **replaced**, never written through, for #310's reason.
- `Cargo.toml` — `reqwest` gains `rustls-tls`. Bundled roots over native roots so an integration cannot work on one machine and fail on another over a differently-arranged system store; `native-tls` rejected for the reason already written on `lettre` (five release triples, no C toolchain).

**Parity surface**
- `desktop/parity/github_vectors.json` + `github_parity_test.go` — 20 tool schemas, 6 hosting cases (disabled service, non-empty allowlist, empty-set-means-everything), 44 live call cases including non-2xx, a 302 with and without `Location`, and malformed workflow inputs. Regenerate with `-update-github-vectors`; otherwise byte-compared, so a Go-side change fails Go's own suite.
- `desktop/parity/jsonschema_reflect_vectors.json` + `jsonschema_reflect_parity_test.go` + `desktop/src-tauri/src/claude/schema_vectors.rs` — the divergence map, per shape class, pinning what a port must write instead.
- `internal/integrations/github/parity.go` — `SetAPIBase`, the only new Go export: a thin seam so the vectors come from `github.Start` rather than from a restatement of it, exactly as #310 did with `FormatCurrentTime`.

**Divergences kept, each documented at its site** (all unreachable from this issue): Go slice → `["null","array"]` vs bare `array` (no integration takes a list — every multi-value input is a CSV string); nested struct inlined vs `$defs`/`$ref` (every params struct is flat); sized/unsigned ints (use `i64`); pointer field ordering and required-ness; `encoding/json`'s syntax-error vocabulary; and the Go MCP SDK **rounding an integer argument above 2^53** through its `map[string]any` defaults pass, which `rmcp` does not — unreachable with real GitHub run ids, and degrading Rust to match would be worse.

**Not ported, by design:** `ValidatePAT`, and `Start`'s auth check and credential parse — they have no native caller until #311 reads credentials for the first time. Noted in the module doc naming that issue.

## Testing

- [x] `make test` — passes, including `TestGitHubVectors`, `TestJSONSchemaReflectVectors` and the extended `TestGoURLVectors`
- [x] `cargo test` — 728 lib + 41 integration + 2 doc, **0 failed**
- [x] `cargo clippy --all-targets -- -D warnings` — clean; `cargo fmt --check` — clean
- [x] `golangci-lint run` on both changed Go packages — **0 issues in this PR's files**. The one finding on `gourl_parity_test.go` is the identical pre-existing line on `origin/desktop`, untouched by this PR's hunks: the known local v2.12.2 vs CI-pinned v2.5.0 drift.
- [x] All three vector files regenerate **byte for byte**
- [x] Credentials: a test asserts a failing call's message carries no token

## Related

Closes #312
Unblocked by #282 (PR #325); builds on #310 (PR #343). Leaves #311 the registry and the routes.

---
🤖 Implemented by [Claude Code](https://claude.com/claude-code) via `implement-issue` workflow

### DIFF
--- a/desktop/CLAUDE.md
+++ b/desktop/CLAUDE.md
@@ -780,9 +780,15 @@ three by #310 and one by #312 — and all six inherit every one:
   `Arc`, not through it: `rmcp` memoizes one schema per input type and hands
   every route a clone, so an in-place edit would reach into a process-wide
   cache. The walk is structure-aware rather than a key sweep — all three are
-  legal property *names* too. `#[serde(deny_unknown_fields)]` is what produces
-  `additionalProperties: false`, which Go reflects onto every struct and its
-  server *validates* against.
+  legal property *names* too — and it follows every position `schemars` can put
+  a subschema in, including the ones nothing has reached yet
+  (`unevaluatedProperties`/`unevaluatedItems`, which 1.x emits for
+  `#[serde(flatten)]` under `deny_unknown_fields`, plus `if`/`then`/`else`,
+  `contains` and `dependentSchemas`): an unwalked keyword leaves a nested
+  `$schema`/`format`/`default` in a schema the model reads and nothing says so.
+  `#[serde(deny_unknown_fields)]` is what produces `additionalProperties:
+  false`, which Go reflects onto every struct and its server *validates*
+  against.
 - **The reflector divergence map is generated, and it is the file to read
   before porting an integration** (#312).
   `desktop/parity/jsonschema_reflect_vectors.json` reflects one reference struct
@@ -817,6 +823,18 @@ three by #310 and one by #312 — and all six inherit every one:
   differs: Go's `validating "arguments": …` against `rmcp`'s `failed to
   deserialize parameters: …`. It is a property of `new_tool`, so every ported
   tool has it; there is no missing conversion to add.
+
+  Nothing in this port implements it, though, and that is the part worth
+  knowing: `into_tool_argument_error` **prefix-matches a hardcoded string**
+  (`"failed to deserialize parameters:"`) against its own extractor's message,
+  so an `rmcp` upgrade rewording either half would flip all 62 ported tools to
+  protocol errors at once — which the CLI renders as "Tool result missing due
+  to internal error", nothing for the model to retry against.
+  `malformed_arguments_are_a_tool_error_rather_than_a_protocol_error`
+  (`native/integrations/github/tests_vectors.rs`) sends a missing field and an
+  unknown field and pins the **kind**, deliberately not the text. Every ported
+  integration inherits the property from `new_tool`, so one test covers all of
+  them; add another only if a port stops going through `new_tool`.
 - **A tool's name is not renameable.** `mcp__local-tools__current_time` is in
   agents' stored `capabilities.local` allowlists and in every `tool_use` block
   already written to `chat_messages`. `desktop/parity/local_tools_vectors.json`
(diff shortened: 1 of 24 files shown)


vibexp-agent:
## Review round

A full review returned **REQUEST CHANGES** on one blocking defect. Both it and every SHOULD FIX are addressed in `da1c2b3`, along with all five CONSIDER items.

### The MUST FIX — dot path segments built a different request in Rust than in Go

`reqwest`/`url` applies WHATWG dot-segment removal for http/https; Go's `net/http` does not normalize, and `url.PathEscape` leaves `.` and `..` untouched. So identical arguments diverged after the path was built:

```
Go   built /repos/../../issues  → server saw /repos/../../issues
Rust built /repos/../../issues  → sent       /issues            *** DIVERGES ***
```

Escaping does not help — `%2E%2E` is collapsed too. `owner`/`repo`/`workflow_id` are model-supplied and tool results carry attacker-authored GitHub content, so it is reachable under prompt injection. The consequence was scope confusion on a PAT-bearing request: `list_issues(owner="..", repo="..")`, meant to be scoped to one repository, instead called `/issues` — the authenticated user's issues across **every** repository. The write tools too.

**Pinned before it was closed.** Five call vectors now cover `..` and `.` across all three URL builders (`call`, `callRaw`, `getRedirectURL`) and the write path. The first run failed worse than predicted — `left: "/"`, `right: "/repos/../.."`: `reqwest` was sending to the API root, not merely a shorter path. Note the subtlety: adding `..` to the `gourl` vectors would **not** have caught this, since `path_escape("..") == ".."` in both languages and the normalization happens downstream in `reqwest`.

**Closed** by `client::absolute`, applied at all three sites: it compares the parsed URL's path and query against the ones the tool built and refuses with the site's own `request failed` when they differ. `reqwest` offers no way to send an unnormalized target, so refusing is the faithful option — reaching a *different* endpoint than Go would is the one outcome worse than not calling. Recorded as an explicit divergence (`rust_no_request` + `rust_text`) rather than hidden.

The exact comparison is safe rather than lucky, and all 50 call vectors pass: Go's `encodePathSegment` already escapes every byte in `url`'s path encode set and `encodeQueryComponent` every byte in its query set, so `url` has nothing left to percent-encode; it preserves existing escapes case and all; a path with no `?` compares `""` against `query()`'s `None`; and host/scheme/port are never compared, so default-port and case normalization cannot reach the guard. A `..` inside a query *value* is correctly left alone.

### SHOULD FIX
- **The malformed-argument invariant is now tested rather than assumed.** `CLAUDE.md` promises a bad field yields `CallToolResult{IsError}` and never a JSON-RPC error, which rests entirely on `rmcp`'s `into_tool_argument_error` **prefix-matching a hardcoded string**. An upgrade rewording it would flip all 62 ported tools to "Tool result missing due to internal error" and nothing would have noticed. New test pins the *kind*, not the text.
- **TLS root store now matches Go.** `rustls-tls` is an alias for `rustls-tls-webpki-roots` — a compiled-in Mozilla snapshot — where Go uses the platform store. The old comment's justification ("public SaaS APIs, not behind a corporate MITM proxy") was the case that breaks: such a proxy intercepts `api.github.com` too and its CA lives only in the system store. Switched to `rustls-tls-native-roots`, which is equally pure-Rust and toolchain-free, so the five-triple reasoning is untouched. **Zero new packages** — `rustls-native-certs` was already in the lockfile via `tauri-plugin-updater`. Consequence handled: reqwest reports an unusable store as a *builder* error where Go's `&http.Client{}` cannot fail, so the clients are now `OnceLock<Option<Client>>` and answer `request failed` rather than panicking inside a detached `rmcp` handler.
- **A comment that described behaviour the code did not have** is gone: `set_api_base`/`API_BASE`/`api_base()` are now `#[cfg(test)]` outright, which also closes the "test seam in the production binary" concern on the Rust side.
- **Three comments claiming `tools/list` carries registration order** are corrected — both SDKs sort by name (`rmcp` `router/tool.rs:588`, go-sdk `featureSet.sortedKeys`), which is why the vectors' `tools` array starts at `create_issue`. Registration order is still pinned, by `GITHUB_TOOL_NAMES`; only the explanation was wrong.

### CONSIDER — all five taken
- **gzip.** Go's transport sends `Accept-Encoding: gzip` and decompresses transparently; reqwest with `default-features = false` sent none. Added. Confirmed reqwest decompresses in the service stack before `bytes_stream()`, so `read_capped` still caps *decompressed* bytes exactly as Go's `io.LimitReader` over `resp.Body` does — the cap semantics do not move.
- **`normalize_go_schema`** now walks `unevaluatedProperties`/`unevaluatedItems`/`if`/`then`/`else`/`contains` and `dependentSchemas`. Unreachable from #312, but a nested `$schema`/`format`/`default` under one would have reached the model silently — a trap removed from the five remaining ports.
- **Zero-fraction floats** (`{"per_page": 30.0}`, which models do emit) are accepted by Go and rejected by Rust — Go's `applySchema` re-marshals `float64(30)` to `30` before the typed decode. Pinned as a vector beside the 2^53 case. Documented rather than fixed: accepting them cleanly needs a newtype on 21 integer fields *and* a hand-written `JsonSchema` impl to keep the schema inlined rather than lifted into `$defs`, which is a real schema-parity risk taken for a wording difference.
- **Go's `SetAPIBase`** keeps its export (a `parity` build tag would either break `go test ./...` or drop the vectors out of CI) but now opens with `Do not call outside tests.`
- **Non-ASCII `Location`** now uses `from_utf8_lossy` rather than answering "missing Location header".

### Not changed, deliberately
The `required,` prefix leaking into every property description is a genuine latent defect **in the Go original** — `jsonschema-go` reads the whole struct tag as the description. Reproducing it verbatim is the correct behaviour for a parity port; fixing it would change the advertised surface of every stored agent allowlist. It deserves its own issue against the Go side.

### Verification
- [x] `make test` — pass; `cargo test` — 730 lib + all integration targets, 0 failed
- [x] `cargo clippy --all-targets -- -D warnings` and `cargo fmt --check` — clean
- [x] `golangci-lint` — 0 issues in this PR's files (the one on `gourl_parity_test.go:224` is pre-existing on `desktop`, from #294)
- [x] All three vector files regenerate byte-identically
- [x] All 4 CI checks green

Issue #318, opened 2026-08-16
Port: the integration OAuth flows (start, status, validate)


vibexp-agent:
## Scope

The OAuth surface, deferred whole by PR #300 because its three routes move together or not at all.

| Route | Go source |
|---|---|
| `POST /api/integrations/{id}/auth/start` | `internal/api/integrations.go` |
| `GET /api/integrations/{id}/auth/status` | ” |
| `POST /api/integrations/{id}/auth/validate` | ” |

`GET /auth/status` is one of the four reads #272 deliberately left with Go, for this reason.

## Why all three or none

From PR #300:

> in-memory flow state, must open the user's **real browser**, and writes credentials Go's MCP servers consume.

- **In-memory flow state**: the pending authorization lives in the process that started it. Split the routes across two processes and `status` polls a map that will never be populated.
- **The consumer matters**: the credentials this writes are read by the integration MCP servers. While those are Go's, a Rust write is invisible to them.

## Traps

- **Consent opens in the user's real browser, not the webview.** `tauri-plugin-opener` is already wired for exactly this — inside the app's webview there is no address bar to check who is asking, and the provider's session would live in a container we throw away.
- The redirect lands on a **loopback listener**. Go's is the sidecar's; the desktop shell has its own proxy port. Whichever answers must be the one the authorization URL declared, and it must match what is registered with the provider.
- Token refresh is shared with the Google MCP server. One implementation, not two.

**Blocked by:** #311 (the credential consumers), which is in turn blocked by #282.

---
**Branch:** target `desktop`, **never `main`**. One PR per issue, merged before the next starts.
**Bar:** byte-identical JSON. Any field-name, key-order, escaping or float-spelling drift is a regression.
**Verify:**
```bash
cd desktop && eval "$(./scripts/parity-instance.sh start)"
(cd src-tauri && cargo test --no-fail-fast -- --ignored --nocapture)
./scripts/parity-instance.sh stop
```
**Read first:** `desktop/CLAUDE.md`, then the seam rules in `src-tauri/src/native/mod.rs` and `native/writes.rs`.
There is no `--test live_parity` target: parity suites are one binary per area (`tests/parity_<area>.rs`), and `cargo test` stops at the first failing binary, so a real sweep needs `--no-fail-fast`.
Never diff against the server on :8990 (it is a stale binary). Seed the scratch instance if the endpoint has no local data, or an empty response diffs clean and proves nothing.

vibexp-agent:
[...]

Issue #315, opened 2026-08-16
Port: the Slack integration MCP server


vibexp-agent:
## Scope

`internal/integrations/slack/` — Agento's Slack integration, an in-process MCP server (675 lines of Go).

One of six sibling issues, one per backend. **They are independent of each other** — all six are blocked only by #282, and can be worked in parallel by six agents once the MCP host is chosen.

## Acceptance

- Every tool the Go server exposes, under the **same tool name and the same input schema** — the names travel in each agent's `capabilities.mcp` allowlist and in every stored `tool_use` block, so a rename silently breaks existing agents and rewrites history.
- Tool *results* are what the model reads and what gets persisted; the parity bar is the result payload, not just the schema.
- Credentials come from the stored integration row and never appear in a response or a log line.

**Blocked by:** #282

---
**Branch:** target `desktop`, **never `main`**. One PR per issue, merged before the next starts.
**Bar:** byte-identical JSON. Any field-name, key-order, escaping or float-spelling drift is a regression.
**Verify:**
```bash
cd desktop && eval "$(./scripts/parity-instance.sh start)"
(cd src-tauri && cargo test --test live_parity -- --ignored --nocapture)
./scripts/parity-instance.sh stop
```
**Read first:** `desktop/CLAUDE.md`, then the seam rules in `src-tauri/src/native/mod.rs` and `native/writes.rs`.
Never diff against the server on :8990 (it is a stale binary). Seed the scratch instance if the endpoint has no local data, or an empty response diffs clean and proves nothing.

vibexp-agent:
Ported in #358, merged to `desktop` as 96caa4b87.

`native/integrations/slack/` — seven tools in one service group, over a workspace token. `slack` joined `registry::HOSTED_TYPES`, so the sidecar now runs with `AGENTO_INTEGRATIONS=off:github,confluence,jira,slack`.

Two things here that no earlier port had:

- **The token can come from the `auth` column.** `resolveToken`'s `oauth` arm reads `ParseOAuthToken()`, so `HostingRow` now carries that column — which until now was collapsed to a SQL boolean precisely so a stored token could not exist in the process to be echoed. The constraints on it are unchanged: no `Serialize`, no `Debug`, private, only a `&str` leaving. There is also a third arm a port drops: an unrecognized `auth_mode` falls back to a non-empty bot token.
[...]

PR #360, opened 2026-08-17
feat(desktop): port the Google MCP server, measured rather than read (#313)

vibexp-agent:
Closes part of #313 — **the port. The hosting flip is a follow-up PR.**

Eight tools across three service groups (`calendar`, `gmail`, `drive`) over an OAuth2 token that refreshes itself. Last of the six integrations, and the only one that is not a port of hand-rolled HTTP.

## Why this one is measured rather than read

The other five build their requests with `http.NewRequest` and `json.Marshal`, so the port reproduces bytes visible in `tools.go`. Google calls the **generated** client libraries (`calendar/v3`, `gmail/v1`, `drive/v3`) over an `oauth2` transport, and what those put on the wire is in neither this repository nor the port.

So every URL, query parameter, body field and sentence was *recorded* off the real libraries against a fake endpoint before it was written. `desktop/parity/google_vectors.json` is not a confirmation of the port — it is the **only written specification** of what that third-party code generator emits, and it is what will notice when the generator changes.

## Four things this found that no amount of reading would have

Three were port bugs, caught the first time the vectors were replayed:

1. **Path parameters use `googleapi.Expand`, not `url.PathEscape`.** The generated clients expand RFC 6570 templates, escaping everything outside the unreserved set. `PathEscape` leaves the sub-delims alone — so a Gmail message id containing `&` came out as `…%3Fd&e`, an `&` that starts a query parameter.
2. **Every JSON body carries a trailing newline**, because `googleapi` writes it with `json.NewEncoder(…).Encode`. Both the plain `application/json` POSTs and the metadata part of the `multipart/related` upload.
3. **`oauth2.RetrieveError` has two forms** and picks between them on whether the body named an error code, *not* on the status. A Google refresh refusal takes the one the port did not have: `oauth2: "invalid_grant" "Token has been expired or revoked."`.

The fourth is Go's, and was found by trying to vector it:

4. **A response body of exactly `null` panics every one of the eight Go tools.** The generated clients decode into a `**T`; `json.Unmarshal` of `null` into a pointer-to-pointer nils the pointer, and the handler reads a field off it. That is a 200 with a two-word body — what a proxy or a misconfigured gateway produces. It joins the two nil-dereferences the handlers own (`msg.Payload.Headers`, `ev.Start.DateTime`). The port returns the zero value in all three. The case is deliberately **absent** from the vectors: recording it would mean recording a crash.

## Google-specific behaviour the other five never reached

- **`gourl::Values` stopped being single-valued for this.** Gmail's `MetadataHeaders("Subject","From","Date")` is three `metadataHeaders=` pairs in insertion order under one sorted key, so `Values` is now a multimap with `set` and `add`.
- **`search_email` makes N+1 requests from one tool call**, skips a failed fetch rather than surfacing it, and reports `len(list.Messages)` — the *listed* count. A partial failure produces a sentence whose number does not match its body. Go's, and pinned.
- **`create_file`'s media type is sniffed from the content**, not taken from `mime_type`, which reaches the metadata JSON alone.
- **Drive's upload path is an absolute reference**, replacing the base's whole path: `/upload/drive/v3/files`.
- **Every clamp differs** (100 / 100 / 50, all falling back to 10), **`q` is conditional for Drive and unconditional for Gmail**, and **`download_file` is the only call that is not `alt=json`**.

## Pinned as divergence rather than matched

All recorded in the vectors so they cannot drift silently: `X-Goog-Api-Client` and `User-Agent` (they embed the Go toolchain and library versions — no Rust build can emit `gl-go/…`), the random `multipart/related` boundary (the *parts* are compared instead), and Go's `*url.Error` wrapper around a transport or refresh failure.

One value is not deterministic — `view_events`' `time_min` defaults to `time.Now()` — so that vector records `«now»`, and both languages assert the value it replaced is a seconds-precision RFC3339 instant in UTC before substituting.

## The Go side

`internal/integrations/google/parity.go` exports `SetEndpoints`: a **wider** seam than the other integrations' `SetAPIBase`, because both the API base and the OAuth2 token endpoint have to be redirectable at a fake. It is test-only — the token URL receives the `client_secret` and refresh token, the durable credentials. The three registrars now build through `clientOptions`. No behaviour or wording changes; `Start` remains the only way the app builds this server.

## Why `google` is not in `HOSTED_TYPES`

Nothing here runs in a shipped build. The sidecar still hosts Google and this is dormant code with its parity pinned. The flip is its own PR, because the flip is where the risk in this series has actually lived — #315's hosting of Slack silently broke `completeOAuth`'s reload, and Google is the *other* provider `startProviderCallback` supports, so it lands on the same hook. Splitting lets that be reviewed on its own and reverted without reverting the port.

`client::TokenSource` is also the refresh implementation #318 says must not be written twice.

## Gates

`cargo test` (896 + 19 + 15 …, all green), `cargo fmt --check`, `cargo clippy --all-targets -- -D warnings`, `go test ./...`, `golangci-lint run ./...` at the pinned v2.5.0.

### DIFF
--- a/desktop/CLAUDE.md
+++ b/desktop/CLAUDE.md
@@ -600,7 +600,7 @@ of the `Err` arms the cut-over has to turn into a real response.
 | 1 ✅ | Sidecar + proxy | — | done |
 | 2 ← | Pricing + analytics | `internal/pricing`, `internal/claudesessions` | Pure computation over JSONL + SQLite. No external deps. **In progress**: `/api/pricing/catalog`, `/api/claude-sessions`, `/api/claude-sessions/facets`, `/api/claude-analytics`, `/api/claude-sessions/insights/summary` and the agent reads are native and diff clean. |
 | 3 ← | Storage + tasks | `internal/storage`, `internal/scheduler` | **In progress (#274, #275).** `db.rs` is read-write, the 27 migrations are ported, and the writes whose every effect Rust owns are native — see below. The scheduler's *schedule computation* is ported and pinned (#275); its ownership is not, and is blocked — see below. |
-| 4 ← | Integrations | `internal/integrations`, `internal/trigger` | OAuth2 + MCP servers. Six of them: google, github, slack, jira, confluence, telegram, plus `internal/tools`. **How to host one is settled (#282)** — `claude::ToolServer`, see "Hosting a tool" below; do not invent a second way. **`internal/tools` is done (#310)** — `native/tools/`, and it is the worked example the six should be read against. **GitHub is done (#312)** — `native/integrations/github/`, the worked example for an *integration*. **The registry is done (#311)** — `native/integrations/registry.rs` plus `PUT`/`DELETE /api/integrations/{id}`, and the sidecar now runs with `AGENTO_INTEGRATIONS=off:<the types the shell hosts>`, so every type has exactly one owner. **Confluence is done (#317)** — `native/integrations/confluence/`, the smallest of the six and the one that shows what an integration adds over #312: a per-row API base, basic auth, and a `Start` check that is a decision. **Jira is done (#316)** — `native/integrations/jira/`, Confluence's twin, and the one that shows that a shared credential type does not mean shared behaviour: `jira.Start` validates nothing, so a bad base is answered per call instead of by refusing to host. **Slack is done (#315)** — `native/integrations/slack/`, the first that is not Atlassian-shaped: no model input in the URL at all, and the only integration whose token can come from the `auth` column. **Telegram is done (#314)** — `native/integrations/telegram/`, the outbound half, and the one that reached two reflector divergences the map had recorded as unreachable. #313 is the last, and is "add a starter **and its name to `HOSTED_TYPES`**", which is the one list both processes are configured from. WhatsApp is **not** among them — but Go still hosts its rows, because its starter opens a live connection rather than just a port; see below. |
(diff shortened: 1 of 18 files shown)


vibexp-agent:
## Review round: eight divergences found, all confirmed, all fixed

Review flagged eight behaviours where the port differed from Go. Rather than fix from the review's reading, I added **a vector case for each** and read what the real Go answered — every one was confirmed. The vectors grew by sixteen call cases and one refresh case.

The interesting part is that two of them were the *same mistake*: in both places the port had written a **summary** of a Go function instead of a transcription of it, and in both places the summary was wrong in cases it did not enumerate.

### `googleapi.Error.Error()` — a four-case `match` with three wrong branch conditions

- **The one-error form is conditional** on the sub-error's message *equalling* the top-level one. The normal validation-error shape has them differ (generic top level, specific item) and takes the `More details` branch. The original vector passed only because its fixture used the same message in both places.
- **`error.details` was dropped entirely** — the modern `google.rpc.ErrorInfo` / `QuotaFailure` envelope, which is what a real quota error carries. Go renders it as an indented JSON block.
- **The raw-form test is "no errors *and* no message"**, not "no code and no message", so an `errors` array alone takes the structured branch with the code defaulting to the HTTP status.
- **A body beginning `[`** is unwrapped by `errorReplyFromBody`.

### `detect_content_type` — "the reachable subset" was not reachable-subset-shaped

It was written as the subset a JSON string can reach, reasoning that a PNG's `0x89` would be UTF-8 encoded before it arrived. Sound for `0x89`, wrong for most of the table: `BM`, `%PDF-`, `%!PS-Adobe-`, `GIF87a`/`GIF89a`, `ID3`, `OTTO`, `ttcf`, `wOFF`, `wOF2`, `RIFF…WAVE` and `FORM…AIFF` are pure ASCII, and `PK\x03\x04`, `MThd`, the icon signatures and the 34-NUL EOT one are reachable through a `\u0000` escape.

**`BM` is the one that bites**: content beginning "BMI calculator results…" uploads as `image/bmp`. It also skipped the **512-byte bound**, so a control byte anywhere in a long document flipped the part to `octet-stream`.

Now the whole table, in order, with all four match flavours and `sniffLen`.

The note in each header states the lesson, because it generalizes: an argument that some subset of a third party's table is unreachable has to be right about *every* entry, and transcribing is cheaper than being right.

### The other six

- **`omitempty` reaches past `description`** — `calendar.Event.Summary`, `EventDateTime.DateTime` and `drive.File.Name` carry it too, so an empty `start` sends an object holding nothing but its time zone.
- **`tokenRefresher` refuses before opening a socket** when the stored token has no refresh token (the state a re-consent without `prompt=consent` leaves a row in). Reproducing it is not only about the sentence: without it the port POSTs the user's `client_secret` on a request Go never sends.
- **Gmail's `parts` is the one array where a `null` element is graceful**, because `extractBody` nil-checks it — so it is `Vec<Option<GoStruct<_>>>` where its siblings are not.
- **`googleapi.Error.Errors` needs the same**, because it is a Go **value** slice: a `null` element is the zero item, not a decode failure. (An array element is still refused, which is what keeps `GoStruct` meaningful there.)
- **An empty 2xx body is a zero value**, not a decode failure — `DecodeResponse` returns untouched on a 204.
- **The multipart boundary now comes from a CSPRNG** (found before the review landed): it was derived from clock + thread id, and since `create_file`'s `content` is a tool argument, a predictable boundary is a multipart-injection surface. Go uses `crypto/rand`; this now uses two v4 UUIDs.

### Deliberately left standing, and now written down

Four small ones, in `client.rs`'s header rather than left implicit: dot-segment normalisation in a path parameter (fixing it means not using `url::Url` at all, since both `parse` and `join` normalise), `status_line`'s canonical reason phrase vs Go's wire phrase, the hardcoded `Bearer` scheme, and `expires_in` having to be a JSON number. Plus the 16 MiB resumable-upload threshold this port does not implement — reaching it means a 16 MiB tool argument.

### What the review checked and found correct

Worth recording, since it is most of the surface: every clamp; the conditional/unconditional `q` split; every field mask; the three `metadataHeaders` in insertion order under a sorted key; `search_email`'s N+1 with skip-on-failure and the listed count; the last-header-wins switch; `extractBody`'s padded-base64 rejection; `%q` quoting; `download_file`'s `alt=media` and verbatim body; the three bases and the absolute upload reference; `expand_path_segment` against `uritemplates`' `unreserved` regexp; the whole `doTokenRoundTrip` ordering; `Token::valid`'s delta and zero-expiry rule; and that the Go-side `clientOptions`/`SetEndpoints` seam is inert in production.

No credential reaches a tool result or a log line — the module writes nothing to a log at all.

PR #361, opened 2026-08-17
feat(desktop): host google, closing the integration phase (#313)

### LINKED ISSUE #313: Port: the Google integration MCP server

vibexp-agent:
Closes #313.

The second half. The port landed in #360 as dormant code with its parity pinned; this adds `google` to `HOSTED_TYPES` and gives it a starter.

**With this, all six integration types are hosted by the Rust shell** and `hosting_env_value()` tells the sidecar to drop every one of them. Only `whatsapp` stays Go's — dropped rather than deferred, because its starter opens a live whatsmeow connection registered in a package global that the status, reconnect and QR endpoints read.

## Why the flip was worth splitting out

The flip is where the risk in this series has actually lived. #315's hosting of Slack silently broke `completeOAuth`'s reload, and Google is the *other* provider `startProviderCallback` supports — so it lands on the same hook.

With both hosted, `reload_if_secrets_changed` now covers **every** OAuth integration in the product. There is no longer a case it does not have to catch, which is a better place to be than the one #315 left us in.

Google is also the one where being served late costs the most, for a reason peculiar to it: **a Google token expires**. A Slack row served at the next boot is merely served late; a Google row carries an `expiry` that `TokenSource` reads, so a token first picked up hours later may already need a refresh on the first tool call. That still works — it is what the refresh is for — but it is worth knowing when reading a bug report about a slow first Google call.

## `start_google` is the only starter needing both secret columns

`credentials` carries the OAuth2 client pair; `auth` carries the token. Slack reads `auth` too, but only for an access token. Here the whole `oauth2.Token` matters, because `expiry` decides whether the first tool call refreshes.

That is also why **`google_oauth_token` refuses a row whose `expiry` does not parse**, where Slack's equivalent deliberately skips the field. A corrupt expiry means *not hosted* — not hosted with a token that never refreshes.

## The boundary is measured, not read

This is the lesson the first half taught, applied up front. A new `starting` section in `google_vectors.json` records what the **real** `google.Start` does with twenty-three credential and auth shapes, and a Rust test replays it.

That is how `parse_go_rfc3339` came to exist. **`chrono` is more permissive than Go in two ways a stored token can actually reach:**

| value | Go | chrono |
|---|---|---|
| `2030-01-01t00:00:00z` | parse error (strict since 1.20) | accepted |
| `2030-06-30T23:59:60Z` | `second out of range` | accepted |

Neither was guessable from either side's source. The rest agree: a numeric offset is accepted, fractional seconds of any length are accepted, and a trailing space, an out-of-range month, a date-only value, an empty string and a numeric expiry are all refused. A `null` or absent expiry is Go's zero `time.Time`, which never expires.

Some other things the vectors settled, all Go's behaviour and none of it obvious:

- An **empty** `auth` column and a literal `null` are both "no auth token", refused before any parse — but an empty **object** `{}` is authenticated, parses, and hosts a server whose every call carries an empty bearer token.
- A `null` credentials blob is a no-op, so the client id ends up empty and the row still hosts.
- Neither client id nor secret is checked for emptiness: Go hosts a row with both empty and every refresh then fails at Google.

## Stand-ins

Three "unported type" tests move from `google` to `whatsapp`, where they stay. One of them said in a comment that google was *"the last move available"* — it was right.

Clippy caught two of these as `for loop over a single element`, which is a fair description of what a stand-in list becomes when only one type is left; both are now direct assertions.

## Gates

`cargo test` (898 lib + the rest, all green), `cargo fmt --check`, `cargo clippy --all-targets -- -D warnings`, `go test ./...`, `golangci-lint run ./...` at the pinned v2.5.0.

### DIFF
--- a/desktop/CLAUDE.md
+++ b/desktop/CLAUDE.md
@@ -600,7 +600,7 @@ of the `Err` arms the cut-over has to turn into a real response.
 | 1 ✅ | Sidecar + proxy | — | done |
 | 2 ← | Pricing + analytics | `internal/pricing`, `internal/claudesessions` | Pure computation over JSONL + SQLite. No external deps. **In progress**: `/api/pricing/catalog`, `/api/claude-sessions`, `/api/claude-sessions/facets`, `/api/claude-analytics`, `/api/claude-sessions/insights/summary` and the agent reads are native and diff clean. |
 | 3 ← | Storage + tasks | `internal/storage`, `internal/scheduler` | **In progress (#274, #275).** `db.rs` is read-write, the 27 migrations are ported, and the writes whose every effect Rust owns are native — see below. The scheduler's *schedule computation* is ported and pinned (#275); its ownership is not, and is blocked — see below. |
-| 4 ← | Integrations | `internal/integrations`, `internal/trigger` | OAuth2 + MCP servers. Six of them: google, github, slack, jira, confluence, telegram, plus `internal/tools`. **How to host one is settled (#282)** — `claude::ToolServer`, see "Hosting a tool" below; do not invent a second way. **`internal/tools` is done (#310)** — `native/tools/`, and it is the worked example the six should be read against. **GitHub is done (#312)** — `native/integrations/github/`, the worked example for an *integration*. **The registry is done (#311)** — `native/integrations/registry.rs` plus `PUT`/`DELETE /api/integrations/{id}`, and the sidecar now runs with `AGENTO_INTEGRATIONS=off:<the types the shell hosts>`, so every type has exactly one owner. **Confluence is done (#317)** — `native/integrations/confluence/`, the smallest of the six and the one that shows what an integration adds over #312: a per-row API base, basic auth, and a `Start` check that is a decision. **Jira is done (#316)** — `native/integrations/jira/`, Confluence's twin, and the one that shows that a shared credential type does not mean shared behaviour: `jira.Start` validates nothing, so a bad base is answered per call instead of by refusing to host. **Slack is done (#315)** — `native/integrations/slack/`, the first that is not Atlassian-shaped: no model input in the URL at all, and the only integration whose token can come from the `auth` column. **Telegram is done (#314)** — `native/integrations/telegram/`, the outbound half, and the one that reached two reflector divergences the map had recorded as unreachable. **Google is ported (#313)** — `native/integrations/google/`, the last and the only one whose requests are built by a generated client rather than by the handlers, so its vectors are a specification rather than a confirmation; it is deliberately **not** in `HOSTED_TYPES` yet, and that flip is its own change. WhatsApp is **not** among them — but Go still hosts its rows, because its starter opens a live connection rather than just a port; see below. |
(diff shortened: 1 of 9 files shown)


vibexp-agent:
## Review round: five findings, all confirmed, all fixed

Same method as the first half — each finding got a vector, and the real Go answered. All five were confirmed.

### The one that mattered: Go's zero time means *never expires*

`omitempty` does not suppress a struct. So `SetOAuthToken` emits `"expiry":"0001-01-01T00:00:00Z"` for a token with no expiry rather than omitting the key — and `Token.Valid()` treats that zero `time.Time` as **never expiring**. Measured:

```
json.Marshal(&oauth2.Token{AccessToken:"ya29.x", RefreshToken:"1//y"})
→ {"access_token":"ya29.x","refresh_token":"1//y","expiry":"0001-01-01T00:00:00Z"}
round-trip: IsZero=true  Valid=true
```

This port read it as an ordinary instant, which is permanently expired — the exact inverse. A Google row in that state refreshed on the first call of *every* server, and where the row also had no refresh token it failed every call outright.

The vector that was supposed to cover this tested the **absent** key, which is a spelling Go's own writer can never produce. That is the more useful half of the lesson: a vector for the shape you imagine is not a vector for the shape that occurs.

### The second: there were about to be two Go-RFC3339 parsers

`native/schedule` already had one for `run_at` (#275) — `chrono::parse_from_rfc3339` plus three guards, derived from a differential run of 636 shapes through a real `gocron.Scheduler`. This PR was adding a second of the same shape with two guards.

Review then showed **both guard lists were incomplete, in the same direction**. `chrono` also *refuses* three forms `time.Parse` accepts, because `parseStrictRFC3339`'s extra checks are compiled out behind `case true:`:

| value | Go | `parse_from_rfc3339` |
|---|---|---|
| `2026-06-01t12:00:00z` | error | accepted |
| `2026-06-30T23:59:60Z` | `second out of range` | accepted |
| `2026-06-01T12:00:00,5Z` | accepted | error |
| `2026-06-01T5:04:05Z` | accepted | error |
| `2026-06-01T12:00:00+24:00` | accepted | error |

The last three are Go being **laxer**, which a stricter port turns into refusing input Go takes — a schedule that will not build, or an integration that will not host.

So `gotime::parse_rfc3339` is now the single implementation, and it parses the grammar rather than wrapping a library. Guarding a convenient library has now been wrong twice for one reason: **the guard list has to be right about every disagreement**, and enumerating a third party's edge cases is the thing that keeps turning out incomplete. That is the same failure mode as the first half's `detect_content_type` and `googleapi.Error` — three times in one issue, which is why it is written down in the module header rather than just fixed.

`GoTime::parse` stays on chrono deliberately: it reads `effective_from`, which only this application writes, normalized to second precision.

### The other three

- **The `starting` replay was a tautology.** It re-implemented the starter's three checks instead of calling it, so deleting the auth precheck or swapping the two parses would have left it green. It now calls `google_start_inputs` — the function the starter itself calls — and two new vectors have more than one thing wrong, so the **order** is pinned rather than assumed. Sentences are compared exactly wherever Go's is reproducible; `googleRustStartErrors` is a data table of the fifteen that are not, so "which sentences diverge" is something a reader counts rather than something a classifier decides.
- **`oauth2.Token` types `token_type` and `expires_in`**, so `{"token_type": 5}` fails the whole document. A three-field struct here silently hosted rows Go refuses.
- **Nothing exercised the starter's happy path.** `a_hosted_type_always_has_a_starter` inserts empty credentials, so for google it only ever reached the `credentials are empty` arm. A swapped client id and secret would have surfaced only at refresh time. Two tests now cover it, one of them asserting which column lands in which parameter.
- **Five doc sites still asserted the pre-flip state**, including one in a file this branch had already edited.

### Confirmed clean

The reviewer walked the flip's whole blast radius — sidecar env, both dispatch sites, `can_host`, `start_filtered_server`, the native `PUT`/`DELETE`, all three reload hooks, `available_tools` — and found nothing half-flipped. Credential leakage in the new paths: clean, and the assertion that checks it now has a vector carrying an actual secret, which it previously did not.

### Gates

`cargo test` (904 lib + the rest), `cargo fmt --check`, `cargo clippy --all-targets -- -D warnings`, `go test ./...`, `golangci-lint run ./...` at the pinned v2.5.0. All four CI checks green against the tip commit.

PR #377, opened 2026-08-19
fix(desktop): render AskUserQuestion prompts, drop redacted-thinking boxes

vibexp-agent:
Fixes the two live-chat defects from the report (`several_issues.jpg`): the agent's question never became an answerable prompt, and "Thinking" disclosures opened onto nothing. Both root-caused against a captured SSE stream from a real turn, both verified end-to-end in the running app. Frontend-only.

## 1. AskUserQuestion showed raw JSON, no prompt

Captured stream shows why: the CLI **disables AskUserQuestion in SDK sessions** (`No such tool available: AskUserQuestion. AskUserQuestion is disabled for this session`), so when the model calls it anyway the call is freehand — and the `user_input_required` frame then carries `questions` as a **JSON-encoded string**, not an array:

```
event: user_input_required
data: {"input":{"questions":"[{\"question\":\"Which color do you prefer?\",...}]"}}
```

`readQuestions` only accepted an array, silently produced zero questions, and the prompt rendered as a bare "Type your answer…" box with no question and no options — exactly the report. It now parses stringified `questions` (and a stringified `input`), covering both the freehand shape and the proper array shape if a future CLI enables the tool.

Also: the answer travels over `POST /input` and is not persisted server-side, so it vanished from the conversation. It is now echoed into the transcript as a user bubble for the rest of the sitting. (Persisting it durably needs coordinated Go+Rust backend changes — left as a follow-up.)

**Verified live**: forced an AskUserQuestion turn → prompt renders with header, question and clickable options → picking one and sending continues the same parked turn → model acknowledges the choice.

## 2. Empty "Thinking" boxes

Current models redact thinking entirely: every block arrives as `thinking: ""` plus a signature — in stream deltas and stored transcripts alike. Sampled the local corpus across model generations (fable-5, sonnet-5, opus-4-8): **0 of 388 thinking blocks carried any text**. The openable-but-empty box was showing something that does not exist client-side.

- Empty thinking blocks are no longer rendered — live stream, stored chats, and the session detail view.
- During a live turn the status line now shows **"Thinking… (~N tokens)"**, driven by the `system`/`thinking_tokens` estimate events — the one signal redacted thinking does emit.
- Models that do emit thinking text render exactly as before.

### DIFF
--- a/desktop/src-tauri/src/native/chat/turn.rs
+++ b/desktop/src-tauri/src/native/chat/turn.rs
@@ -7,9 +7,11 @@
 //! Four things can happen and the loop selects over all of them: an event
 //! arrives from the CLI, the permission handler asks a question, the permission
 //! handler asks for an allow/deny, or the client disconnects. There is **no
-//! timeout** and **no heartbeat** — a long tool call sends nothing for minutes,
-//! by design, which is exactly why the disconnect arm has to be explicit rather
-//! than inferred from a failed send.
+//! timeout** — a long tool call sends nothing for minutes, by design, which is
+//! exactly why the disconnect arm has to be explicit rather than inferred from
+//! a failed send. There **is** a heartbeat: a one-second SSE comment frame,
+//! because WebKitGTK strands a frame that arrives right before a silence (see
+//! [`HEARTBEAT`]); it is transport padding, never an event.
 //!
 //! # Rules that are silent when broken
 //!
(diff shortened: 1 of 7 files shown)


vibexp-agent:
Third commit (`f219abf`) closes the remaining gap the last report surfaced: the prompt rendered in Chrome but **never in the Tauri webview**, even with the parsing fix in place.

**Root cause, proven end to end in the real webview** (driven via WebKitGTK's remote inspector against the actual `target/debug/agento`): the backend emitted `user_input_required` (instrumented log), the app's stream reader never received another chunk — and answering `/input`, which makes the server write again, delivered the stranded frame together with the continuation. WebKitGTK's fetch on the app's long-lived connection only delivers a frame that arrives right before a silence once further bytes push it through, and `user_input_required` is by construction the last frame before an unbounded park.

**Fix:** the turn's SSE stream never goes quiet — a one-second `: hb` comment frame from the event loop and from the result-path park. SSE comments are event-invisible (the frontend's `parseFrame` drops them), so no event sequence changes; Go sends none and the divergence is transport padding. Also covers permission prompts and any long silent tool run, which would hit the same stall.

**Verified in the failing environment:** drove the real Tauri webview through the exact reported flow — question prompt rendered with header/options, clicked an option, Sent — answer echoed, model continued ("Got it — you picked **Black / espresso**"). `cargo fmt`/`clippy`/`test` green (one test taught to skip comment frames, as any SSE parser does).

PR #379, opened 2026-08-19
feat(desktop): native window chrome, folder pickers, and platform-convention fixes

vibexp-agent:
## Problem

The desktop window did not behave like a native app on any platform:

- **macOS**: no traffic lights (minimize/maximize/close) at all — `decorations: false` removes the native titlebar, and `titleBarStyle: "Overlay"` only takes effect when decorations are on.
- **macOS/Linux**: the window could not be moved. Dragging relied on `-webkit-app-region: drag` CSS (`shell.css`), which is an Electron feature — Tauri's WKWebView/WebKitGTK do not honor it. Tauri needs the `data-tauri-drag-region` attribute, which appeared nowhere.
- **Linux/Windows**: no native titlebar, only hand-drawn window-control buttons.
- **New chat**: the working directory had to be typed by hand.

### Window chrome is the OS's now
- `decorations: true` with macOS-only `titleBarStyle: "Overlay"` + `hiddenTitle` + `trafficLightPosition` (`{x:13,y:13}`, vertically centered in the 38 px strip). macOS gets real inset traffic lights over the app's titlebar strip; Linux/Windows get their ordinary decorated titlebar. The three overlay keys are documented macOS-only, so no `tauri.<platform>.conf.json` split is needed ([platform-specific config](https://v2.tauri.app/develop/configuration-files/#platform-specific-configuration) stays available if we ever diverge further).
- `TitleBar.tsx` uses `data-tauri-drag-region` (drag + double-click-maximize handled by Tauri); the hand-drawn min/max/close buttons and their CSS are gone, along with the dead `-webkit-app-region` rules.

### Native folder picker for every working-directory field
`tauri-plugin-dialog` + `pickDirectory()` in `lib/tauri.ts`, wired into:
- New Chat (Browse… next to the working-directory input) — the field non-technical users could previously only type into,
- Scheduled Task editor (same field, same problem),
- Settings → Workspace (the existing in-app `DirBrowser` is kept as the plain-browser fallback — it depends on `/api/fs`, which is Unix-only, so the native dialog also makes the picker work on Windows).

### Window state, single instance
- `tauri-plugin-window-state`: size/position/maximized survive relaunch instead of every launch being a centered 1280×820.
- `tauri-plugin-single-instance`: a second launch focuses the running window. Two processes on one `~/.agento/agento.db` are two schedulers and two SQLite writers — the exact collision `desktop/CLAUDE.md` documents for dev.

### macOS menu conventions
- A real application menu: About, Settings…, Services, Hide/Hide Others/Show All, **Quit ⌘Q** (previously the app had no Quit anywhere and "File" occupied the app-menu slot).
- The menubar is now macOS-only (`cfg(target_os = "macos")`): an in-window GTK/win32 menubar is not the convention for this class of app, and with native decorations on Linux it rendered as a second bar. Custom items still carry no accelerators on purpose — `App.tsx`'s keydown handler owns shortcuts, and WKWebView sees key equivalents before the menu, so declaring both double-fires.
- Removed the unimplemented Edit ▸ Find; Help ▸ Documentation / Star on GitHub now actually open (they emitted ids nothing handled).

### External links actually open
`window.open`/`target="_blank"` don't leave a Tauri webview. `openExternal` (promoted from IntegrationsView into `lib/tauri.ts`) now backs the About page buttons, session PR links, Settings release-notes link, and the Help menu.

### Keyboard + input polish
- Shortcuts are Mac/PC aware: `⌘` on macOS, `Ctrl` elsewhere — accepting either meant Ctrl+B/K/N hijacked macOS's own text-editing bindings (and the send chord in the composer had the same problem).
- Global chords no longer fire while typing in an input/textarea (except ⌘K / ⌘,).
- The blanket `contextmenu` suppression now exempts editable fields and selected text, so Copy/Paste/spellcheck are reachable again.
- Explicit Light/Dark now also reaches the native window (`setTheme`), so the OS-drawn titlebar follows the app theme; nested panes no longer rubber-band on macOS (`overscroll-behavior: contain` on `.scroll`).

### Packaging honesty
- `dragDropEnabled: false` — nothing listens for drops, and on Windows/Linux OS-level interception breaks HTML5 DnD inside the webview. Flip back when a drop target exists.
- Dropped `libayatana-appindicator3-1` / `libappindicator-gtk3` from deb/rpm depends — they exist for a tray icon the app doesn't have.

## Verified
- `npm run build` (tsc + vite) clean
- `cargo fmt --check`, `cargo clippy --all-targets -- -D warnings`, `cargo test` clean

## Known follow-ups (found during the audit, deliberately not in this PR)
1. **Windows 501 surfaces**: `/api/fs`, the Claude settings + profiles routes, and the config-dir probe answer 501 on Windows (`gopath.rs` is Unix `filepath` only). The native dialog removes the user-facing pain for directory picking; the rest needs a Windows `filepath` port.
2. **No SIGTERM on Windows** for the Claude CLI (`claude/process.rs`): `terminate()` is a no-op there, so shutdown waits the full grace period then hard-kills.
3. **About page shows "Version dev"** in shipped builds — `AGENTO_BUILD_VERSION` is never stamped by `desktop-release.yml`; `host_info` already carries `CARGO_PKG_VERSION`.
4. **Scrollbars are invisible until hover** and `scrollbar-color` suppresses the `::-webkit-scrollbar` rules on WebView2 — needs a per-platform pass.
5. **Cursor conventions**: `cursor: default` on buttons is macOS-correct but Windows/Linux expect a pointer; needs a platform class on the root.
7. **Clipboard-write feedback** in SessionsView copy actions.
8. macOS **Reopen** (dock click after last window closed) — the app currently exits with the last window, which sidesteps it.

---
**Update:** rebased onto `desktop` @ c0cab94 (PRs #376–#378 merged meanwhile and conflicted, which was silently blocking all CI checks on this PR). Two later commits address macOS-run feedback: direct AppKit traffic-light positioning (config `trafficLightPosition` is applied in a `drawRect:` the webview never lets fire), and an app-wide type-scale raise to a 14px base with control metrics scaled to match. The hairline follow-up was fixed upstream by #376 and is dropped from the list.


vibexp-agent:
Second commit (b1b0139) replaces the config-based traffic-light positioning with direct AppKit positioning: `trafficLightPosition` is applied by tao inside the content view's `drawRect:`, which the WKWebView covers — so the value stored but the buttons stayed at the default corner position, misaligned with the titlebar strip (reported from a real macOS run). The shell now moves the `standardWindowButton`s itself (transcribed from tao's own `inset_traffic_lights`), centers them in the `--titlebar-h` band so they share a centerline with the strip's icons by construction, insets the close button to x=16, and re-applies on resize/fullscreen/theme/focus events — the same reset events `tauri-plugin-decorum` re-applies on.

PR #394, opened 2026-08-20
fix(desktop): scope the opener ACL, or every external link is silently dead

shahariaazam:
## The bug

Creating a Google integration and clicking **Authorise** shows "Waiting for you to finish in the browser…" — but no browser opens, and **Reopen link** does nothing either.

## Root cause

`capabilities/default.json` granted `opener:allow-open-url`. The plugin's own manifest describes that permission as:

> Enables the open_url command **without any pre-configured scope**.

The scope entries live in a *separate* permission, `opener:allow-default-urls` (`mailto:*`, `tel:*`, `http://*`, `https://*`), which `opener:default` bundles and this capability never included. Nothing supplied a global scope either — `tauri.conf.json` has only an `updater` block.

`open_url` checks that scope before doing anything:

```rust
// tauri-plugin-opener-2.5.4/src/commands.rs:36
if scope.is_url_allowed(&url, with.as_deref()) { app.opener().open_url(url, with) }
else { Err(Error::ForbiddenUrl { url, with }) }
```

and `is_url_allowed` ends in `self.allowed.iter().any(...)` — an **empty allow list returns false for every URL**.

## Why nobody noticed for the app's whole history

`git log -p --follow` shows `opener:allow-open-url` has been the entry since the first desktop commit (`c23e816`) — never scoped. It stayed invisible because `openExternal` (`desktop/src/lib/tauri.ts:53`) catches the rejection, logs a `console.warn`, and falls back to `window.open`, which is inert in WebKitGTK (no new-window handler is registered). The wrapper then **returns normally**, so the OAuth panel advances as if a browser had opened.

So this was never Google- or OAuth-specific. Every external link was dead:

| Call site | Link |
|---|---|
| `IntegrationsView.tsx:854`, `:934` | OAuth Authorise / Reopen link |
| `SessionsView.tsx:1008` | PR links |
| `chat/Markdown.tsx:47` | every link in chat output |
| `SettingsView.tsx:1690` | release URL |
| `App.tsx:274,277`, `AboutView.tsx:72` | repo / readme |

## The fix

One line: grant `opener:allow-default-urls` beside the command.

Chosen over `opener:default` because it is the same four schemes without also granting `reveal-item-in-dir`, which nothing calls. Keeping it to those schemes matters here specifically because `Markdown.tsx` routes *model-generated* links through `openExternal`.

## Verification

Reproduced and verified in the **real Tauri webview** (`npm run app:alongside` + WebKit remote inspector), not in Chrome:

| | before | after |
|---|---|---|
| `invoke("plugin:opener\|open_url")` | `REJECTED: Not allowed to open url …` | `RESOLVED` |
| `openExternal(...)` | returned normally, `console.warn` only, nothing opened | returned normally, no warn, opened |
| OS launcher (`xdg-open` PATH shim) | never invoked | `OPENED https://accounts.google.com/o/oauth2/auth?...` |

The real **Authorise** button was then clicked in the running window and handed the OS the actual Google consent URL; **Reopen link** likewise. Finally confirmed unshimmed (a real browser tab opened) and by a **real end-to-end Google authorization**, which succeeded.

## Tests

`desktop/src-tauri/tests/capabilities.rs` asserts that granting `opener:allow-open-url` also grants a URL scope, and — same file, same silent-failure class — that `remote.urls` still covers the loopback release origin (the #385 case). Both are states in which the capability is valid JSON *and* a valid ACL, so nothing but an assertion catches them. Confirmed the first test fails when the fix is reverted.

## Unrelated drive-by

`tests/parity_session_detail.rs` had a `nonminimal_bool` clippy failure (`!…is_none()`) that breaks `cargo clippy --all-targets -- -D warnings` on current stable — already failing CI on this branch. Fixed so CI can go green.

## Skill update

`.claude/skills/local-verify` gained the gaps this hunt exposed: a pre-flight script (stale `:1420` and stale `node_modules` both fail with errors that don't name their cause), a reusable zero-dependency inspector driver (`drive.mjs` — the documented `ws`-via-playwright-core recipe pointed at an uninstalled tree, and `Runtime.evaluate`'s `awaitPromise` silently returns `{}` in WebKit), plus new hops for Tauri IPC/ACL probing and OS-handoff verification.

### DIFF
--- a/.claude/skills/local-verify/SKILL.md
+++ b/.claude/skills/local-verify/SKILL.md
@@ -19,10 +19,38 @@ hops, bisect them instead of guessing.
 | Browser engine | Chrome via Vite dev server | frontend logic, CSS, UI flows |
 | WebKitGTK engine | Python `gi` WebKit2 probe | engine-level behavior without the app |
 | **Real Tauri webview** | WebKit remote inspector | the only place webview-specific bugs exist |
+| **Tauri IPC / ACL** | `__TAURI_INTERNALS__.invoke` in that webview | whether a command is *permitted*, separate from whether the UI calls it |
+| **OS handoff** | PATH shim over the launcher | whether the OS was actually asked to do the thing |
 
 A bug reported from the desktop app is not verified fixed until the fix is
 observed **in the real webview** (or the root cause is proven to live in a hop
 that Chrome shares).
+
+**Bisect down the table, not up.** A dead button has at least four candidate
+hops (handler never fired → command denied by the ACL → OS never asked →
+OS asked and did nothing). Probing `invoke` directly settles the middle two in
+one call and costs seconds; clicking through the UI settles none of them,
+because every one of those failures looks identical from the UI.
+
+## Pre-flight — run this before the first launch, always
+
+`npm run app:alongside` has two failure modes that report themselves as
+something else, and both cost a full launch cycle to diagnose:
+
+- **A stale `:1420`.** A previous Vite survives `pkill -f vite` often enough
+  that the port is the only reliable test. `tauri dev` then dies with
+  `The beforeDevCommand terminated with a non-zero status code.` — the real
+  line (`Port 1420 is already in use`) is ~10 lines above the tail you read.
+- **A stale `desktop/node_modules`.** Branches add frontend deps; the tree on
+  disk is from whatever branch you last built. Vite fails *after* announcing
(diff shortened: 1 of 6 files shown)


shahariaazam:
Superseded by a PR rebased onto the post-#392 root layout — this branch was cut before `desktop/` moved to the repository root, so it conflicts wholesale.

PR #399, opened 2026-08-22
feat: sunset the Go/web build and migrate users to Agento Desktop

### LINKED ISSUE #389: v0.11.2: sunset the web version (cutoff 1 Sept 2026) and migrate users to Agento Desktop
## Goal

Ship a final patch release on `main` (`v0.11.2`) that announces the sunset of the Go/web version of Agento and migrates users to the Tauri desktop app.

**Hard cutoff: 1 September 2026.** That date is the end of *support*, not the end of *use*. The web version keeps running indefinitely after it — see "Decided: no notice wall" below. What stops on that date is the self-updater.

Related: the repo restructure is tracked in #388. This issue is on `main` and is independent of it.

## Background: what actually happens when the desktop tags lose their prefix

Desktop releases are tagged `desktop-v*` today and will eventually become plain `v*` in this same repo. Checked `go-selfupdate` v1.6.0's matcher against the real releases before designing around it:

- `Updater.getSuffixes()` builds every candidate as `<os><sep><arch><ext>` — **the OS token is mandatory**.
- No Tauri asset carries one: `Agento_aarch64.app.tar.gz`, `Agento_0.1.1_x64-setup.exe`, `Agento_0.1.1_amd64.deb`, `Agento_0.1.1_x64.dmg`.
- `findReleaseAndAssetForArch` iterates *all* releases and keeps only those with a matching asset, so a desktop-tagged release is skipped and the picker falls back to the newest release that does have `agento_Darwin_arm64.tar.gz` / `agento_Linux_x86_64.tar.gz` / `agento_Windows_x86_64.zip`.
- Verified for darwin/arm64, darwin/amd64, windows/amd64, linux/amd64, linux/arm64 — all resolve to `v0.11.1` today.

**Conclusion:** renaming the desktop tags can never cause a legacy binary to mis-install a `.dmg` or an NSIS `.exe` over the CLI. The worst case is that `agento update` reports **"already up to date" forever**, silently.

That silence is the actual risk, and it is what this release exists to prevent. It also means the notice must ship *before* the tags are renamed.

## Reaching users who never upgrade

A banner compiled into `v0.11.2` only reaches people who install `v0.11.2`. For everyone still on `≤v0.11.1`, the **only** channel we control is what their existing update check displays: the version string, the release title and the release notes, surfaced in the UI update banner and by `agento update`.

So:

- [ ] Title the release so the message survives truncation, e.g. **`v0.11.2 — Agento web sunsets 1 Sept, migrate to Agento Desktop`**
- [ ] Put the migration instructions and the desktop download link in the first three lines of the release notes
- [ ] Keep publishing nothing after this on `v0.*`, so this stays the newest thing a legacy checker can see

### A. Sunset notice

- [ ] Pre-run hook prints the sunset notice before every command. Make it a **static, offline** notice — no GitHub call — so it cannot degrade, time out, or collide with the renamed tags. Rate-limit to once per day via the existing `update-check.json` cache dir.
- [ ] `agento web` prints the full notice on startup (cutoff date, why, download link).
- [ ] Web UI: an alert visible to every user on every page, **always dismissible**, and the dismissal must persist. Never modal, never blocking, never re-armed on a timer.
- [ ] Notice must carry: the 1 Sept cutoff, the desktop releases URL, and the fact that the desktop app reads the **same `~/.agento/agento.db`** — no export, no data migration, install and go.

### B. Replace the updater rather than delete it

- [ ] `agento update` stops self-updating the Go binary. Instead it resolves the latest desktop release and prints the **direct download link for the user's GOOS/GOARCH**, plus the release page.
  - Needs its own small GitHub query — `go-selfupdate` will not find these assets, by the mechanism described above.
  - Asset → platform map: `.dmg` (macOS, `x64`/`aarch64`), `.deb`/`.rpm`/`.AppImage` (Linux, `amd64`/`arm64`/`aarch64`), `x64-setup.exe` (Windows).
- [ ] Homebrew path: publish a cask and make `HomebrewUpgradeMessage` point at it instead of `brew upgrade agento`.
- [ ] On/after 1 Sept 2026, `agento update` stops offering an update and prints only the migration message. This disables *updating*, not the app — every other command is unaffected.

### C. Leftover background service

The DB is explicitly **not** the concern — the desktop app reloads on first run either way. The concern is a leftover `agento service` unit (launchd / systemd user unit):

- it holds `:8990`;
- it runs **its own scheduler**, so every scheduled task fires twice once the desktop app is installed;
- it re-registers the Telegram webhook under whichever instance started last.

That is the documented two-scheduler hazard and it is completely invisible to the user.

- [ ] **Do not uninstall silently.** Removing a user's launchd/systemd unit unprompted is destructive and out of scope for a patch release.
- [ ] `agento update` checks `daemon.Status().Installed` and, if so, offers to run the existing `Manager.Uninstall` (already exposed as `agento service uninstall`). `-y` accepts.
- [ ] The desktop app checks for a leftover unit on **first run** and offers to remove it — that is where the user actually ends up, and where they will otherwise hit the double-fire.

### D. Docs

- [ ] `README.md` on `main`: sunset banner at the top, desktop install instructions, cutoff date
- [ ] `docs/getting-started.md`: point at the desktop app
- [ ] Pin a repo announcement / discussion with the same content

## Decided: no notice wall

After 1 Sept the web version **keeps working**, indefinitely. Staying on it is the user's call.

Explicit non-goals — none of these may be built:

- no paywall, no licence check, no kill switch, no date-gated refusal to start;
- no blocking or modal notice, no non-dismissible banner, no interstitial between the user and the app;
- no feature degradation, no throttling, and nothing that re-arms a dismissed notice.

Every notice in section A is informational and dismissible. The only thing that changes behaviour on 1 Sept is that the self-updater stops offering updates — the app itself is untouched.

Rationale: bricking or nagging installed binaries on a date only punishes the users who *did* upgrade to `v0.11.2`. Anyone still on `≤v0.11.1` would never see it anyway, so it buys nothing.

## Sequencing

`v0.11.2` must ship **before** the desktop tags are renamed to plain `v*` (#388's flip). It is 12 days to the cutoff, and the update check is cached for 24h, so adoption is slow by design — the sooner this ships the more people see it.

---
## WHAT

**Summary:** Ship `v0.11.2` on `main` as the terminal Go/web release — a static, dismissible sunset notice across CLI/web/UI, an `agento update` that hands users the desktop installer instead of self-updating, and an offer to remove the leftover `agento service` unit.

**Scope:**
- **A. Sunset notice** — offline/static (no GitHub call), once-per-day rate limited, carrying the 1 Sept 2026 cutoff, the desktop releases URL, and the fact that the desktop app reads the **same `~/.agento/agento.db`** (no export, no migration).
  - Pre-run hook (`cmd/root.go`) prints it before **every** command except `completion`/`__complete`, TTY or not, on **stderr**; honours `AGENTO_SKIP_UPDATE_CHECK`.
  - `agento web` prints the full notice in its startup banner.
  - Web UI: a dismissible alert on every page, dismissal persisted in `localStorage`.
- **B. `agento update` becomes a migration pointer** — resolves the latest desktop release and prints the direct download link for the user's GOOS/GOARCH plus the release page. Never self-updates the Go binary again. On/after 1 Sept 2026 it prints only the migration message.
- **C. Leftover service unit** — `agento update` checks `daemon.Status().Installed` and *offers* (never performs unprompted) `Manager.Uninstall`; `-y` accepts.
- **D. Docs + release framing** — README/`docs/getting-started.md` banners, a pinned announcement, and a release **title** and first three lines of notes that carry the migration message (the only channel that reaches users on `≤v0.11.1`).
- **E. Homebrew** — publish a cask in `shaharia-lab/homebrew-tap` and repoint `HomebrewUpgradeMessage` at it. Cross-repo; see the ordering constraint in HOW.

**Affected Components:**
- `cmd/root.go` — `NewRootCmd` `PersistentPreRunE`, new sunset-notice call ahead of `runAutoUpdateCheck`; new gating (not `shouldRunAutoCheck`'s)
- `cmd/update.go` — `runUpdate` rewritten; `maybeRestartService` becomes dead for the update path; keep the `newServiceManager` seam for the uninstall offer
- `internal/updater/` — new desktop-release resolver + platform→asset map; `HomebrewUpgradeMessage` (`installer.go:94`) repointed; `Install`/`go-selfupdate` path retired from the update command
- new `internal/sunset/` — the cutoff date, the notice text, and the once-per-day stamp file (single source of truth for every Go caller)
- `cmd/web.go` — `printBanner`/`printPlainBanner` (~L560–611) gain the notice
- `frontend/src/components/SunsetBanner.tsx` (new), mounted in `Layout.tsx:34` beside `UpdateBanner`; `frontend/src/lib/sunset.ts` mirrors the cutoff
- `README.md`, `docs/getting-started.md`
- `shaharia-lab/homebrew-tap` → `Casks/agento.rb` (does not exist today; the tap has only `Formula/`)

**Acceptance Criteria:**
- [ ] `agento ask …` and `agento service status` each print the sunset notice once, then not again within 24h; a second invocation in the same day prints nothing. `AGENTO_SKIP_UPDATE_CHECK=1` suppresses it. `agento __complete` output is byte-identical to today's.
- [ ] `agento update` on a `v0.11.2` binary prints a resolvable direct download URL for the running GOOS/GOARCH (asserted per platform against the desktop asset names, table-driven), never calls `updater.Install`, and exits 0.
- [ ] With the clock faked to ≥ 2026-09-01, `agento update` prints only the migration message and offers no download; **every other command still runs unchanged** (`agento web`, `agento ask` unaffected).
- [ ] With a stubbed `Manager` reporting `Installed: true`, `agento update` prompts before uninstalling and does nothing on a declined/EOF answer; `-y` calls `Uninstall` exactly once. With `Installed: false` no prompt appears.
- [ ] The web UI banner is dismissible, stays dismissed across a reload and across navigation, is never modal, and is never re-armed by a timer (Vitest on `SunsetBanner`).
- [ ] The `v0.11.2` release title contains the cutoff date and the word *Desktop*, and the first three lines of the notes carry the desktop download link.

**Out of Scope:**
- **The desktop app's own first-run check for a leftover service unit.** `main` has no `desktop/` tree — it lives only on the `desktop` branch. Needs its own issue on that branch; the CLI half here is what ships in `v0.11.2`.
- Everything in *Decided: no notice wall* below: no paywall, licence check, kill switch, date-gated refusal to start, modal/blocking/non-dismissible notice, feature degradation, throttling, or re-arming a dismissed notice.
- Silent removal of a user's launchd/systemd unit.
- Any DB export or migration tooling — the desktop app reads `~/.agento/agento.db` directly.
- Renaming the `desktop-v*` tags. That is a separate, unfiled decision (see Context).

## WHY

**Business Value:** `v0.11.2` is the last message this project can send to everyone running the Go/web build. Once the desktop tags lose their prefix, the legacy updater goes quiet rather than wrong — and a silent updater is indistinguishable from a maintained project, so users sit on an unsupported binary indefinitely without ever being told.

**Problem Statement:** `go-selfupdate` v1.6.0 requires an `<os><sep><arch><ext>` asset token that **no** Tauri asset carries (verified against darwin/arm64, darwin/amd64, windows/amd64, linux/amd64, linux/arm64 — all resolve to `v0.11.1` today). So a desktop-tagged release is skipped entirely and `agento update` reports *"already up to date" forever*. The failure mode is silence, not a bad install. Meanwhile a leftover `agento service` unit holds `:8990`, runs **its own scheduler** — every scheduled task fires twice once the desktop app is installed — and re-registers the Telegram webhook under whichever instance started last, all invisibly.

**Expected Impact:** Every user who reaches `v0.11.2` gets an unambiguous, non-hostile migration path and a one-command escape from the two-scheduler hazard. Users on `≤v0.11.1` get the message through the one channel we still control — the release title and notes their existing 24h-cached check renders in the UI banner and in `agento update`. Adoption is slow by design (24h cache × 9 days to the cutoff), so ship time is the dominant variable.

**Context:**
- `Related: #388` — the repo restructure, on the `desktop` branch. **Now CLOSED.** This issue is on `main` and independent of it.
- **Partial staleness corrected while refining:**
  - The body attributes the tag rename to *"#388's flip"*. #388 explicitly puts it **out of scope** ("Do not rename the release tags… that's a flip-day migration with its own plan"). There is **no issue tracking the rename today** — the sequencing constraint is real, but it has no tracked counterpart to be blocked by. File one, or the ordering lives only here.
  - "12 days to the cutoff" was written on ~2026-08-20. As of 2026-08-23 it is **9 days**.
  - The section-B asset map flattens arch tokens that actually differ per package format — corrected in HOW against the real `desktop-v0.1.1` assets.
- **Decisions taken during refinement** (all three the recommended option):
  1. The desktop-side first-run unit check is **split out** to a `desktop`-branch issue; `v0.11.2` stays a shippable `main`-only patch.
  2. The pre-run notice uses **its own gating**, not `shouldRunAutoCheck`'s. Reusing that would skip exactly the users this release must reach — non-TTY installs, `agento service`, and `agento update` itself.
  3. The Homebrew cask and the pinned announcement stay **tracked deliverables here**, marked as landing outside this repo, so `HomebrewUpgradeMessage` is not repointed at a cask that does not exist yet.

## HOW

**Technical Approach:** Add a small `internal/sunset` package owning the cutoff date, the notice text and a once-per-day stamp — the single definition every Go caller reads, mirroring how `internal/config` owns values the layers above must agree on. Gut `runUpdate`'s self-update half and replace it with a desktop-release resolver anchored on the **fixed `desktop-latest` manifest tag**, which is rename-proof. Reuse the existing dismissible-banner pattern verbatim on the frontend. Nothing here touches the scanner, storage, or any request path.

**Implementation Details:**

- **`internal/sunset/sunset.go` (new)**
  - `const CutoffDate = "2026-09-01"` + `var Cutoff = time.Date(2026, 9, 1, 0, 0, 0, 0, time.UTC)`; `Passed(now time.Time) bool`.
  - `Notice()` (short, one-line, for the pre-run hook) and `FullNotice()` (multi-line, for `agento web`). Both static strings — **no network call**, so they cannot time out, degrade, or collide with a renamed tag.
  - `ShouldPrint(dataDir string, now time.Time) bool` + `Stamp(dataDir, now)`: a **separate** file, `sunset-notice.json`, in `cfg.DataDir`. Do **not** reuse `update-check.json` — `updater.CheckResult` is keyed on `CurrentVersion` *and* `Platform` and invalidates on either, which would re-fire the notice on unrelated conditions. Same directory, different file. Every failure path (unreadable dir, corrupt JSON) degrades to *printing*, not to silence.

- **`cmd/root.go`**
  - In `PersistentPreRunE`, call `maybePrintSunsetNotice(cmd, cfg)` **before** `runAutoUpdateCheck`. Write to `os.Stderr` so stdout of machine-readable subcommands stays clean.
  - New `shouldPrintSunset(cmd) bool` — deliberately **not** `shouldRunAutoCheck`. Skip only: `completion` and `__complete` (cobra parses that stdout; a stray line corrupts shell completion), plus `AGENTO_SKIP_UPDATE_CHECK`. **No TTY requirement**, no `update`/`service` skip, and dev builds are skipped as they are today.

- **`cmd/update.go` — `runUpdate` rewritten**
  - Keep the dev-build guard. Delete the `checker.Check` → `updater.Install` path; `updater.Install`, `ErrHomebrewManaged` and `maybeRestartService` fall out of this command (leave the package functions in place — the API handler behind `/version/update-check` still uses the checker).
  - `sunset.Passed(time.Now())` → print the migration message and return. This disables **updating**, not the app.
  - Otherwise `updater.ResolveDesktopRelease(ctx)` (below) → print the platform-specific direct link + the release page.
  - Then the service check: `mgr, err := newServiceManager(cfg)` (existing seam), `mgr.Status(ctx)`; if `Installed`, explain the double-scheduler hazard and prompt. `-y` accepts; a declined or EOF answer is a **no-op**. Read the answer with `bufio.Reader` (`fmt.Scanln` leaves bytes in stdin — the reason `promptAndMaybeUpdate` already uses one).

- **`internal/updater/desktop.go` (new) — release resolution**
  - Anchor: `GET https://github.com/shaharia-lab/agento/releases/download/desktop-latest/latest.json`. That tag is **fixed** and re-pushed by `desktop-release.yml` on every publish, so it survives `desktop-v*` → `v*`. Take `version` and derive the release tag from any `platforms[*].url`.
  - Three traps a naive query hits, all live today: `/releases/latest` returns **`v0.11.1`** (the Go release); `desktop-latest` is a **pre-release**; `desktop-v0.1.2` is a **draft**. Any "newest release" heuristic resolves wrong. If the manifest fetch fails, fall back to listing releases and matching by **asset shape**, not tag name — never by `desktop-` prefix, which disappears at the flip.
  - `latest.json` only lists *auto-updater* artifacts (`.app.tar.gz`, `.AppImage`, `.exe`). Build the **installer** URL from the resolved tag + version. Verified names on `desktop-v0.1.1` — note the arch token differs per format, which the issue body's map flattened:

    | GOOS/GOARCH | asset |
    |---|---|
    | darwin/amd64 | `Agento_<v>_x64.dmg` |
    | darwin/arm64 | `Agento_<v>_aarch64.dmg` |
    | linux/amd64 | `Agento_<v>_amd64.deb` · `Agento-<v>-1.x86_64.rpm` · `Agento_<v>_amd64.AppImage` |
    | linux/arm64 | `Agento_<v>_arm64.deb` · `Agento-<v>-1.aarch64.rpm` · `Agento_<v>_aarch64.AppImage` |
    | windows/amd64 | `Agento_<v>_x64-setup.exe` |

  - Unknown GOOS/GOARCH → print the release page only, never a guessed URL.

- **`internal/updater/installer.go:94`** — `HomebrewUpgradeMessage` → `brew install --cask shaharia-lab/tap/agento`. **Ordering constraint: the cask must exist first.** The tap has `Formula/agento.rb` and **no `Casks/` directory at all**; add `Casks/agento.rb` pointing at the `.dmg` assets and merge it *before* this string changes, or the message tells users to run a command that fails.

- **`cmd/web.go`** — emit `sunset.FullNotice()` after the version/URL table in both `printBanner` and `printPlainBanner` (~L596–611). Unconditional on startup; the once-a-day stamp governs the pre-run hook only.

- **Frontend**
  - `frontend/src/lib/sunset.ts` — `SUNSET_CUTOFF = '2026-09-01'`, the desktop releases URL, and the `~/.agento/agento.db` line. Comment it as the mirror of `internal/sunset`.
  - `frontend/src/components/SunsetBanner.tsx` — copy `UpdateBanner`'s shape: `localStorage` key `agento-sunset-dismissed`, `try/catch` around every access (private browsing), state initialised from storage on mount. Unlike `UpdateBanner` it takes **no version key** — dismissal is permanent, and there is no `setInterval` re-check.
  - Mount in `Layout.tsx` beside `UpdateBanner` (`:34`) so it renders above every route.

**Testing Strategy:**
- `internal/sunset/sunset_test.go` — `Passed` on both sides of the boundary with an injected clock; `ShouldPrint`/`Stamp` round-trip over a `t.TempDir()`; corrupt and unreadable stamp files both print.
- `internal/updater/desktop_test.go` — table-driven GOOS/GOARCH → asset name over the verified matrix; a stubbed `latest.json` server; manifest-fetch failure exercises the asset-shape fallback; unknown platform yields the release page only.
- `cmd/update_test.go` (extend) — pre-cutoff vs post-cutoff output with a faked clock; `Installed: true/false` × `-y`/declined/EOF against a stub `Manager` through `newServiceManager`, asserting `Uninstall` call counts.
- `cmd/root_test.go` (extend) — notice prints for a normal subcommand, is suppressed by `AGENTO_SKIP_UPDATE_CHECK`, is suppressed for `__complete`, prints without a TTY, and does not re-print inside 24h.
- `frontend/src/components/SunsetBanner.test.tsx` — renders, dismisses, stays dismissed on remount, has no `role="dialog"`, and no timer re-arms it.

- `README.md` — sunset banner above the badges; swap the `brew install shaharia-lab/tap/agento` quick-start for the desktop cask; keep the Go install block under a "legacy, unsupported after 1 Sept 2026" heading.
- `docs/getting-started.md` — point at the desktop app first.
- Pin a repo announcement/discussion with the same content.
- Release framing (**not code, but a hard deliverable**): title `v0.11.2 — Agento web sunsets 1 Sept, migrate to Agento Desktop`; migration instructions + desktop link in the first three lines of the notes; publish nothing further on `v0.*` so this stays the newest thing a legacy checker sees.
- No new deps, no new env vars. `AGENTO_SKIP_UPDATE_CHECK` gains a second meaning — document it.

**Files to Create:** `internal/sunset/sunset.go`, `internal/sunset/sunset_test.go`, `internal/updater/desktop.go`, `internal/updater/desktop_test.go`, `frontend/src/lib/sunset.ts`, `frontend/src/components/SunsetBanner.tsx`, `frontend/src/components/SunsetBanner.test.tsx`; `Casks/agento.rb` in `shaharia-lab/homebrew-tap`.

**Files to Modify:** `cmd/root.go`, `cmd/root_test.go`, `cmd/update.go`, `cmd/update_test.go`, `cmd/web.go`, `internal/updater/installer.go`, `frontend/src/components/Layout.tsx`, `README.md`, `docs/getting-started.md`.

**Risks:**
- *The cutoff date lives in two languages.* → One Go const, one TS const, each commenting the other; it is a fixed literal that never changes after release, and the Vitest/Go tests both assert it.
- *Repointing `HomebrewUpgradeMessage` before the cask merges* ships a command that fails. → Cask PR merges first; the acceptance criteria treat them as one deliverable.
- *A `latest.json` fetch failure could leave `agento update` with nothing to say.* → Asset-shape fallback, then the release page; never an error exit, never a guessed URL.
- *The notice becomes nagging.* → Once per 24h, stderr only, permanently dismissible in the UI, and the *Decided: no notice wall* non-goals are enforced in review.
- *Shipping after the tag rename inverts the whole point.* → `v0.11.2` must be published **before** the desktop tags become plain `v*`.

**Complexity:** Medium (multi-file across the Go CLI, the web startup path, the React frontend and docs, plus one cross-repo cask; well under 500 LOC, every touched surface has an existing pattern to copy)

---
**Refined by:** AI Issue Refiner · **Date:** 2026-08-23 · **Status:** Ready

vibexp-agent:
Closes #389

## WHAT

The final release of the Go/web build: a static, dismissible sunset notice across CLI/web/UI, an `agento update` that hands users the desktop installer instead of self-updating, and an offer to remove the leftover `agento service` unit.

- **Sunset notice** — offline/static, once-per-day rate limited, carrying the **1 September 2026** cutoff, the desktop releases URL, and the fact that the desktop app reads the **same `~/.agento/agento.db`**.
- **`agento update`** — resolves the current desktop release and prints the direct installer link for the running GOOS/GOARCH. Never self-updates again. Past the cutoff it prints only the migration message.
- **Leftover service unit** — `agento update` checks `daemon.Status().Installed` and *offers* `Manager.Uninstall`; `-y` accepts.
- **Docs** — README + `docs/getting-started.md` lead with the desktop app; the Go build moves under a "legacy, unsupported after 1 Sept 2026" heading.

## WHY

`go-selfupdate` v1.6.0 builds every candidate asset as `<os><sep><arch><ext>`, and **no Tauri asset carries the OS token**. So once the desktop tags lose their `desktop-` prefix, a legacy binary does not mis-install a `.dmg` — it reports **"already up to date" forever**. The failure mode is silence, and a silent updater is indistinguishable from a maintained project.

A leftover `agento service` unit meanwhile holds `:8990`, runs **its own scheduler** — every scheduled task fires twice once the desktop app is installed — and re-registers the Telegram webhook under whichever instance started last. All invisible.

## HOW — three decisions worth reviewing

**1. Release resolution anchors on the fixed `desktop-latest` manifest tag.** All three obvious alternatives are wrong *today*:

| Approach | What it resolves to now |
|---|---|
| `GET /releases/latest` | `v0.11.1`, the Go build — **by design**: desktop publishes with `--latest=false` precisely so the CLI's go-selfupdate keeps resolving it |
| newest release | `desktop-latest`, which is a **prerelease** |
| newest `desktop-v*` tag | `desktop-v0.1.2`, a **draft** whose assets aren't publicly downloadable |

The release tag is read back **out of** the manifest rather than reconstructed from the version, so the eventual `desktop-v*` → `v*` rename changes nothing. `TestResolveDesktopReleaseSurvivesTagRename` pins that.

**2. The pre-run notice uses its own gating, not `shouldRunAutoCheck`'s.** That helper requires an interactive TTY on both stdin and stdout and skips `update` and `service` — right for a prompt expecting an answer, wrong for a one-way notice, because it would exclude exactly the installs this release must reach: scripted, cron-driven, and service-only. The notice skips only `completion`/`__complete` (cobra parses that stdout) and `web`/`update`, which print the **full** notice themselves — skipping a duplicate line, not a user.

**3. The rate-limit stamp is its own file, not `update-check.json`.** That cache holds an `updater.CheckResult` keyed on both the running version *and* the GOOS/GOARCH and invalidates when either changes — sound for an update check, but it would re-fire the notice on conditions that say nothing about whether the user has read it. Same directory, separate file. Every failure path (unreadable dir, corrupt stamp, clock moved backwards) degrades to **printing**.

## Deliberate non-goals

Per the issue's *Decided: no notice wall*: no paywall, licence check, kill switch, or date-gated refusal to start; no blocking/modal/non-dismissible notice; no feature degradation, no throttling, nothing that re-arms a dismissed notice. **Only updating stops on 1 Sept** — the app keeps working indefinitely, and the notices say so.

## Acceptance criteria

- [x] `agento service status` prints the notice once, then not again within 24h; `AGENTO_SKIP_UPDATE_CHECK=1` suppresses it; `__complete` output is untouched
- [x] `agento update` prints a resolvable download URL for the running platform, never calls `updater.Install`, exits 0 — verified live against the real manifest (resolved `desktop-v0.1.1` → `Agento_0.1.1_amd64.deb`)
- [x] With the clock faked past the cutoff, `agento update` offers no download; every other command is unaffected
- [x] Stubbed `Manager` with `Installed: true` prompts before uninstalling and does nothing on decline/EOF; `-y` calls `Uninstall` exactly once; `Installed: false` prompts nothing
- [x] The web banner is dismissible, stays dismissed across reload and navigation, is never modal, is never re-armed by a timer
- [ ] Release title + first three lines of notes carry the migration message — **manual, at tag time**

## Out of scope

- **The desktop app's own first-run check for a leftover service unit.** `main` has no `desktop/` tree; it needs its own issue on the `desktop` branch.
- **The Homebrew cask** ships as a companion PR against `shaharia-lab/homebrew-tap` — see the cross-link below. `HomebrewUpgradeMessage` in this PR already points at `brew install --cask shaharia-lab/tap/agento`, so **the cask PR must merge first**.
- The pinned repo announcement (GitHub UI, no code).

## Testing

New: `internal/sunset/sunset_test.go`, `internal/updater/desktop_test.go`, `cmd/sunset_notice_test.go`, `frontend/src/components/SunsetBanner.test.tsx`. Rewritten: `cmd/update_test.go` (the old suite covered the restart path, which no longer exists).

Local gates, all green: `go build` · `go vet` · `gofmt` · `golangci-lint v2.5.0` (the version CI pins — **0 issues**) · full `go test ./...` · frontend `typecheck` · `lint` · `format:check` · 207 tests · `build`.

> Note for maintainers: `golangci-lint` 2.12.2 (a local drift from CI's pinned v2.5.0) reports 13 pre-existing issues across `internal/pricing`, `internal/api` and `internal/claudesessions` that are unrelated to this PR. Verified against the pinned version instead.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,29 @@
 # Agento
 
+> ## 🌇 Agento (web) is being retired — move to Agento Desktop
+>
+> **This is the final release of the Go/web build.** Support ends **1 September 2026**;
+> after that date `agento update` stops offering updates. The app itself keeps
+> working indefinitely — nothing shuts off, nothing expires, and staying on it is
+> your call.
+>
+> **[Agento Desktop](https://github.com/shaharia-lab/agento/releases)** is where
+> development continues. It reads the **same `~/.agento/agento.db`**, so there is
+> no export and no migration: install it and your history, agents and analytics
+> are already there.
+>
+> ```bash
+> brew install --cask shaharia-lab/tap/agento     # macOS
+> ```
+>
+> Or grab the `.dmg`, `.deb`, `.rpm`, `.AppImage` or `.exe` from
+> [Releases](https://github.com/shaharia-lab/agento/releases).
+>
+> **One thing to do after installing:** if you ran `agento service install`, remove
+> the old background service with `agento service uninstall`. Left in place it runs
+> a second scheduler, so every scheduled task fires twice. `agento update` offers to
+> do this for you.
+
 [![Release](https://img.shields.io/github/v/release/shaharia-lab/agento)](https://github.com/shaharia-lab/agento/releases)
 [![Stars](https://img.shields.io/github/stars/shaharia-lab/agento?style=flat)](https://github.com/shaharia-lab/agento/stargazers)
 [![License](https://img.shields.io/github/license/shaharia-lab/agento)](https://github.com/shaharia-lab/agento/blob/main/LICENSE)
@@ -16,11 +40,10 @@ Claude Code forgets everything the moment it exits. Agento reads the session fil
 ![Agento Insights dashboard showing cost cards and productivity metrics](docs/images/insights.png)
 
 ```bash
-brew install shaharia-lab/tap/agento
-agento web
+brew install --cask shaharia-lab/tap/agento
 ```
 
-That is the whole setup. Agento opens at `http://localhost:8990`, finds your Claude Code history, and starts building your dashboards.
+That is the whole setup. Agento finds your Claude Code history and starts building your dashboards.
 
 > ### ⭐ Star this repository
 >
@@ -35,28 +58,48 @@ That is the whole setup. Agento opens at `http://localhost:8990`, finds your Cla
 <table>
 <tr><td width="50%">
 
-**Homebrew**
+**Homebrew (macOS)**
 
 ```bash
-brew install shaharia-lab/tap/agento
-agento web
+brew install --cask shaharia-lab/tap/agento
 ```
 
 </td><td width="50%">
 
 **Direct download**
 
+Grab the installer for your platform from
+[Releases](https://github.com/shaharia-lab/agento/releases):
+`.dmg` (macOS), `.deb` / `.rpm` / `.AppImage` (Linux),
+`-setup.exe` (Windows).
+
+</td></tr>
+</table>
+
+Installers for Linux (x86_64, arm64), macOS (Intel, Apple Silicon) and Windows are on the [Releases page](https://github.com/shaharia-lab/agento/releases).
+
+<details>
+<summary><b>Legacy: the Go/web build (unsupported after 1 September 2026)</b></summary>
+
+The single Go binary that serves the browser UI on `localhost:8990`. It still
+installs and still runs, but it receives no further releases — see the notice at
+the top of this file.
+
 ```bash
-# grab the archive for your platform from Releases
+brew install shaharia-lab/tap/agento     # the formula, not the cask
+agento web
+```
+
+```bash
+# or grab the archive for your platform from Releases
 tar -xzf agento_Linux_x86_64.tar.gz
 sudo mv agento /usr/local/bin/
 agento web
 ```
 
-</td></tr>
-</table>
+Everything below this point describes both builds unless it says otherwise.
 
-Binaries for Linux (x86_64, arm64), macOS (Intel, Apple Silicon) and Windows are on the [Releases page](https://github.com/shaharia-lab/agento/releases).
+</details>
 
 Useful flags: `agento web --port 3000` to change the port, `--no-browser` to skip opening a tab. To keep it running in the background across reboots:
 
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -7,12 +7,14 @@ import (
 	"fmt"
 	"os"
 	"strings"
+	"time"
 
 	"github.com/mattn/go-isatty"
 	"github.com/spf13/cobra"
 
 	"github.com/shaharia-lab/agento/internal/build"
 	"github.com/shaharia-lab/agento/internal/config"
+	"github.com/shaharia-lab/agento/internal/sunset"
 	"github.com/shaharia-lab/agento/internal/updater"
 )
 
@@ -46,6 +48,7 @@ func NewRootCmd(cfg *config.AppConfig) *cobra.Command {
 		Long:    "A platform for running Claude agents defined in YAML configuration files.",
 		Version: build.String(),
 		PersistentPreRunE: func(cmd *cobra.Command, _ []string) error {
+			maybePrintSunsetNotice(cmd, cfg)
 			runAutoUpdateCheck(cmd, cfg)
 			return nil
 		},
@@ -73,6 +76,78 @@ func Execute() {
 		fmt.Fprintln(os.Stderr, err)
 		os.Exit(1)
 	}
+}
+
+// sunsetSkipCommands lists subcommands that must never print the one-line
+// sunset notice, and why each one is on the list.
+//
+// The list is deliberately far shorter than updateCheckSkipCommands, and it
+// skips for two different reasons:
+//
+//   - `completion` and `__complete`: cobra's shell integration parses their
+//     stdout, and although the notice goes to stderr, a shell that merges the
+//     two streams would ingest it as a completion candidate.
+//   - `web` and `update`: both print the FULL notice themselves, so the
+//     one-liner would be an immediate duplicate of a longer message the user is
+//     about to read anyway. This skips a redundant line, not a user.
+//
+// Everything else — `service`, `help`, a piped `ask`, a cron-driven run — still
+// prints, because none of those has a notice of its own and they are exactly
+// the installs a TTY-gated notice would miss.
+var sunsetSkipCommands = map[string]struct{}{ //nolint:gochecknoglobals
+	"completion": {},
+	"__complete": {},
+	"web":        {},
+	"update":     {},
+}
+
+// sunsetNow is a seam over time.Now so tests can drive the rate limit.
+var sunsetNow = time.Now //nolint:gochecknoglobals
+
+// maybePrintSunsetNotice prints the retirement notice ahead of the user's
+// command, at most once a day.
+//
+// It deliberately does NOT reuse shouldRunAutoCheck's gating. That helper
+// requires an interactive TTY on both stdin and stdout and skips `update` and
+// `service` — reasonable for a prompt that expects an answer, but wrong for a
+// one-way notice, because it would exclude exactly the people who most need to
+// see it: anyone running agento from a script, from a cron job, or under
+// `agento service`.
+//
+// The notice is static and offline. It makes no network call, so it cannot slow
+// a command down, cannot time out, and cannot break when the desktop releases
+// stop carrying their tag prefix.
+func maybePrintSunsetNotice(cmd *cobra.Command, cfg *config.AppConfig) {
+	if !shouldPrintSunsetNotice(cmd) {
+		return
+	}
+	now := sunsetNow()
+	if !sunset.ShouldPrint(cfg.DataDir, now) {
+		return
+	}
+	fmt.Fprintf(os.Stderr, "\n%s\n\n", sunset.Notice())
+	// Best-effort: a failed stamp only means the notice shows again sooner.
+	_ = sunset.Stamp(cfg.DataDir, now) //nolint:errcheck
+}
+
+// shouldPrintSunsetNotice applies the notice's own skip rules.
+func shouldPrintSunsetNotice(cmd *cobra.Command) bool {
+	// A dev build is not an install anyone needs to migrate.
+	if build.IsDevBuild(build.Version) {
+		return false
+	}
+	// Honor the existing opt-out. It already means "do not talk to me about
+	// versions", which is the same wish.
+	if v := os.Getenv(skipUpdateCheckEnv); v == "1" || strings.EqualFold(v, "true") {
+		return false
+	}
+	if cmd == nil || !cmd.Runnable() {
+		return false
+	}
+	if _, skip := sunsetSkipCommands[cmd.Name()]; skip {
+		return false
+	}
+	return true
 }
 
 // runAutoUpdateCheck is the PersistentPreRunE body. It performs a cached
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -1,9 +1,6 @@
 package cmd
 
 import (
-	"bufio"
-	"context"
-	"errors"
 	"fmt"
 	"os"
 	"strings"
@@ -151,7 +148,7 @@ func shouldPrintSunsetNotice(cmd *cobra.Command) bool {
 }
 
 // runAutoUpdateCheck is the PersistentPreRunE body. It performs a cached
-// update check and, on a fresh hit, prompts the user to update.
+// update check and, on a fresh hit, announces the newer release.
 //
 // The function is split out so it can be unit-tested without spinning up cobra.
 // All paths return without raising errors — auto-check must never fail the user's command.
@@ -171,7 +168,7 @@ func runAutoUpdateCheck(cmd *cobra.Command, cfg *config.AppConfig) {
 		return
 	}
 
-	promptAndMaybeUpdate(cmd.Context(), result)
+	announceUpdate(result)
 }
 
 // shouldRunAutoCheck applies all skip rules and returns true only when the
@@ -188,7 +185,10 @@ func shouldRunAutoCheck(cmd *cobra.Command) bool {
 		return false
 	}
 
-	// Non-interactive (CI, pipes, redirected stdin/stdout) — no point prompting.
+	// Non-interactive (CI, pipes, redirected stdin/stdout). The announcement no
+	// longer prompts, but a scripted run has already had the sunset notice —
+	// which is the message that matters now — so this stays a skip rather than
+	// becoming a second unsolicited line in every pipeline.
 	if !isInteractive() {
 		return false
 	}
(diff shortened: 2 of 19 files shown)


vibexp-agent:
Companion cask PR: **shaharia-lab/homebrew-tap#12**

Merge order matters: this PR's `HomebrewUpgradeMessage` already points at `brew install --cask shaharia-lab/tap/agento`, so the cask PR must land first or that message names a cask that does not exist.

vibexp-agent:
Cask merged: shaharia-lab/homebrew-tap#12. `brew install --cask shaharia-lab/tap/agento` now resolves, so the `HomebrewUpgradeMessage` and the docs in this PR point at something real.

**Review round 2 — all four findings applied and verified:**
1+2. `promptAndMaybeUpdate` → `announceUpdate`; `updater.Install` is now unreachable from any command, so "this build no longer self-updates" is true of the binary rather than of one subcommand.
3. `internal/sunset/mirror_test.go` reads `frontend/src/lib/sunset.ts` and compares — proven to fail on a deliberate one-sided edit, then restored.
4. `cmd/web_banner_test.go` covers the startup notice — proven to fail when the notice is removed.

Plus one CI-caught fix: `captureStdout` issued a single `r.Read`, capturing only the first chunk of multi-write output. Scheduling-dependent, so it passed locally and failed in CI. Now drains to EOF; verified with `-count=30` and `-race`.

Issue #439, opened 2026-08-24
search: search_live real-corpus test + performance measurement pass


vibexp-agent:
Part of epic #432. Blocked by #435 and #436.

## Scope

The measurement pass that turns "should be fast" into numbers — a `tests/search_live.rs` in the family of `scan_live` and `insights_live` (`#[ignore]`d, run by hand against the real corpus, because a three-file fixture cannot tell a healthy index from one that indexed nothing):

- Copy the real database, run a full index build over the real transcript corpus, and measure: cold build time, incremental reindex time for one changed session, index size on disk, and p50/p95 query latency for representative queries with filters and facets composed in.
- Assert correctness on real data: a query for known content returns the expected session; row counts are non-empty (the "an identical over an empty result set is not evidence" rule).
- Record the numbers in the PR and tune the #434 caps if tool-text dominates the index disproportionately.

## Acceptance criteria

- [ ] Test refuses to pass vacuously: zero indexed sessions or zero-size index is a failure, not a pass.
- [ ] Prints cold build time, per-session incremental time, index size, and p50/p95 query latency.
- [ ] A known-content query (chosen from the local corpus at run time) returns its session ranked on the first page.
- [ ] Index size and build time are within agreed bounds for the reference corpus, or the caps are adjusted and re-measured in the same PR.
- [ ] Run instructions documented alongside `scan_live`'s (`cargo test --test search_live -- --ignored --nocapture`).

Complexity: medium. Blocked by: #435, #436.

---
## WHAT
**Summary:** A real-corpus `search_live` test plus a measurement pass — the evidence that the index is correct and fast on real data, and the input for tuning the caps.
**Scope:** New `#[ignore]`d integration test in the `scan_live`/`insights_live` family; recorded measurements; cap tuning if warranted.
**Affected Components:** New `src-tauri/tests/search_live.rs`; possibly the cap constants in `src-tauri/src/native/search/normalize.rs` (tuning only).
**Acceptance Criteria:**
- [ ] Test refuses to pass vacuously: zero indexed sessions or a zero-size index is a failure (the "identical over an empty result set is not evidence" rule).
- [ ] Prints cold full-index build time, per-session incremental reindex time, index size on disk, and p50/p95 query latency with filters + facets composed.
- [ ] A known-content query, chosen from the local corpus at run time, returns its session on the first page.
[...]

Issue #437, opened 2026-08-24
search: match_snippet on the wire + relevance sort with keyset cursor


vibexp-agent:
Part of epic #432. Blocked by #436.

## Scope

Surface the match on the wire, on `GET /api/claude-sessions`:

- **`match_snippet`** per session row, produced by FTS5 `snippet()` (highlight markers included), populated only when `q` is set and the row matched via content. `skip_serializing_if` empty — every frozen parity golden stays byte-identical, and non-search responses carry no new field.
- **`sort=relevance`**: sort key is the weighted bm25 rank, tiebreak on the full row key `(session_id, project_path)` exactly as keyset pagination already requires (the #364 tie rule). It becomes the default when `q` is non-empty; an explicit `sort` param overrides.
- Cursor support: the cursor encodes the sort, so a relevance cursor round-trips and the existing cursor/sort-mismatch 400 applies unchanged.

## Acceptance criteria

- [ ] `match_snippet` present only on search responses; all existing parity goldens pass unmodified.
- [ ] Relevance sort paginates without skipping or duplicating rows across pages, including under bm25 ties (fixture with deliberate ties on rank, distinct row keys).
- [ ] `q` set with no explicit sort → relevance order; `q` set with `sort=recent` → recent order (override respected).
- [ ] A cursor minted under `relevance` used with a different `sort` answers the existing 400; a pre-existing cursor without the new field still decodes (Go missing-field rule, `#[serde(default)]`).
- [ ] Facets bar totals still agree with the row set under relevance sort.

Complexity: medium. Blocked by: #436.

---
## WHAT
**Summary:** Surface the match on the wire: an omit-when-empty `match_snippet` per row and a `relevance` sort (bm25) with full keyset-cursor support, defaulting on when `q` is set.
**Scope:** `GET /api/claude-sessions` response + sort/cursor semantics. Backend only.
**Affected Components:** `src-tauri/src/native/sessions/` — the list handler and row struct (`corpus.rs` / the module serving the list), `query.rs` (sort resolution + cursor), reusing `search::match_sql()` for `snippet()`/rank.
**Acceptance Criteria:**
- [ ] `match_snippet` present only when `q` is set and the row matched via content (`skip_serializing_if` empty); the full parity golden suite passes unmodified.
- [ ] Relevance sort paginates without skipping or duplicating rows across pages under deliberate bm25 ties (fixture with tied ranks, distinct row keys — the #364 tie rule).
[...]

Issue #436, opened 2026-08-24
search: FTS MATCH in add_search — safe query building + LIKE fallback


vibexp-agent:
Part of epic #432. Blocked by #433.

## Scope

Extend `sessions/query.rs::add_search` so a non-empty `q` composes:

```sql
(  (session_id, project_path) IN
     (SELECT session_id, project_path FROM session_search WHERE session_search MATCH ?)
   OR <the existing six-column LIKE clause> )
```

Query building (the injection boundary — safe by construction, not sanitization):

- Split on whitespace; wrap every token in double quotes with internal quotes escaped, so FTS operators (`-`, `OR`, `NOT`, `^`, `*`, parens) typed by a user match literally and can never produce a syntax error.
- Implicit AND between tokens — `fix auth bug` matches sessions containing all three words in any order.
- User-typed `"quoted phrases"` pass through as FTS phrases.
- The final token gets a `*` prefix for as-you-type matching.
- The LIKE clause stays OR'd in: it still matches session ids, project paths, and sessions the worker has not indexed yet.
- Degradation: if the FTS table is missing or MATCH errors, fall back to LIKE-only. Search must never answer 500.

Because this is one more `Filter` clause, facets, config-dir scoping, numeric/time filters and keyset pagination compose unchanged — assert it rather than assume it.

This issue is testable independently of #435 by hand-populating `session_search` in fixtures.

## Acceptance criteria

- [ ] Multi-word query matches all-words-any-order in indexed content; word order in the query is irrelevant.
- [ ] Every FTS metacharacter a user can type (`- * ^ ( ) : "` and bare `OR`/`NOT`/`AND`) matches literally; a fuzz-style test asserts no input string can make the query error.
- [ ] A session with no index row still matches via the metadata LIKE clause (id, titles, preview, project path).
- [ ] With the FTS table dropped, search answers exactly as today — no error surfaced.
- [ ] Facets computed under a search agree with the filtered row set.
- [ ] `%`/`_` LIKE-escaping behavior is unchanged (existing `search_wildcards_are_escaped_so_they_match_literally` still passes).

Complexity: medium. Blocked by: #433.

---
## WHAT
**Summary:** Compose an FTS5 MATCH into `add_search` with injection-proof query building, implicit AND, phrases, trailing-prefix matching, and total LIKE fallback.
**Scope:** Query construction + composition into the existing `Filter`; degradation path; no wire-format change.
[...]

Issue #435, opened 2026-08-24
search: indexing pipeline in the insight worker (populate, reconcile, version rebuild)


vibexp-agent:
Part of epic #432. Blocked by #433 and #434.

## Scope

Populate `session_search` from the insight worker (`native/insights/worker.rs`) — the component that already re-reads each changed transcript one session at a time and owns the queue, the 5-minute sweep, the boot sweep and the reconcile:

- In the per-session pass, build the index document (via the #434 normalizer, from the same transcript read that feeds the nine insight processors) and write it **in the same transaction** as the `session_insights` row, keyed on `(session_id, project_path)`.
- Reconcile: an index row whose cache row no longer exists is deleted. Keyed on "no cache row remains", never on a path — a moved transcript (claim shift, #245) is an update, not a delete, and rows protected by an unreadable config dir must survive.
- `search_index_version`: bumping the constant triggers a full reindex through the worker's sweep. It must NOT touch the scanner's staleness markers — a search-index rebuild must not force `claude_session_cache` transcript re-reads (same separation as `CURRENT_PROCESSOR_VERSION`).
- First-run behavior: on upgrade, the boot sweep indexes the whole corpus. All database work through `db::blocking`; the queue-overflow → sweep behavior (`SWEEP_REQUESTED`) covers the first-scan flood exactly as it does for insights.

## Acceptance criteria

- [ ] A session announced as changed by a scan has a fresh index row after one queue pass.
- [ ] Deleting a session's cache row removes its index row on reconcile; a claim-shifted (moved-path) session keeps one index row, updated, not deleted-and-orphaned.
- [ ] Bumping `search_index_version` reindexes every session while `claude_session_cache.file_mtime` values are untouched (no scanner re-read).
- [ ] Index write and `session_insights` write commit or fail together (one transaction).
- [ ] A contended write lock does not stall the runtime — the #366 test pattern (one-worker runtime, outside thread holding the lock, ticker gap bounded) passes with indexing active.
- [ ] Two cache rows sharing a `session_id` across projects get two independent index rows (the #362-family assertion).

Complexity: high. Blocked by: #433, #434.

---
## WHAT
**Summary:** Populate `session_search` from the insight worker: per-session index write in the same transaction as `session_insights`, reconcile on cache-row existence, and a `search_index_version` full-rebuild path.
**Scope:** Worker per-session pass, reconcile, version constant + sweep trigger, boot behavior on upgrade.
[...]

PR #441, opened 2026-08-24
feat(search): FTS5 foundation — session_search migration and store layer (#433)

### LINKED ISSUE #433: search: FTS5 foundation — migration, session_search schema, store layer
Part of epic #432.

## Scope

The storage foundation for full-text session search:

- New migration (authored in `parity/migrations_vectors.json`, applied by `native/migrate.rs`) creating the `session_search` FTS5 virtual table:
  ```sql
  CREATE VIRTUAL TABLE session_search USING fts5(
      title, user_text, assistant_text, tool_text,
      session_id UNINDEXED, project_path UNINDEXED,
      tokenize = 'unicode61 remove_diacritics 2'
  );
  ```
- `search_index_version INTEGER NOT NULL DEFAULT 0` on `claude_cache_metadata`, so a fresh install and every existing database start at "nothing indexed".
- A store layer (new `native/search/` module or similar) with: replace-by-pair (delete + insert keyed on `(session_id, project_path)`), delete-by-pair, delete-all (for version rebuilds), and a match-query helper exposing bm25 with column weights `8.0, 4.0, 2.0, 0.5`.
- All writes through `db::open_read_write` / `db::blocking` conventions.

FTS5 availability is a build-time fact (rusqlite `bundled` compiles `-DSQLITE_ENABLE_FTS5`; verified in `libsqlite3-sys 0.38.2`), so no runtime feature probe is needed — but the migration failing on a build without it must be a loud startup error, not a silent skip.

## Acceptance criteria

- [ ] Migration applies cleanly on both a fresh database and an existing populated one; `migrate::verify()` passes.
- [ ] Store layer can insert, replace and delete index rows keyed on `(session_id, project_path)`; a replace for one pair never touches the other project's row for the same `session_id`.
- [ ] Match helper returns rows ranked by the weighted bm25 (title hit outranks an equivalent tool-text hit — unit-tested).
- [ ] `search_index_version` reads back 0 on first run after upgrade.
- [ ] No existing endpoint's response changes by a byte (parity goldens untouched).

Complexity: medium. Blocked by: nothing.

---
## WHAT
**Summary:** Create the FTS5 storage foundation: the `session_search` virtual table (migration), the `search_index_version` metadata column, and a store layer for keyed writes and ranked reads.
**Scope:** Migration authored in `parity/migrations_vectors.json`; new `native/search` store module with replace/delete/delete-all/match helpers; bm25 weight constants.
**Affected Components:** `parity/migrations_vectors.json` — new migration (next number after current tail); `src-tauri/src/native/migrate.rs` — `verify()` covers the new objects; new `src-tauri/src/native/search/mod.rs` (store layer); `src-tauri/src/native/mod.rs` — module declaration only (no endpoint yet).
**Acceptance Criteria:**
- [ ] Migration applies on a fresh database and on an existing populated one; `migrate::verify()` passes both.
- [ ] `replace`/`delete` are keyed on `(session_id, project_path)`; a replace for one pair never touches the other project's row for the same `session_id` (unit test with two pairs sharing an id).
- [ ] Match helper ranks a title hit above an equivalent `tool_text` hit (bm25 weights 8/4/2/0.5, unit-tested).
- [ ] `search_index_version` reads back 0 after upgrade on an existing database.
- [ ] No existing endpoint's response changes by a byte (full parity golden suite passes unmodified).
**Out of Scope:** Populating the index (#435), querying it from the sessions list (#436), any API/UI change.

## WHY
**Business Value:** Foundation for content search — the sessions list currently searches only six metadata columns via `LIKE`, so text inside a conversation is unfindable. · **Problem Statement:** Conversation text never reaches the database; only a 120-char `preview` is cached.
**Expected Impact:** Enables the two parallel lanes of epic #432 (indexer #435, query #436) to proceed independently once this lands.
**Context:**
Parent: #432
Blocks: #435
Blocks: #436

## HOW
**Technical Approach:** A regular (content-storing) FTS5 table — required for `snippet()`/`highlight()` in #437 — keyed by UNINDEXED `session_id`/`project_path` columns, following the codebase rule that every statement over the corpus keys on the pair (#362 family). FTS5 is compiled in via rusqlite `bundled` (`-DSQLITE_ENABLE_FTS5`, verified in `libsqlite3-sys 0.38.2`).
**Implementation Details:**
- Migration SQL: `CREATE VIRTUAL TABLE session_search USING fts5(title, user_text, assistant_text, tool_text, session_id UNINDEXED, project_path UNINDEXED, tokenize = 'unicode61 remove_diacritics 2');` plus `ALTER TABLE claude_cache_metadata ADD COLUMN search_index_version INTEGER NOT NULL DEFAULT 0;` — authored directly in `parity/migrations_vectors.json` (migrations ≥31 are authored there per `parity/README.md`).
- `native/search/mod.rs`: `pub struct SearchDoc { session_id, project_path, title, user_text, assistant_text, tool_text }`; `pub fn replace(conn, &SearchDoc)` (DELETE by pair + INSERT, one transaction supplied by the caller); `pub fn delete(conn, session_id, project_path)`; `pub fn delete_all(conn)`; `pub const BM25_WEIGHTS: ...` and a `match_sql()` helper returning the `MATCH ?` subquery with `bm25(session_search, 8.0, 4.0, 2.0, 0.5)` for #436/#437 to share.
- Writers take a `&Connection` from the caller (the insight worker owns the transaction in #435) — the store layer itself opens nothing.
**Testing Strategy:** In-module unit tests against a temp database built through `db::ensure_database` + `migrate::apply` (fresh-install path) and against a database pre-seeded at the previous migration version (upgrade path). Ranking test inserts two docs differing only in which column holds the term.
**Documentation / Configuration:** Note the new table in `CLAUDE.md`'s layout section when the epic completes (not per-issue). No new dependencies.
**Files to Create / Modify:** Create `src-tauri/src/native/search/mod.rs`. Modify `parity/migrations_vectors.json`, `src-tauri/src/native/migrate.rs` (verify), `src-tauri/src/native/mod.rs` (mod decl).
**Risks:** Migration failure on a hypothetical FTS5-less build → must be a loud startup error, not a silent skip (accepted: `bundled` always compiles FTS5). Existing large databases: `CREATE VIRTUAL TABLE` is O(1) — no data backfill in this issue, so migration time is negligible.
**Complexity:** Medium (multi-file, <500 LOC, clear pattern)

---
**Refined by:** AI Issue Refiner · **Date:** 2026-08-24 · **Status:** Ready

vibexp-agent:
Closes #433. Part of epic #432.

## What

Migration **33** and a new `native/search/` store layer — the storage foundation for full-text session search. No endpoint is claimed, no response changes.

- **`session_search`**: a content-storing FTS5 virtual table over `title`, `user_text`, `assistant_text`, `tool_text`, plus `session_id`/`project_path` as UNINDEXED key columns, tokenized `unicode61 remove_diacritics 2`.
- **`claude_cache_metadata.search_index_version`** `INTEGER NOT NULL DEFAULT 0` — a fresh install and every existing database alike start at "nothing indexed", so the upgrade path needs no distinguishing case.
- **`native/search/mod.rs`**: `replace` / `delete` (both keyed on the whole pair), `delete_all` (version rebuilds), `search` (one ranked read), and the shared `match_sql()` / `BM25_EXPR` / `BM25_WEIGHTS` that #436 and #437 compose with.

## Why

The sessions list searches six metadata columns with `LIKE`, so text inside a conversation is unfindable — transcripts are JSONL on disk and only a 120-char `preview` is cached. This is where that text will land.

## How

- The migration is authored directly in `parity/migrations_vectors.json` as a **pure append** (4 inserted lines, spliced as text — never through a JSON round-trip, which would rewrite Go's `>` escaping across the frozen entries). Migrations 1–30 stay byte-identical and their SHA-256 pin is untouched.
- Content-storing rather than `content=''`, because `snippet()`/`highlight()` in #437 read the stored text.
- **Every statement keys on `(session_id, project_path)`** — the #362 family rule. A session id under two project paths is two legitimate rows, so a replace keyed on the id alone would silently delete the other project's text.
- Writers take a `&Connection` from the caller and open nothing: #435 writes the index row and the `session_insights` row in one transaction, so the transaction has to belong to it. `Transaction` derefs to `Connection`, so a caller passes `&tx`.
- FTS5 availability is a build-time fact — `libsqlite3-sys 0.38.2`'s `build.rs` passes `-DSQLITE_ENABLE_FTS5` unconditionally under `bundled`. A build without it fails at the migration, loudly, which is the intent.

## Note on the migration number

This was written as migration **32** and rebased to **33**: #422 / PR #440 was worked in parallel, took 32, and merged first. Nothing but the hardcoded count in `migrate.rs` catches that collision — appending a *sibling* object to the JSON array need not conflict textually — so `migrate.rs`'s module docs now say so in as many words. `a_database_at_any_earlier_version_upgrades_to_this_one`, which #440 added, is kept and covers this migration too.

## Acceptance criteria

- [x] Migration applies on a fresh database and on an existing populated one; `migrate::verify()` passes both — `the_migration_builds_an_fts5_table` and `the_migration_applies_to_a_populated_database` (seeds the singleton metadata row one version short of the tail, upgrades, asserts the row survives untouched).
- [x] `replace`/`delete` keyed on the pair; a replace for one pair never touches the other project's row — `a_replace_for_one_pair_leaves_the_other_projects_row`, `a_delete_for_one_pair_leaves_the_other_projects_row`.
- [x] Match helper ranks a title hit above an equivalent `tool_text` hit — `a_title_hit_outranks_the_same_hit_in_tool_text`, plus `the_four_columns_rank_in_weight_order`, which pins all four rather than just the two ends.
- [x] `search_index_version` reads back 0 after upgrade — `search_index_version_starts_at_zero` and the populated-database test.
- [x] No existing endpoint's response changes by a byte — the full frozen golden suite is unmodified and green (1207 tests across 13 binaries).

## Notes for reviewers

- **`migrate::verify()` is a version check, not a schema-object check**, so "verify covers the new objects" is satisfied by `expected_version()` moving to 33 (derived from the vector) plus the hardcoded counts in its tests. Those counts stay hardcoded on purpose — a count derived from the list agrees with itself no matter what the list lost — and they are the only thing that catches a number collision.
- **`bm25()` is negated: smaller is better**, and `rank` is *also* a magic FTS5 column (bm25 with all weights 1.0). `ORDER BY rank` resolves to the output alias, so the weighted expression sorts — but both resolutions return genuine hits in a plausible order, so a regression would be invisible. `the_returned_rank_is_the_weighted_bm25` compares the returned value against both spellings rather than trusting the rule.
- **A delete scans.** FTS5 will not index the UNINDEXED key columns and accepts no index on them, so `DELETE … WHERE session_id = ? AND project_path = ?` scans the backing `%_content` table. Correct and cheap at one row; not free at corpus scale, where `replace` runs once per changed session. Deferred to **#439** and cross-linked there — deliberately not solved here, since a rowid side table would be a schema change beyond this issue.
- `CLAUDE.md`'s layout line said "31 migrations" (main was already one behind, since #440 did not update it); corrected to 33. The layout entry for `native/search/` is deferred to epic completion per the issue's HOW.

## Local gates

`npm run build` · `cargo fmt --check` · `cargo clippy --all-targets -- -D warnings` · `cargo test --no-fail-fast` (1207 pass, 3 ignored) · `cargo build --release` — all green, re-run after the rebase.


vibexp-agent:
Deferred to #439: the `session_search` delete/replace scan (FTS5 accepts no index on the UNINDEXED key columns) — measured and decided in the epic's performance pass. Cross-linked at https://github.com/shaharia-lab/agento/issues/439#issuecomment-5395016079

Issue #438, opened 2026-08-24
search: UI — snippet rendering, relevance sort control, phrase hint


vibexp-agent:
Part of epic #432. Blocked by #437.

## Scope

The user-facing half, in `src/views/SessionsView.tsx` (+ a per-view CSS file per convention):

- Render `match_snippet` under the row title for content matches, with the matched terms highlighted. The snippet's highlight markers are the only markup honored — everything else is rendered as escaped text (no HTML injection from transcript content).
- Surface `relevance` in the sort control when a query is active, reflecting the server default and allowing override.
- A subtle hint that `"quotes"` mean phrase search (placeholder text or affordance near the field).
- Keep the 250 ms debounce; wire type update in `src/lib/types.ts` (`match_snippet` is optional — Go-style `omitempty` field, default it).

Desktop conventions apply: 14 px type, 28 px rows may grow for the snippet line, hairline borders, reuse existing CSS classes, focus-aware selection unchanged.

## Acceptance criteria

- [ ] A content match shows its snippet with highlighted terms; metadata-only matches (id/title/path) render as today.
- [ ] Snippet text originating from a transcript containing `<script>` or raw HTML renders inert (escaped), with only the highlight styling applied.
- [ ] Sort control shows and allows changing relevance when searching; leaving search restores the previous sort.
- [ ] Empty state ("No indexed session matches…") and all existing filters behave unchanged.
- [ ] List remains responsive while typing (debounce intact; no per-keystroke layout thrash from snippet rows).

Complexity: medium. Blocked by: #437.

---
## WHAT
**Summary:** Render content-match snippets with highlighted terms in the sessions list, surface the relevance sort, and hint phrase syntax — the user-facing half of epic #432.
**Scope:** `SessionsView` row rendering, sort control behavior while searching, wire-type mirror, per-view CSS.
**Affected Components:** `src/views/SessionsView.tsx` — snippet row + sort control; `src/lib/types.ts` — optional `match_snippet` on the session row type; a per-view CSS addition (existing sessions view stylesheet).
**Acceptance Criteria:**
- [ ] A content match shows its snippet under the row title with matched terms highlighted; metadata-only matches render as today.
- [ ] Snippet text containing `<script>` or raw HTML renders inert — highlighting is done by splitting on the server's sentinel markers into React elements, never `dangerouslySetInnerHTML`.
- [ ] Sort control shows and allows changing `relevance` while a query is active; clearing the search restores the previous sort.
[...]

Issue #434, opened 2026-08-24
search: transcript text normalizer (markdown → indexable text, size caps)


vibexp-agent:
Part of epic #432.

## Scope

A pure normalization module that turns decoded transcript message text into indexable plain text. No file I/O of its own — it consumes strings the existing `native/insights/transcript.rs` decoder already produces (so the `gojson::null_is_zero_value` decode rules apply upstream, not here).

Rules:

- Markdown link syntax `[text](url)` → keep `text`; drop the URL from indexed text (URLs are ranking noise).
- Strip code-fence language tags and HTML tags; keep code content itself (users search for code snippets and error strings).
- Collapse runs of whitespace so `snippet()` output reads cleanly in the UI.
- Enforce the size caps: ~8 KB per message, ~2 KB per tool result, ~512 KB per session (constants in one place, tunable after #439's measurements).

Note: full markdown parsing is deliberately NOT the job. FTS5's `unicode61` tokenizer already treats punctuation as separators, so `**bold**` indexes as `bold` and the phrase "some bold text" matches `some **bold** text` without any stripping. The normalizer exists for URL noise, snippet readability, and the caps.

## Acceptance criteria

- [ ] `[link text](https://…)` indexes as `link text`; the URL does not match.
- [ ] Phrase queries match across markdown markers (tokenizer-level test: "some bold text" vs source `some **bold** text`).
- [ ] All three caps enforced, with boundary unit tests (cap-1, cap, cap+1 — char-boundary safe, no mid-UTF-8 truncation panic).
- [ ] Deterministic pure function; a pathological input (one multi-MB line, deeply nested markdown) completes without quadratic blowup.
- [ ] No dependency on the database or filesystem — testable with plain strings.

Complexity: medium. Blocked by: nothing.

---
## WHAT
**Summary:** A pure text-normalization module turning decoded transcript message text into indexable plain text, with the three size caps.
**Scope:** Markdown-link reduction, tag stripping, whitespace collapsing, per-message / per-tool-result / per-session caps. Pure functions only — no I/O, no database.
**Affected Components:** New `src-tauri/src/native/search/normalize.rs` — the whole change (plus its `pub mod` line in `native/search/mod.rs` once #433 exists, or a standalone module if this lands first — the two issues are deliberately independent).
**Acceptance Criteria:**
- [ ] `[link text](https://…)` normalizes to `link text`; the URL does not appear in output.
[...]

Issue #432, opened 2026-08-24
Epic: Full-text search for the Claude sessions list (FTS5 content index)


vibexp-agent:
## Problem

The sessions list search is a `LIKE '%q%'` over six metadata columns (`src-tauri/src/native/sessions/query.rs::add_search`): `session_id`, `preview`, `custom_title`, `native_title`, `ai_title`, `project_path`. Conversation text never reaches the database — transcripts are JSONL files on disk, and only a 120-char `preview` of the first genuine user message is cached — so exact text a user remembers from a session body finds **nothing**. The single-substring match also fails on word order (`fix auth bug` will not match "auth bug fix"), there is no ranking, and results give no indication of why they matched.

## Solution

An SQLite **FTS5** content index (`session_search`), populated by the insight worker's existing per-session transcript pass (zero extra file reads), queried through the existing `Filter` machinery so facets, numeric/time filters, config-dir scoping and keyset pagination compose unchanged. Markdown is neutralized by tokenization plus a light index-time normalizer, so raw-text queries match rendered text. Results gain a highlighted `match_snippet` and a relevance sort.

FTS5 is already compiled into this build: rusqlite's `bundled` feature builds SQLite with `-DSQLITE_ENABLE_FTS5` (verified in `libsqlite3-sys 0.38.2` from `Cargo.lock`). No new dependency.

## Design decisions

1. **FTS5, in-process, same DB file** — rejected alternatives: full text as a LIKE-scanned column (linear scans, no ranking), tantivy (second on-disk index beside SQLite, heavy dependency, own consistency story), on-demand ripgrep over JSONL (seconds per keystroke, cannot compose with SQL filters, matches JSON-escaped bytes).
2. **Index keyed on `(session_id, project_path)`** via UNINDEXED columns — the #362-family rule; `session_id` alone is not unique. Reconcile keyed on "no cache row remains", never on a path (a claim shift is an update, not a delete — #245).
3. **Indexing lives in the insight worker** (`native/insights/worker.rs`), not the scanner — it already re-reads each changed transcript one session at a time (bounded memory), and has the queue, 5-minute sweep, boot sweep and reconcile the index needs. Its own `search_index_version` must NOT force scanner re-reads (same reasoning as `CURRENT_PROCESSOR_VERSION`).
[...]

PR #444, opened 2026-08-24
feat(search): indexing pipeline in the insight worker (#435)

### LINKED ISSUE #435: search: indexing pipeline in the insight worker (populate, reconcile, version rebuild)

vibexp-agent:
Closes #435. Part of epic #432, on top of #433 (FTS5 foundation) and #434 (the normalizer).

## What

`session_search` is now populated by the insight worker. The per-session pass builds the index document from the **same decoded transcript** the nine insight processors already consume, and writes it in the **same transaction** as the `session_insights` row — so indexing costs no additional file I/O and index freshness equals insight freshness.

## How

- **`insights/index.rs`** (new) — routes a decoded event's text into the three columns. `search::normalize` (#434) still owns all the *shaping*; this module only decides where text goes.
- **`processors::run`** takes a `&mut DocAccumulator` and `feed` observes each event alongside the processors. A parameter rather than a tenth `Processor`, because a `Processor` finalizes into a `SessionInsight`, which has nowhere to put three text columns — and because `PIPELINE`'s order is load-bearing for the numbers.
- **`store::select_pending`** gains a `Scope`, and **`store::display_title`** reads the cache row's title with the UI's own precedence.
- **`search::{insert, delete_orphans, stored_version, record_version}`** plus `SEARCH_INDEX_VERSION`.
- **`scan.rs`** reconciles the index beside the insights, after the delete pass.

## The rules that are silent when wrong

- **Routing is by content shape, never by `isSidechain`.** `is_user_turn_content` answers `false` to a `tool_result` carrier *and* to injected machinery, so the `tool_result` check has to come first and be its own — routing on the predicate alone drops every real tool result, and `tool_text` would still look populated (from `tool_use` blocks) while never containing an error message.
- **Everything keys on `(session_id, project_path)`** — the #362 family in its sixth form. `session_search` is a *new* table, so nothing forced the pair on it; keyed on the id alone, `replace`'s delete removes both projects' rows.
- **The reconcile keys on "no cache row remains", never on a path** (#245), and runs *after* the cache's delete pass so it inherits the unreadable-config-dir protection.
- **A `SEARCH_INDEX_VERSION` bump touches no scanner staleness marker**, so `claude_session_cache.file_mtime` is untouched and the scanner re-reads nothing.
- **An unreadable transcript is a skip, not a batch failure.** This looks lenient and is strict: every real corpus holds a session whose file has since been deleted, so counting a skip as failure would mean the version is never recorded on any real machine and the whole index rebuilds every five minutes for ever.
- **A rebuild empties the index once rather than replacing per session** — see performance below.

## Performance

`search::delete` is keyed on the UNINDEXED key columns, and FTS5's `xBestIndex` accepts only `rowid` and `MATCH`, so it is a **full scan of the `%_content` table** (~9.5 ms at 17 MB, ~63 ms at 176 MB, per call). Re-indexing a corpus with `replace` per session is therefore quadratic, inside a write transaction competing with the boot scan `lib.rs` starts two lines later.

A rebuild now calls `delete_all` once and uses the new `search::insert`. Measured over the real corpus: **1,178 sessions rebuild in 20.7s, down from 54s.**

That also removes the worse half of a silent failure — a session the rebuild skips now has *no* row, where per-session replacement left its previous row carrying the old format's text under a version stamp saying otherwise. Missing is recoverable; silently stale is not.

## Acceptance criteria

- [x] A session announced as changed has a fresh index row after one queue pass — asserted through `search::search`, not a row count.
- [x] Reconcile removes an index row only when no cache row remains; a claim-shifted session keeps one updated row.
- [x] Bumping `SEARCH_INDEX_VERSION` reindexes every session while every `file_mtime` is untouched.
- [x] Index write and `session_insights` write commit or fail together (fault-injected by dropping the table).
- [x] The #366 contended-write-lock pattern passes with indexing active — see the honesty note below.
- [x] Two cache rows sharing a `session_id` across projects produce two independent index rows, each with its own transcript's text.

## Verification

Each core rule was **falsified before being trusted** — every test below fails against the specific defect it exists to catch:

| Injected defect | Test that fails |
|---|---|
| index delete keyed on `session_id` alone | `one_session_id_under_two_projects_gets_two_index_rows`, `the_reconcile_drops_only_rows_with_no_cache_row` |
| insight and index in two transactions | `a_failed_index_write_rolls_back_the_insight_row_with_it` |
| rebuild never widens the sweep's scope | `a_version_bump_reindexes_without_touching_the_scanners_mtimes` |
| a skipped transcript counted as batch failure | `a_rebuild_completes_over_a_corpus_with_an_unreadable_transcript` |
| batch moved onto a runtime task | the contention test (ticker: ~150 advances → 0) |

**One honesty note.** The contended-write-lock test is named `a_contended_batch_still_commits_within_the_busy_timeout`, not for #366's property, because the worker owns an OS thread by construction and `run()` is an unexported infinite loop — so no change to non-test code can make it fail. It genuinely pins that a batch (which now writes two tables) still commits inside the 5s `busy_timeout` under outside contention. Making it falsifiable against production code needs a `run_once` extraction: #447.

Local gates: `npm run build`, `cargo fmt --check`, `cargo clippy --all-targets -- -D warnings`, `cargo test --no-fail-fast` (14 suites, green on repeated runs), `cargo build --release` — all re-run after rebasing onto the two gateway PRs that landed mid-review (which added migration 34; this change adds none).

**Against the real corpus** (`cargo test --test insights_live -- --ignored`): 1,178 cached sessions → **1,178 insight rows and 1,178 index rows**, reachable through FTS5, version stamped, in ~21s.

That run also caught a harness gap this PR fixes: `insights_live.rs` started the worker on an **unmigrated** copy of the installed database. The machine's real database is at migration 31, where `session_search` does not exist — so every batch's index write failed and, because the two writes are deliberately one transaction, took the insight write with it. The test reported *0 insight rows over 1,178 sessions* against a build that was fine. Production is unaffected (`lib.rs` migrates at startup, long before `worker::start`), but the copy has to be migrated the same way.

## Deferred

- **#446** — per-row index version, so a rebuild retries the sessions it skipped. The global stamp cannot describe a partial rebuild: an unreadable-but-still-cached session stays unindexed until its transcript next changes.
- **#447** — `run_once` extraction, plus queue-path and `scan.rs` reconcile-wiring coverage.

Out of scope: query-side changes (#436), snippet/sort (#437), the UI (#438), and the `CLAUDE.md` update, which #435 defers to epic close.

One limitation recorded at `store::display_title`: `custom_title` is user-writable through `PATCH /api/claude-sessions/{id}`, which changes neither `file_mtime` nor `processor_version` — so a rename alone does not re-index; the title catches up the next time the transcript changes.

### DIFF
--- a/src-tauri/src/native/insights/index.rs
+++ b/src-tauri/src/native/insights/index.rs
@@ -103,16 +103,21 @@ impl DocAccumulator {
     }
 
     fn observe_user(&mut self, content: &Value) {
-        let blocks = transcript::parse_content_blocks(content);
         let mut carried_a_result = false;
-        for block in blocks.iter().filter(|b| b.block_type == "tool_result") {
+        for payload in tool_result_payloads(content) {
             carried_a_result = true;
-            self.builder.push_tool(&tool_result_text(block));
+            self.builder
+                .push_tool(&transcript::extract_text_content(payload));
         }
         if carried_a_result {
             return;
         }
-        if transcript::is_user_turn_content(content) {
+        // `is_injected_user_content`, not `is_user_turn_content`. The latter is
+        // "no `tool_result` block **and** not injected", and the early return
+        // above has already settled the first half — so calling it here would
+        // re-run `parse_content_blocks` over the same array to re-derive an
+        // answer we have, and that call clones every block's `Value`.
+        if !transcript::is_injected_user_content(content) {
             self.builder
                 .push_user(&transcript::extract_text_content(content));
         }
@@ -167,14 +172,33 @@ pub fn normalize_title(raw: &str) -> String {
     normalize::normalize_text(raw, normalize::MESSAGE_CAP)
 }
 
-/// A `tool_result` block's payload as text.
+/// Each `tool_result` block's payload, borrowed straight out of the message's
+/// content array.
 ///
-/// The block's `content` is a string for most tools and an array of `text`
-/// blocks for the rest, which is exactly the shape
-/// [`transcript::extract_text_content`] already decides between — so this is
-/// that function plus the `is_error` note below, not a second decoder.
-fn tool_result_text(block: &ContentBlock) -> String {
-    transcript::extract_text_content(&block.content)
+/// **Deliberately not routed through [`transcript::ContentBlock`]**, which is
+/// the obvious way to write this and is the expensive one. That struct does not
+/// decode a `tool_result`'s `content`, and adding the field to it would reach
+/// far past this module: `parse_content_blocks` is called four times per message
+/// by the processors, and `parse_content_blocks_raw` is on the
+/// `GET /api/claude-sessions/{id}` read path — where the payload is currently
+/// skipped by the decoder entirely. A `Read` of a large file is a multi-megabyte
+/// `tool_result`, so adding the field would make every one of those callers
+/// materialize and retain payloads none of them look at, to serve one reader
+/// here.
+///
+/// Borrowing costs nothing instead: the content is already a `Value` on
+/// `Message`, so this only walks it. The block shape is checked by hand for the
+/// same reason — one `type` comparison against a shared type's whole decode.
+fn tool_result_payloads(content: &Value) -> impl Iterator<Item = &Value> {
+    content
+        .as_array()
+        .into_iter()
+        .flatten()
+        .filter(|block| block.get("type").and_then(Value::as_str) == Some("tool_result"))
+        // The payload a tool's output arrives in. Absent is `Value::Null`, which
+        // `extract_text_content` answers `""` to, so a malformed block
+        // contributes nothing rather than needing its own arm.
+        .map(|block| block.get("content").unwrap_or(&Value::Null))
 }
 
 /// A `tool_use` block as text: the tool's name, then the string leaves of its
@@ -202,11 +226,16 @@ fn tool_use_text(block: &ContentBlock) -> String {
 /// Append every string leaf of `value`, in document order, space separated.
 ///
 /// **Bounded by [`normalize::TOOL_RESULT_CAP`]**, which is the cap the result
-/// will be truncated to anyway: a `Write` call's input carries a whole file, and
-/// building a multi-megabyte `String` to hand `push_tool` two kilobytes of it is
-/// the shape #434's own "every forward scan needs a bound" note is about. The
-/// bound is checked between leaves rather than inside one, so a single enormous
-/// leaf is still materialized once — `normalize_text` is what trims it.
+/// will be truncated to anyway: building a multi-megabyte `String` so that
+/// `push_tool` can keep two kilobytes of it is the shape #434's own "every
+/// forward scan needs a bound" note is about.
+///
+/// The bound is applied **inside a leaf as well as between leaves**, and that is
+/// the half worth stating: the case it exists for is a `Write` call, whose input
+/// is `{"file_path": "…", "content": "<the whole file>"}` — *one* enormous leaf.
+/// A guard that only checked between leaves would leave the dominant case
+/// completely uncovered while looking as though it handled it. The many-small-
+/// leaves shape is real too, so both are tested.
 fn push_string_leaves(value: &Value, out: &mut String) {
     if out.len() >= normalize::TOOL_RESULT_CAP {
         return;
@@ -215,7 +244,10 @@ fn push_string_leaves(value: &Value, out: &mut String) {
         Value::String(s) => {
             if !s.is_empty() {
                 out.push(' ');
-                out.push_str(s);
+                out.push_str(truncate_on_boundary(
+                    s,
+                    normalize::TOOL_RESULT_CAP.saturating_sub(out.len()),
+                ));
             }
         }
         Value::Array(items) => {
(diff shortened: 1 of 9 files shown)


vibexp-agent:
Deferred to #446 (per-row search index version, so a rebuild retries the sessions it skipped) and #447 (worker `run_once` extraction, plus queue-path and scan-wiring coverage).

PR #462, opened 2026-08-25
fix(desktop): seal the macOS bundle, so an Apple Silicon download is not "damaged" (#461)

### LINKED ISSUE #461: bug(desktop): Apple Silicon macOS builds are "damaged and can't be opened" — the .app is never codesigned, so it has no resource seal
## What happens

Downloading `Agento_1.0.0_aarch64.dmg` from the [v1.0.0 release](https://github.com/shaharia-lab/agento/releases/tag/v1.0.0) in a browser, mounting it and dragging **Agento** to Applications, the first launch fails with:

Image

The dialog offers **only** Move to Trash and Cancel. This is not the "unidentified developer" prompt — **there is no "Open Anyway" button, and no corresponding line appears in System Settings → Privacy & Security**, so the path that `README.md`, `docs/installation.md` and `docs/troubleshooting.md` all tell macOS users to take does not exist for this build. The app cannot be launched at all by following the documentation.

## Environment

| | |
|---|---|
| Machine | MacBook Pro, `Mac15,6` |
| Chip | Apple M3 Pro (11-core: 5P + 6E) |
| Memory | 18 GB |
| Architecture | `arm64` (Apple Silicon) |
| macOS | 26.6 (build `25G72`) |
| Agento | 1.0.0 (`com.shaharialab.agento`), Tauri 2 |
| Asset | `Agento_1.0.0_aarch64.dmg` — `sha256 9580d584e1ffe7a258f293dbdf0dca5fa716f0c89865d9c0c407171a80de1597` |
| How it was obtained | Downloaded in Chrome from the release page (so `com.apple.quarantine` is set) |

The download is not corrupt: re-downloading the same asset with `gh release download` reproduces the identical hash, and every finding below is from that clean copy.

## Root cause: the `.app` is never codesigned, so its resource seal is missing

`tauri build` bundles the app but **never runs `codesign`** on it. The workflow (`.github/workflows/release.yml`) sets `TAURI_SIGNING_PRIVATE_KEY` — that is Tauri's own minisign key for the *updater manifest*, nothing to do with Apple — and sets none of `APPLE_SIGNING_IDENTITY`, `APPLE_CERTIFICATE`, `APPLE_CERTIFICATE_PASSWORD`. `src-tauri/tauri.conf.json` has no `bundle.macOS.signingIdentity` either. In `tauri-bundler`, `sign::keychain()` returns `None` when neither is present, and the whole sign/notarize block in `macos/app.rs` is skipped.

What ships instead is the **ad-hoc signature the linker applies automatically to every arm64 Mach-O**. That covers the executable but produces no `Contents/_CodeSignature/CodeResources` for the bundle, and a bundle whose embedded signature declares resources while sealing none is **invalid** — not merely untrusted:

```console
$ codesign -dv --verbose=4 /Volumes/Agento/Agento.app
Executable=/Volumes/Agento/Agento.app/Contents/MacOS/agento
Identifier=agento-527c51eb3773d164          # linker-generated, not com.shaharialab.agento
Format=app bundle with Mach-O thin (arm64)
CodeDirectory v=20400 size=125784 flags=0x20002(adhoc,linker-signed) hashes=3927+0
Signature=adhoc
Info.plist=not bound
TeamIdentifier=not set
Sealed Resources=none

$ codesign --verify --deep --strict --verbose=2 /Volumes/Agento/Agento.app
/Volumes/Agento/Agento.app: code has no resources but signature indicates they must be present

$ ls /Volumes/Agento/Agento.app/Contents/_CodeSignature
ls: .../Contents/_CodeSignature: No such file or directory
```

Apple's own pre-distribution checker calls it **fatal**:

```console
$ syspolicy_check distribution /Volumes/Agento/Agento.app
App has failed one or more pre-distribution checks.
---------------------------------------------------------------
Codesign Error
    File: Agento.app
    Severity: Fatal
    Full Error: Code has no resources but signature indicates they must be
        present. Please re-sign using the latest version of macOS and codesign.
    Type: Notary Error
    Suggested Fix: Re-sign the application using the codesign command.
...
Notary Ticket Missing
    Severity: Fatal
```

That is the whole bug. Gatekeeper evaluates a quarantined bundle, finds a signature that fails integrity, and reports **damaged** — the verdict reserved for a broken signature, which deliberately has no user override. An app with *no* signature, or with a valid-but-unnotarized one, gets the ordinary "cannot verify the developer" prompt that *does* offer Open Anyway.

The executable's own signature is fine in isolation — copy the Mach-O out of the bundle and it verifies:

```console
$ cp /Volumes/Agento/Agento.app/Contents/MacOS/agento ./agento-bare
$ codesign --verify --verbose=2 ./agento-bare
./agento-bare: valid on disk
./agento-bare: satisfies its Designated Requirement
```

So the app runs fine un-quarantined (in CI, and for anyone who built it locally) and fails only for the one population that matters: people who download it. That is why this got through.

## Why Intel is unaffected, and why the docs read as if this works

The two macOS legs fail differently, because the linker only force-signs `arm64`:

| Asset | Signature state | Quarantined launch |
|---|---|---|
| `Agento_1.0.0_aarch64.dmg` | `adhoc, linker-signed`, `Sealed Resources=none` → **invalid bundle** | ❌ "damaged", **no Open Anyway** |
| `Agento_1.0.0_x64.dmg` | `code object is not signed at all` | ⚠️ "unidentified developer", **Open Anyway works** |
| `Agento_aarch64.app.tar.gz` (updater payload) | same defect as the aarch64 `.dmg` | in-app updates are not quarantined, so this currently still works |
| `Agento_1.0.0_aarch64.dmg` (the disk image itself) | `code object is not signed at all` | — |

```console
$ codesign --verify --deep --strict /Volumes/Agento/Agento.app   # x64 dmg
/Volumes/Agento/Agento.app: code object is not signed at all
```

So the documented "click Open Anyway" instructions are correct on Intel and impossible on Apple Silicon — i.e. on every Mac sold since 2020. Three places currently promise the wrong thing:

- `README.md` → *"macOS blocks it once … click **Open Anyway**"*
- `docs/installation.md` → the numbered Privacy & Security steps
- `docs/troubleshooting.md` → *"### macOS says the app is damaged or cannot be verified"* — names the exact symptom and then gives the remedy that does not apply to it
- `.github/workflows/release.yml` → the release-notes body baked into every release: *"open **System Settings → Privacy & Security** and choose **Open Anyway**"*

## Verified fix: ad-hoc sign the bundle

Re-signing the shipped app locally repairs it completely, which confirms the diagnosis and the remedy:

```console
$ cp -R /Volumes/Agento/Agento.app ./
$ codesign --force --deep --sign - ./Agento.app
./Agento.app: replacing existing signature

$ codesign --verify --deep --strict --verbose=2 ./Agento.app
./Agento.app: valid on disk
./Agento.app: satisfies its Designated Requirement

$ ls ./Agento.app/Contents/_CodeSignature
CodeResources

$ syspolicy_check distribution ./Agento.app
Adhoc Signed App
    Severity: Warning          # ← was: Codesign Error / Fatal
Notary Ticket Missing
    Severity: Fatal
```

The fatal codesign error is gone; what remains is only "ad-hoc signed" (a warning) and "not notarized" — which is exactly the state that produces the **Open Anyway** flow the docs already describe.

### Option A — ad-hoc signing in the build (free, no Apple Developer account)

One line, and `tauri build` does the `codesign` itself:

```diff
   "macOS": {
-    "minimumSystemVersion": "10.15"
+    "minimumSystemVersion": "10.15",
+    "signingIdentity": "-"
   },
```

`tauri-macos-sign` runs `codesign --force -s <identity> …`, so `-` is codesign's ad-hoc identity and produces the proper seal. Because the bundler signs the `.app` before it builds the `.dmg` and the `.app.tar.gz`, all three artifacts inherit it. It also fixes local dev builds, and `APPLE_SIGNING_IDENTITY` still overrides it in CI later without touching the config.

**This will not break the release build.** With an identity set, the bundler then *attempts* notarization, but `notarize_auth()` returning `MissingCredentials` only logs `skipping app notarization` and continues. The one hard-failure case is setting `APPLE_ID` + `APPLE_PASSWORD` **without** `APPLE_TEAM_ID` (`NotarizeAuthError::MissingTeamId`) — so if notarization is wired up later, all three must land together.

This does **not** make the warning go away — it turns an unfixable dead end into the one-time, documented Open Anyway prompt. That is the whole win, and it costs nothing.

### Option B — Developer ID + notarization (the real fix, $99/yr)

Removes the prompt entirely, on both arches, and is the only option that also gets the `.dmg` notarized. Needs an Apple Developer Program membership and these repo secrets, consumed by `tauri build` with no workflow logic beyond passing them through:

- `APPLE_CERTIFICATE` (base64 `.p12`) + `APPLE_CERTIFICATE_PASSWORD`
- `APPLE_SIGNING_IDENTITY` (e.g. `Developer ID Application: Shaharia Lab (TEAMID)`)
- `APPLE_ID` + `APPLE_PASSWORD` (app-specific) + `APPLE_TEAM_ID`

Worth doing eventually; Option A is worth doing today either way, since it is the fallback whenever those secrets are absent (forks, dry runs, local builds).

## Workaround for anyone hitting this now

Copy Agento to Applications first (the mounted `.dmg` is read-only), then either:

```bash
# strip the quarantine flag, so Gatekeeper never assesses the bundle
xattr -dr com.apple.quarantine /Applications/Agento.app

# …or repair the signature, which is what the build should have done
codesign --force --deep --sign - /Applications/Agento.app
```

Both are one-time. Note that `docs/troubleshooting.md` already lists the `xattr` command, but only as a fallback *after* the Privacy & Security steps that Apple Silicon users never get.

## Also broken, found while diagnosing this

The other documented escape hatch does not currently work either. The release notes suggest `brew install --cask --no-quarantine …`, but `shaharia-lab/homebrew-tap`'s `Casks/agento.rb` is pinned to **`version "0.1.1"`** and builds its URL as `releases/download/desktop-v#{version}/…`. Release tags dropped the `desktop-` prefix at 1.0.0 (#420), and the cask's `livecheck` regex still only matches `desktop-v*`, so it will not see 1.0.0 either — `brew install --cask` installs a three-versions-old build. The cask also does not set `quarantine: false`, so without `--no-quarantine` a Homebrew install hits this same "damaged" error.

## Acceptance criteria

- [ ] `codesign --verify --deep --strict Agento.app` passes on the `aarch64` **and** `x64` artifacts from a release build, and `Contents/_CodeSignature/CodeResources` exists in both.
- [ ] `syspolicy_check distribution Agento.app` reports no **Fatal** *Codesign Error* (a "Notary Ticket Missing" fatal is expected until Option B).
- [ ] A browser-downloaded `.dmg` on Apple Silicon produces the "unverified developer" prompt with a working **Open Anyway**, not "damaged".
- [ ] The `.app.tar.gz` updater payloads carry the same valid signature, so the updater does not reintroduce an unsealed bundle.
- [ ] README, `docs/installation.md`, `docs/troubleshooting.md` and the release-notes body in `release.yml` describe what Apple Silicon users actually see.
- [ ] Homebrew cask bumped to the current `v*` tag scheme and version, with the quarantine behaviour stated.

amims71:
Fixes #461.

## The bug

Nothing ran `codesign` over the `.app`. The arm64 artifact still carried the ad-hoc signature the **linker** applies to every arm64 Mach-O, which seals the executable and writes no `Contents/_CodeSignature/CodeResources` — so the bundle's signature declared resources it did not seal. macOS reads that as **damaged**, not unsigned, and that dialog offers only *Move to Trash*: no Open Anyway, no Privacy & Security entry. Every Apple Silicon download of 1.0.0 was unlaunchable by the route `docs/installation.md` describes.

```console
$ codesign --verify --deep --strict /Volumes/Agento/Agento.app          # shipped v1.0.0 aarch64
/Volumes/Agento/Agento.app: code has no resources but signature indicates they must be present

$ syspolicy_check distribution /Volumes/Agento/Agento.app
Codesign Error … Severity: Fatal
```

## The fix

One line — `bundle.macOS.signingIdentity: "-"` — so `tauri build` runs `codesign --force -s - --options runtime` over the bundle. It carries no Apple identity and buys no trust; it buys a **valid** signature, which is the only thing standing between "damaged" and the ordinary unverified-developer prompt the docs already describe.

Set in `tauri.conf.json` rather than as `APPLE_SIGNING_IDENTITY` in the workflow, so a local `tauri build` produces the same sealed bundle CI does and the new verify step checks an artifact a developer can reproduce. The env var still overrides it when a real Developer ID arrives, so that path needs no edit here.

This does not remove the first-launch warning. It converts an unfixable dead end into the one-time Open Anyway prompt, at no cost. Notarisation (#461, Option B) remains the real fix and needs a paid account.

## Why a CI guard, not a test

Three things kept this hidden for a whole release:

- **It needs quarantine.** The executable's own signature is valid, so the app ran from CI, from a local `tauri build`, and after every in-app update — the updater's payload is not quarantined. Only a browser download reproduces it, so there is no unit test to write.
- **Intel was unaffected**, for the opposite reason: nothing force-signs an x86_64 Mach-O, so that bundle was unsigned outright, and unsigned earns the Open Anyway prompt. The two macOS legs failed differently, which is how the docs came to describe a path only half the users had.
- **`codesign -dv` exits 0** on the broken bundle and prints a plausible-looking ad-hoc signature. Only `--verify --deep --strict` rejects it, and only `Identifier=` separates a bundle codesign sealed (`com.shaharialab.agento`, from `CFBundleIdentifier`) from one the linker signed (`agento-<hash>`).

So the guard runs on both macOS legs and checks all three: strict verify, the seal file by name, and the identifier.

It checks the `.app` under `bundle/macos/` rather than mounting the `.dmg`, because both the `.dmg` and the `.app.tar.gz` are built from that bundle after signing, and a bundle signature lives in real files rather than xattrs, so a copy cannot strip it. The shipped v1.0.0 `.dmg` and `.app.tar.gz` were confirmed to carry byte-identical signature state, which is the evidence for that.

## Verification

Against the **shipped v1.0.0 artifacts**, reproducing Tauri's exact invocation (`xattr -crs`, then `codesign --force -s - --options runtime` inside-out):

| | before | after |
|---|---|---|
| `codesign --verify --deep --strict` | `code has no resources but signature indicates they must be present` | `valid on disk` / `satisfies its Designated Requirement` |
| `Identifier=` | `agento-527c51eb3773d164` | `com.shaharialab.agento` |
| flags | `0x20002(adhoc,linker-signed)` | `0x10002(adhoc,runtime)` |
| `Sealed Resources` | `none` | `version=2 rules=13 files=1` |
| `Info.plist` | `not bound` | `entries=15` |
| `syspolicy_check` | **Fatal** Codesign Error | ad-hoc *Warning* only |

**It runs.** Launched with `AGENTO_DATA_DIR` pointed at a scratch directory, the resealed hardened-runtime build boots, seeds the pricing catalog, creates its API signing keypair, listens on 127.0.0.1, starts the scheduler and completes a 1,662-session Claude scan. So ad-hoc + hardened runtime is not a runtime regression — which is why `hardenedRuntime` is left at its default `true`: notarisation requires it anyway, and turning it off now would only defer finding out.

**The guard was tested against three bundles**, by extracting the step's script straight from the workflow YAML and running it under `bash -eo pipefail`:

| bundle | result |
|---|---|
| shipped v1.0.0 `.app.tar.gz`, untouched | ❌ exit 1 at strict verify |
| the same bundle, resealed | ✅ exit 0, `sealed and signed as com.shaharialab.agento` |
| resealed but re-signed `-i agento-527c51eb3773d164` | ❌ exit 1 at the identifier check |

`pre-commit run` passes on every changed file. Note I could not run `cargo`/`tauri build` locally — there is no Rust toolchain on this machine — so the end-to-end build is CI's to confirm; the config value's effect was verified by reproducing the bundler's own `codesign` call, which `tauri-macos-sign` builds as `--force -s <identity>` plus `--options runtime` for executables.

## Docs

`docs/troubleshooting.md` named this exact symptom — *"macOS says the app is damaged or cannot be verified"* — and then gave the Open Anyway remedy that cannot apply to it. It now separates the two dialogs, explains why "damaged" has no override, and gives both escapes (`xattr -dr` or a local re-seal) for copies already downloaded. `README.md` and `docs/installation.md` gain a pointer to it; `docs/releasing.md` and `CLAUDE.md` record why `signingIdentity` must stay.

The release-notes body drops its `brew install --cask --no-quarantine` suggestion: `shaharia-lab/homebrew-tap`'s `Casks/agento.rb` is still pinned to `0.1.1` with `desktop-v#{version}` URLs and a `livecheck` regex matching only `desktop-v*`, so it cannot see 1.0.0 and installs a three-versions-old build. **That needs a separate bump in the tap** — out of scope here, noted in #461.

https://claude.ai/code/session_01FM9Ei7hExU7x1amTz5Zdud


vibexp-agent:
We need to properly plan and review #461 first before a PR is opened for it.

shahariaazam:
Thank you @amims71 for the PR and reporting the issue.
This we have fixed in https://github.com/shaharia-lab/agento/pull/466 this PR

PR #464, opened 2026-08-25
feat(sessions): filter the list by Claude account (#463)

### LINKED ISSUE #463: Sessions list has no account filter on desktop — the web build had one
The Sessions list cannot be narrowed to one Claude account. On a machine set up
with more than one Claude config dir, every account's sessions are interleaved
in one list with no way to separate them, and no indication in the list which
account a row came from.

The Go/web build had this filter. `SessionFilters.configDir` sent `config_dir`
on both the list and the facets request, and the toolbar offered an **All
accounts** dropdown — deliberately rendered only when the corpus actually spanned
more than one account, so a single-account machine saw no extra control.

Reading a session is possible today, filtering by it is not: `SessionsView`'s
inspector already shows a **Config** row per session, so the account is visible
one session at a time but cannot be used to narrow the list.

## Why it matters

Indexing every config dir was a deliberate decision — analytics is
retrospective, and a machine running two accounts wants both in every total.
Multi-account users are therefore the normal case for this list, not an edge
case, and they are the ones who cannot use it. Separating work from personal is
the whole reason someone runs two accounts in the first place.

Settings → Data & Analytics can hide a *project*, but there is no equivalent for
an account, and hiding is the wrong tool anyway: the question is "show me only
this account for a moment", not "exclude this account from reporting".

## Where the gap is

The desktop backend already supports it. `SessionQuery.config_dir` is parsed
from the query string, `build_filter` applies it as `c.config_dir = ?`, and
`SessionFacets.config_dirs` publishes the distinct dirs for the dropdown to
offer. `src/lib/types.ts` already declares both `config_dir` and `config_dirs`.

Nothing in the desktop UI sends the parameter or renders the control:
`SessionsView`'s `filterParams` carries only `q`, `project` and `favorites`.

## Reproduce

Configure a second Claude config dir (Settings → Claude config dirs), let the
scan run, then open Sessions. Both accounts' sessions appear in one list with no
account control in the toolbar.

Confirmed on 1.0.0 with two indexed accounts:

```console
$ sqlite3 -readonly ~/.agento/agento.db \
    "SELECT config_dir, COUNT(*) FROM claude_session_cache GROUP BY config_dir"
/Users/shan/.claude|22
/Users/shan/.claude-personal|57
```

Both dirs are indexed and correctly attributed on the rows; the list simply
offers no way to pick between them.

amims71:
Fixes #463.

## What was missing

The web build could narrow the sessions list to one Claude account; the desktop port never offered the control. With two config dirs indexed, every account's sessions were interleaved in one list, and the account was readable only one session at a time in the inspector.

**Frontend only — the backend already had every piece.** `SessionQuery.config_dir` is parsed from the query string, `build_filter` applies it as `c.config_dir = ?`, `SessionFacets.config_dirs` publishes the distinct dirs, and `src/lib/types.ts` already declared both. `SessionsView` simply never sent the parameter or drew the control. So this is +63/−2 in one file, with no migration, no endpoint and no type change.

## Behaviour

Rendered only when the facet reports more than one account, as the web build did — one account is the common case, and a control whose every state shows the same rows is noise. Labels go through the existing `tildePath` (`/Users/shan/.claude-personal` → `~/.claude-personal`), but the *value* stays the absolute path the scan recorded, because the server matches it exactly.

Two details worth reviewing:

- The facet is computed over every *visible* session rather than the filtered set, so choosing one account never removes the others from the dropdown — an option that vanishes when picked cannot be un-picked. That is the server's existing behaviour; this just relies on it.
- A live selection keeps the control on screen even when the facet stops offering it, which happens if the dir is un-indexed in Settings while it is chosen. Without that the control disappears exactly when it is the only thing that could clear itself, pinning the list to an account with no rows. (`filtersActive` also includes it, so the empty state's **Clear filters** is a second way out.)

## Verification

There is no Rust toolchain on this machine, so `tauri dev` was not available. Rather than stopping at a typecheck — which `local-verify` explicitly says proves nothing about rendering — the frontend was exercised in Chrome through the sanctioned Vite path, against a stub API on `:8991` (the port `vite.config.ts` already proxies to) carrying this machine's real shape: 22 sessions under `~/.claude`, 57 under `~/.claude-personal`.

| Step | Result |
|---|---|
| Sessions toolbar, two accounts in the facet | **All accounts** control renders, after **All projects** |
| Open it | `All accounts` · `~/.claude` · `~/.claude-personal` |
| Pick `~/.claude-personal` | Summary goes `79 sessions · 118.5K · $33.18` → `57 sessions · 85.5K · $23.94`; every row personal |
| The wire | `GET /api/claude-sessions?config_dir=%2FUsers%2Fshan%2F.claude-personal&sort=recent&limit=50` and the matching `/facets` |
| Back to **All accounts** | Parameter dropped entirely — `GET /api/claude-sessions?sort=recent&limit=50`, not an empty `config_dir=` |
| Console | No errors, no React warnings |

The filter semantics were confirmed independently against the live database, so the stub is not the only evidence the parameter does what the UI implies:

```console
$ sqlite3 -readonly ~/.agento/agento.db \
    "SELECT DISTINCT config_dir FROM claude_session_cache ORDER BY config_dir"
/Users/shan/.claude
/Users/shan/.claude-personal

$ # what `c.config_dir = ?` returns per account
work|22   personal|57   all|79
```

`npm run build` (tsc --noEmit + vite build) passes; `pre-commit run` passes.

One case not exercised end to end: the un-indexed-while-selected escape hatch, which needs the facet to change under a live selection. It is a union of the facet dirs with the current value and a second render clause — reviewed by inspection rather than clicked.

https://claude.ai/code/session_01FM9Ei7hExU7x1amTz5Zdud


vibexp-agent (approved):
Thanks for the thorough PR — verified the facet-scoping, cursor-reset and wire-format claims against the backend; all correct.

PR #475, opened 2026-08-25
fix(ui): make content selectable by default, chrome not (#469)

### LINKED ISSUE #469: Text is unselectable outside a few opted-in views — invert the user-select default
## What

Text is not selectable anywhere in the app except a handful of opted-in places. A user cannot select and copy a session transcript, a value in the Gateway panes, or most of the right-hand detail sidebars.

## Why

`src/styles/base.css:34` sets `user-select: none` on `body`, deliberately — the comment reads *"Chrome-like text selection is a web tell. Opt in per element instead."* Opt-in happens via `.selectable` and a few per-view rules (`chats.css`, `settings.css`, `integrations.css`, `views.css`), so every view added since has shipped unselectable by default. The sessions list/detail and the whole `views/gateway/` section were never opted in.

The intent — the app should not feel like a web page — is right; the mechanism is backwards. Native desktop apps (Mail, Xcode, VS Code) allow selection everywhere in *content* and suppress it only on *chrome* (titlebar, sidebar nav, toolbars, status bar, tabs). An allowlist of content is unbounded and drifts; a denylist of chrome is small and stable.

## How

1. Invert the default in `src/styles/base.css`: `body` becomes `user-select: text` (keep the `-webkit-` prefix — WebKitGTK and WKWebView both need it).
2. Scope `user-select: none` to chrome only: `.titlebar` / `[data-tauri-drag-region]`, the sidebar (nav rows and section labels), `.statusbar`, `.srcitem`, `button`, and tab strips. Remove the now-redundant per-view `user-select: text` opt-ins (`.selectable` can stay as a no-op alias or be deleted with its call sites).
3. Cursor follows the same split: `cursor: text` on prose and value cells, `cursor: default` on chrome.
4. Give `::selection` a focus-aware colour, matching the existing focus-aware row selection (accent when the window is focused, grey when not) — that is what keeps it reading as desktop rather than web.
5. Add `CopyButton` to single-value rows where drag-selecting is the wrong affordance regardless: gateway tokens and base URLs, session/chat ids, project paths.
6. Check `App.tsx`'s keydown handler does not swallow Cmd/Ctrl+C now that there is a selection to copy.

## Acceptance criteria

- Selecting and copying works in the sessions list and detail, all Gateway views, and every detail sidebar.
- Dragging the window by the titlebar still works on all three platforms, and clicking a sidebar row or a button does not start a text selection.
- Verified in a **release** build (`npm run app:build`), not only `npm run app` — drag-region behaviour and webview selection semantics are only real there.

---
## WHAT

**Summary:** Invert the global `user-select` default so app *content* is selectable and only *chrome* is not, instead of the current allowlist that every new view forgets to join.

**Scope:**
- Flip `body { user-select: none }` → `text` in `src/styles/base.css`.
- Add an explicit chrome denylist (titlebar/drag region, sidebar, status bar, toolbar, table headers, buttons).
- Add a focus-aware `::selection` colour (none exists today — `grep -rn "::selection" src/` returns nothing, so selected text currently renders in the WebKit default blue on both platforms, in both themes).
- Retire the `.selectable` opt-in and its ~30 call sites.
- Add `CopyButton` to single-value rows where drag-selecting is the wrong affordance regardless.

**Affected Components:**
- `src/styles/base.css:33-50` — the `user-select: none` default and the `.selectable` opt-in rule; new `::selection`.
- `src/styles/tokens.css:107-109,166-168,216-218` — reuse `--bg-select` / `--bg-select-focus` / `--fg-select-focus`; if a distinct text-selection colour is wanted, declare it in **all three** places (bare `:root`, the `prefers-color-scheme` block *and* `:root[data-theme="dark"]`) per the theming rule in `CLAUDE.md`.
- `src/styles/shell.css:29,65,94,104,179,267` — `.titlebar`, `.sidebar`, `.sidebar__caption`, `.srcitem`, `.toolbar`, `.statusbar` become the denylist (the last two of these already carry `none`).
- `src/styles/views.css:392` — `.th` keeps `none`; `views.css:235,287`, `chats.css:117`, `settings.css:342`, `integrations.css:126` lose their now-redundant `text`.
- `src/views/sessions/SessionDetail.tsx`, `src/views/gateway/{ProvidersView,ModelsView,UsageView}.tsx`, `src/views/analytics/*` — the views with **zero** opt-ins today; verified by sweeping for files with no `selectable` token.
- ~13 `.tsx` files carrying `className="… selectable"` — drop the class.

**Acceptance Criteria:**
- [ ] Text can be selected and copied in: the session transcript's tool-call bodies and metadata (not just `.msg__text`, which already opts in via `views.css:235`), the session detail sidebar's token/cost rows, and every Gateway view including Providers, Models and Usage.
- [ ] `grep -rn "user-select: text" src/styles/` returns only the denylist file's counterpart rules — i.e. no view carries a per-view opt-in any more, and `.selectable` has no remaining call sites.
- [ ] Selected text uses `--bg-select-focus`/`--fg-select-focus` while the window is focused and `--bg-select` when it is not, matching the existing `.window--focused` convention (`shell.css:131`, `views.css:52`), in both light and dark.
- [ ] Dragging the window by the titlebar strip still works on macOS, Linux and Windows, and dragging across a sidebar row or a table header starts no text selection.
- [ ] Clicking and dragging a session row still selects the row rather than starting a selection sweep across the table.
- [ ] `CopyButton` is present on the gateway base URLs and minted token, the session id and project path, and the JWKS `kid` in Settings → Security.
- [ ] Verified in a **release** build (`npm run app:build`), not only `npm run app`.

**Out of Scope:**
- Right-click / native context menus (there are none — see the "No multi-window, tray icon, native context menus" gap in `CLAUDE.md`).
- Selecting across pane boundaries, or any "Copy as Markdown"/"Export transcript" feature.
- Changing what `Cmd/Ctrl+C` does: `App.tsx`'s keydown handler already ignores `c` (`src/App.tsx:185-228`), and the macOS Edit menu already carries `.copy()` (`src-tauri/src/menu.rs:52`), so copy works the moment a selection can exist.

## WHY

**Business Value:** Copying a session id, an error string, a token count or a gateway base URL out of the app is a routine, daily action. Today it silently does nothing, and there is no affordance explaining why — the user's report was "we can't select any text to copy", i.e. it reads as the app being broken rather than as a deliberate style.

**Problem Statement:** `src/styles/base.css:34` sets `user-select: none` on `body`, with the comment *"Chrome-like text selection is a web tell. Opt in per element instead."* Opting in happens through a `.selectable` class and five per-view CSS rules. The intent is right — a desktop window is not a document — but the mechanism is an **allowlist over content**, which is unbounded and drifts: every view added since has shipped unselectable, and today `src/views/sessions/SessionDetail.tsx`, `src/views/gateway/{Providers,Models,Usage}View.tsx` and all of `src/views/analytics/` contain not one opt-in. Chrome, by contrast, is a small and stable set. Native desktop apps (Mail, Xcode, VS Code, Finder) suppress selection on chrome only, which is the same rule expressed the maintainable way round.

**Expected Impact:** UX — selection and copy work everywhere content is rendered, in ~19 component/view files that have none today, and stay working in every view added later without anyone remembering a class. Debt — deletes an opt-in class plus ~30 call sites and five per-view CSS rules; net LOC should be negative. Risk — low and contained to CSS; the two regressions worth guarding are window drag and table-row drag, both covered by acceptance criteria.

**Context:** No related issue. One decision was taken without a round trip and should be revisited if it feels wrong in review: **interactive list rows keep `user-select: none` on the row itself while their text cells stay selectable**, because a click-drag on a row means "select this row", not "sweep text across the table" — which is Finder's behaviour and matches what `SessionsView.tsx:953-960` already does by hand. The inversion should preserve that, not flatten it.

## HOW

**Technical Approach:** One CSS inversion plus a mechanical sweep. `base.css` stops declaring `none` on `body` and instead declares `text` there and `none` on a named chrome selector list; the per-view `user-select: text` rules and the `.selectable` class become dead and are removed. `::selection` is declared once in `base.css` against the existing `--bg-select*` tokens, following the `.window--focused` pattern the row highlight already uses. No JS or Rust change.

**Implementation Details:**
- `src/styles/base.css` — replace lines 33-35 with `user-select: text` (keep `-webkit-user-select`, both webviews need it) and leave `cursor: default` on `body`. Replace the `.selectable, input, textarea, [contenteditable]` rule at 42-50 with a chrome rule: `.titlebar, [data-tauri-drag-region], .sidebar, .statusbar, .toolbar, .th, button, select { user-select: none; -webkit-user-select: none; }`. Add `cursor: text` for prose/value surfaces and `::selection` / `.window--focused ::selection`.
- `src/styles/shell.css` — the existing `user-select: none` at 101 (`.sidebar__caption`) and 278 (`.statusbar`) become redundant once the denylist covers their ancestors; drop them or keep them as belt-and-braces, but not both spellings in one file.
- `src/styles/{views,chats,settings,integrations}.css` — delete the five `user-select: text` declarations (`views.css:235,287`, `chats.css:117`, `settings.css:342`, `integrations.css:126`). Keep `views.css:392` (`.th`).
- The `.tsx` sweep — remove `selectable` from every `className` (13 files: `JobsView`, `ChatsView`, `AboutView`, `AgentsView`, `SessionsView`, `SettingsView`, `gateway/OverviewView`, `gateway/SettingsView`, `settings/LogsPane`, `settings/SecurityPane`, and any added since). `AgentsView.tsx:963`'s comment uses the word "selectable" in prose — do not touch it.
- `CopyButton` (`src/components/CopyButton.tsx`, over `lib/clipboard.ts`, which already carries the `execCommand` fallback WebKitGTK needs) — add to `gateway/OverviewView.tsx`'s base URLs, `sessions/SessionDetail.tsx`'s session id and project path, and `settings/SecurityPane.tsx`'s `kid`.

**Testing Strategy:** There is no TypeScript test harness in this repo, so this is verified by hand, in a release build, against the acceptance criteria — `npm run app:build`, then walk Sessions (list, detail sidebar, transcript incl. a tool call), all five Gateway views, Analytics, and Settings → Security. Check the three drag cases (titlebar on each platform, sidebar row, table header) and the row-drag case. `npm run build` must stay green (typecheck only). Guard the CSS side with a grep in review rather than a test: no `user-select: text` outside `base.css`, no `selectable` call site left.

**Documentation / Configuration:** No new deps or env vars. `CLAUDE.md`'s "Desktop, not web" paragraph describes the UI conventions and should gain one sentence recording the inverted rule and why, so the next view is not written against the old one. `src/styles/base.css`'s header comment ("text is not selectable by default") must be rewritten in the same change — leaving it is how the next person re-adds the allowlist.

**Files to Create / Modify:**
- Modify: `src/styles/base.css`, `src/styles/shell.css`, `src/styles/views.css`, `src/styles/chats.css`, `src/styles/settings.css`, `src/styles/integrations.css`, optionally `src/styles/tokens.css`.
- Modify: the 13 `.tsx` files carrying `className="… selectable"`, plus `sessions/SessionDetail.tsx`, `gateway/OverviewView.tsx` and `settings/SecurityPane.tsx` for `CopyButton`.
- Modify: `CLAUDE.md`.
- Create: none.

**Risks:**
- *Window drag breaks* — `data-tauri-drag-region` is declared once as `deep` on the outer titlebar strip (`TitleBar.tsx`); if the strip becomes selectable, a drag can start a selection instead. Mitigation: the denylist names both `.titlebar` and `[data-tauri-drag-region]`, and the drag is checked in a **release** build, since dev-vs-release differences in this shell are a documented trap.
- *Table drag-select* — see the Context decision above; mitigated by keeping `none` on the row and `text` on its cells.
- *Theme regression* — a `::selection` colour declared only inside `@media (prefers-color-scheme: dark)` breaks the explicit `data-theme` switch. Mitigated by the three-place token rule already recorded in `CLAUDE.md`.
- *Scope creep into a context menu* — explicitly out of scope.

**Complexity:** Medium (multi-file, mostly CSS plus a mechanical class sweep, well under 500 LOC, net negative)

---
**Refined by:** AI Issue Refiner · **Date:** 2026-08-25 · **Status:** Ready

vibexp-agent:
Closes #469

## What

`src/styles/base.css` declared `user-select: none` on `body` and let content opt
back in via a `.selectable` class plus five per-view rules. This inverts it:
content is selectable, and one **denylist** in `base.css` names the chrome.

## Why

An allowlist over content is unbounded and drifts. Every view added since shipped
unselectable — the whole `views/gateway/` section, `views/analytics/` and the
session detail carried no opt-in at all, so a user could not select a session id,
a token count, a gateway base URL or an error string. Chrome is a small, stable
set, which is the split native apps draw.

## How

- **`base.css`** — `body` becomes `user-select: text` (both spellings; WebKitGTK
  and WKWebView need the prefix). One denylist rule: `.titlebar`,
  `[data-tauri-drag-region]`, `.sidebar`, `.toolbar`, `.statusbar`, `.srcitem`,
  `.listrow`, `.table thead th`, `.table tbody tr`, `button`, `select`. Form
  fields re-declare `text` so a search field inside `.toolbar` keeps its own.
  `cursor: text` on the three content panes, `default` on the denylist and on `a`.
- **`::selection`** — focus-aware over the existing `--bg-select` /
  `--bg-select-focus` / `--fg-select-focus` tokens, matching `.window--focused`.
  No new token, so the three-place theming rule needed nothing.
- **The sweep** — `.selectable` and its 30-odd call sites across 14 files are
  gone, as are the five per-view `user-select: text` rules and the two now-
  redundant `none` ones (`shell.css`, `views.css`). One spelling of the rule.
- **`CopyButton`** on the five single-value rows where the text is abbreviated to
  fit the pane and drag-selecting would copy the ellipsis: the three gateway
  endpoint URLs, the session id, the project path. The session inspector's old
  "Copy session ID" action button goes with them — it duplicated the new row
  affordance and wrote through `navigator.clipboard` directly, bypassing the
  `execCommand` fallback `lib/clipboard.ts` carries for WebKitGTK.
- **`CLAUDE.md`** records the inverted rule under "Desktop, not web", so the next
  view is not written against the old one.

## Deviations from the issue's HOW

- The HOW said *"keep `views.css:392` (`.th`)"* while also naming `.th` in the
  base denylist. Those are two spellings of one rule in two files, which is the
  drift this issue is about, so `.table thead th` lives in the denylist only.
  There is no `.th` class in the tree — line 392 was `.table thead th`.
- The Context note asked that rows keep `none` **while their cells stay
  selectable**. Those cannot both hold: `td { user-select: text }` makes a drag
  starting in a cell sweep the table, and a double-click — which is what *opens*
  a session — leaves a word selected behind the view it opened. Both are explicit
  acceptance criteria, so the row is `none` and the cells inherit it. The
  Context's intent (row drag = row selection) is preserved.
- Cursor is split at pane granularity (`.pane-list` / `.pane-detail` /
  `.pane-inspector`) rather than per prose element, so the cosmetic half does not
  reintroduce an allowlist. Clickables in this tree are `<button>`, which the
  denylist already claims.

## Acceptance criteria

- [x] Selectable in the sessions list and detail, all Gateway views, every detail
      sidebar, transcript tool-call bodies.
- [x] `grep -rn "user-select" src/styles/` returns `base.css` only;
      `grep -rn "selectable" src/` returns two unrelated prose comments.
- [x] `::selection` focus-aware in both themes over the existing tokens.
- [x] Titlebar drag and row drag unchanged — both in the denylist.
- [x] `CopyButton` on the gateway base URLs, session id, project path; Settings →
      Security's `kid` already had one.
- [x] `Cmd/Ctrl+C` unaffected — `App.tsx`'s handler has no `c` arm and never
      `preventDefault`s one (verified at `src/App.tsx:187-226`).
- [x] Verified in a **release** build of the shell (see below).

## Verification

No TypeScript test harness exists, so this is `npm run build` (typecheck) plus a
hand walk-through of the built release binary in the real WebKitGTK webview at
the release origin.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -472,6 +472,22 @@ status bar, focus-aware selection (accent when the window is focused, grey
 when not), ⌘K palette, no browser affordances. Reuse the existing CSS classes;
 new CSS goes in a per-view file imported by that view.
 
+**Text selection is a denylist over chrome, not an allowlist over content**
+(#469). `base.css` declares `user-select: text` on `body` and `none` on one
+named list — the titlebar and every drag region, the sidebar, the toolbar, the
+status bar, source-list and table rows, table headers, `button` and `select` —
+so a view added later is selectable without anyone remembering a class. It was
+the other way round until #469, and the allowlist drifted exactly as one does:
+the whole `views/gateway/` section, `views/analytics/` and the session detail
+shipped with no opt-in at all, so a user could not copy a session id or an
+error string. **Do not re-introduce a `.selectable` class or a per-view
+`user-select: text`** — the denylist is the whole rule and it lives in
+`base.css`; a second spelling elsewhere is how it comes apart. Rows stay in the
+denylist deliberately: a table row's double-click *opens* the session, and a
+drag across it means "select this row", so selectable cells would leave a word
+highlighted behind the view they opened. `::selection` is focus-aware over the
+existing `--bg-select*` tokens, matching `.window--focused`.
+
 **The window's origin has to be in the capability's scope, or every IPC
 command is denied — silently.** This is the single most expensive thing to
 re-discover in this shell, because it fails *only in release builds* and it
--- a/src/components/TokenReveal.tsx
+++ b/src/components/TokenReveal.tsx
@@ -47,7 +47,7 @@ export function TokenReveal({
         </button>
       </div>
       <div className="secnew__token">
-        <code className="mono selectable">{token}</code>
+        <code className="mono">{token}</code>
         <CopyButton
           text={token}
           title="Copy token"
--- a/src/styles/base.css
+++ b/src/styles/base.css
@@ -1,7 +1,9 @@
 /* ============================================================================
    Base — desktop application shell resets.
-   A desktop window is not a document: it never scrolls as a whole, text is not
-   selectable by default, and there is no native browser affordance anywhere.
+   A desktop window is not a document: it never scrolls as a whole, and there is
+   no native browser affordance anywhere. Content *is* selectable; the window's
+   chrome is not, and the denylist that says which is which lives below — one
+   list, in this file. Do not re-add a per-element opt-in (#469).
    ========================================================================== */
 
 *,
@@ -30,23 +32,72 @@ body {
   text-rendering: optimizeLegibility;
   font-synthesis: none;
 
-  /* Chrome-like text selection is a web tell. Opt in per element instead. */
-  user-select: none;
-  -webkit-user-select: none;
+  /* Content is selectable; the chrome below opts out. The inverse — `none`
+     here plus a `.selectable` opt-in — was the rule until #469, and it drifted
+     exactly as an allowlist over content does: content is unbounded, so every
+     view added since shipped unselectable and a user could not copy a session
+     id or an error string. Chrome is a small, stable set, which is the split
+     native apps draw. Both spellings: WebKitGTK and WKWebView need the prefix. */
+  user-select: text;
+  -webkit-user-select: text;
 
   /* No rubber-band / pull-to-refresh inside an app window. */
   overscroll-behavior: none;
   cursor: default;
 }
 
-/* Content the user genuinely reads or copies opts back in. */
-.selectable,
+/* The chrome denylist — the whole of it, so there is one place to read.
+   These behave as controls, not as text: dragging them must move the window,
+   activate a row or sort a column, and a double-click on a row must open it
+   rather than leave a word selected behind the view it opened. */
+.titlebar,
+[data-tauri-drag-region],
+.sidebar,
+.toolbar,
+.statusbar,
+.srcitem,
+.listrow,
+.table thead th,
+.table tbody tr,
+button,
+select {
+  user-select: none;
+  -webkit-user-select: none;
+  cursor: default;
+}
+
+/* Form fields are content wherever they sit — a search field inside `.toolbar`
+   would otherwise inherit the denylist's `none` and refuse its own text. A
+   direct declaration always beats an inherited one, so order does not matter. */
 input,
 textarea,
 [contenteditable="true"] {
   user-select: text;
   -webkit-user-select: text;
   cursor: auto;
+}
+
+/* Cursor follows the same split, one level coarser: the three content panes
+   read as text, and the denylist above has already claimed the chrome inside
+   them. This half is cosmetic — the selection rule is the denylist. */
+.pane-list,
+.pane-detail,
+.pane-inspector {
+  cursor: text;
+}
+
+/* Selected text follows the window's focus, exactly as a selected row does
+   (`.window--focused`, shell.css / views.css) — accent while the window is key,
+   grey when it is not. That focus-awareness is what keeps a selection reading
+   as desktop rather than as a web page. The colours are the existing selection
+   tokens, so they are already declared in all three theme places. */
+::selection {
+  background: var(--bg-select);
+  color: var(--fg);
+}
+.window--focused ::selection {
+  background: var(--bg-select-focus);
+  color: var(--fg-select-focus);
 }
 
 button,
@@ -60,13 +111,12 @@ select {
   outline: none;
 }
 
-button {
-  cursor: default;
-}
-
 a {
   color: var(--accent);
   text-decoration: none;
+  /* A link is a control here too — the panes above read as text, and an I-beam
+     over something you click is the web tell this shell avoids. */
+  cursor: default;
 }
 
 /* Native focus: keyboard only. Mouse clicks must not leave a ring behind. */
--- a/src/styles/chats.css
+++ b/src/styles/chats.css
@@ -114,7 +114,6 @@
 .prompt__text {
   font-size: var(--text-base);
   line-height: var(--leading-normal);
-  user-select: text;
 }
 
 .prompt__opts {
--- a/src/styles/integrations.css
+++ b/src/styles/integrations.css
@@ -123,8 +123,6 @@
   max-height: 90px;
   overflow: auto;
   word-break: break-all;
-  user-select: text;
-  cursor: text;
 }
 
 /* --- Trigger rules ------------------------------------------------------- */
--- a/src/styles/settings.css
+++ b/src/styles/settings.css
@@ -339,6 +339,4 @@
   max-height: 92px;
   overflow: auto;
   word-break: break-all;
-  user-select: text;
-  cursor: text;
 }
--- a/src/styles/shell.css
+++ b/src/styles/shell.css
@@ -98,7 +98,6 @@
   text-transform: uppercase;
   color: var(--fg-tertiary);
   padding: var(--sp-6) var(--sp-3) var(--sp-2);
-  user-select: none;
 }
 
 .srcitem {
@@ -275,7 +274,6 @@
   border-top: var(--hairline) solid var(--line);
   font-size: var(--text-sm);
   color: var(--fg-tertiary);
-  user-select: none;
 }
 
 .statusbar__item {
--- a/src/styles/views.css
+++ b/src/styles/views.css
@@ -667,8 +667,6 @@
    abbreviated to fit the pane, so drag-selecting it would copy the ellipsis
    rather than the value. The button carries the real one (#469). */
 .insp-row__copy {
-  display: flex;
-  align-items: center;
   gap: var(--sp-3);
   min-width: 0;
 }
--- a/src/styles/views.css
+++ b/src/styles/views.css
@@ -232,7 +232,6 @@
   font-size: var(--text-md);
   line-height: var(--leading-relaxed);
   color: var(--fg);
-  user-select: text;
 }
 /* Paragraph spacing and inline code live in markdown.css, with the rest of the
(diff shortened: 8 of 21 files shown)


vibexp-agent:
Two things the review round surfaced that are behaviour rather than defects, recorded so they are not re-discovered:

- **`.toolcall__head` stays unselectable.** A tool call's header strip is a `<button>` (`views/chat/Transcript.tsx:274`) — it is the disclosure toggle — so the denylist claims it, which is what the issue's own HOW asks for by naming `button`. The tool-call *body*, the message text and the session detail's metadata row are all selectable. Making the header selectable would break "clicking a button does not start a text selection".
- **There is no automated regression guard, and there cannot be a cheap one.** This repo has no TypeScript test harness, and the type-level pin idiom (`lib/typeAssert.ts`) can assert a *value*, not a computed style. Reverting this change fails nothing. Verification was by hand against a **release** build driven through WebKit's remote inspector at the release origin (`http://127.0.0.1:46115`, i.e. the *remote*-classified window, not `npm run app`): every denylist selector computes `user-select: none`, `body`/panes/inputs compute `text`, ranges over content in Sessions, the session inspector, Gateway Providers/Models/Usage, Token Usage and Insights all return their text, and both `::selection` rules are present in the shipped bundle with `--bg-select` `#d8dade` / `--bg-select-focus` `#2f6feb` resolving under `.window--focused`.

Issue #486, opened 2026-08-26
Sessions: right-click actions on a row, and move the inspector actions above the fold


vibexp-agent:
## Problem

The session actions — **View session**, **Add to favourites**, **Continue in chat** — live in an `Actions` group at the very bottom of the Sessions inspector (`src/views/SessionsView.tsx`, the `Inspector` component). On a session with a long metadata block they are below the fold, so users scroll past them or never find them. There is also no way to act on a row without first selecting it and then reaching for the sidebar.

## Asks

**1. Right-click context menu on a session row.** Offer the same actions from the row itself: View session, Add/Remove favourite, Continue in chat — plus Copy session id and Copy project path, which are asked for constantly and have no affordance today.

`src/main.tsx` already suppresses the native context menu everywhere except editable fields and selected text, so an in-app React menu drops in without fighting the webview. This is an in-app menu, not the native macOS one listed under "Not implemented" in `CLAUDE.md`.

**2. Move the actions to the top of the inspector.** Rather than relocating the whole `Actions` group as-is, a compact icon-button row pinned directly under the session title in the inspector header keeps the metadata groups where they are and makes the actions unmissable.

## Notes

- Both halves share one set of action handlers (`toggleFavorite`, `continueInChat`, `setOpenId`), so they should land together.
- Rows are in `base.css`'s `user-select: none` denylist deliberately (a drag across a row means "select this row"), which the context menu does not change.
- Related: #485 fixes what **Continue in chat** actually does when invoked; this issue is only about where it can be invoked from.

---
## WHAT

**Summary:** Make the session actions reachable without hunting — a right-click menu on a session row, and an action strip pinned at the top of the inspector instead of an `Actions` group at the bottom of a scrolling pane.

**Scope:**
- A `ContextMenu` primitive in `src/components/ui.tsx`, opened at the cursor.
- Right-click on a session row: **View session**, **Add / Remove favourite**, **Continue in chat**, **Copy session id**, **Copy project path**.
- Right-clicking a row selects it first, so the inspector agrees with the menu.
- A compact icon-button action strip under the inspector header, outside the scrolling area, replacing the bottom `Actions` group.

**Affected Components:**
- `src/components/ui.tsx` — gains `ContextMenu`, modelled on the existing `Dropdown` (line ~131) which already solves portal positioning, viewport flipping, outside-click and keyboard dismissal.
[...]

PR #487, opened 2026-08-26
fix: land the user in the chat "Continue in chat" creates (#485)

### LINKED ISSUE #485: Continue in chat does nothing visible — the continued chat is created but never opened
## What happens

Clicking **Continue in chat** in the Sessions inspector appears to do nothing. No view change, no composer, no visible confirmation.

## What actually happens (reproduced)

The backend is healthy. Against a dev build on a real corpus:

```
POST /api/claude-sessions/e04d111f-.../continue  ->  201 {"chat_id":"56531d55-..."}
GET  /api/chats  ->  top row: "New Chat", sdk_session_id=e04d111f-..., cwd + model inherited
```

The chat row is created correctly, and `native/chat/runner.rs:340` does pass `sdk_session_id` to the CLI as `--resume`, so the conversation would genuinely resume on the first message.

Everything missing is on the frontend. `src/views/SessionsView.tsx:512` (`continueInChat`) resolves the POST and then only sets `continuedChat`, which renders as a small grey `sess-note` (`Continued as chat <id> — open Chats to resume it`) appended below the button, at the bottom of a scrolling inspector. There is no navigation, no chat selection, and no composer — so from the user's seat nothing happened. This is the gap `CLAUDE.md` records under Status: *"Continue in chat on a session can only report the new chat_id; it cannot switch to the Chats view. Needs a nav context in App.tsx."*

Three secondary defects found alongside it:

- The continued chat lands in the Chats list titled **New Chat** with an empty transcript, indistinguishable from any other new chat — nothing marks it as a resumed session.
- There is **no dedupe**: every click mints another chat row for the same session.
- `src/views/sessions/SessionDetail.tsx` (the full session view) has **no continue action at all**; only the list inspector does.

## Expected

Continuing a session takes the user to that chat, ready to type, with the conversation identifiable as a continuation of the source session. This worked in the previous Go-backed version.

## Scope

- A nav mechanism in `App.tsx` (view id + optional target id) so a view can hand off to another; `ChatsView` needs to accept a chat to preselect (it currently owns `selected` internally with no prop).
- Title or badge the continued chat from the source session rather than leaving it `New Chat`.
- Guard repeat clicks.
- Add the action to `SessionDetail.tsx`.

## Notes

- Error feedback shares the same off-screen spot, so a 404/500 is visually identical to success — worth fixing with the same change.
- `POST /api/claude-sessions/{id}/continue` needs no backend change.

---
## WHAT

**Summary:** Make **Continue in chat** actually land the user in the resumed conversation, instead of silently creating a chat nobody is shown.

**Scope:**
- A navigation intent in `App.tsx` (a view id plus an optional target id) that a deep child view can fire, so Sessions can hand off to Chats.
- `ChatsView` accepts a chat to preselect, mirroring the existing `newChatNonce` prop.
- `continueInChat` navigates on success instead of setting a note string.
- Failure is reported where the user is looking, not in the same off-screen spot as success.
- The continued chat is titled from the source session rather than left as `New Chat`.
- The same action is added to the full session view (`SessionDetail.tsx`), which has none today.

**Affected Components:**
- `src/App.tsx` — holds `view`; gains a nav target and a provider around the view switch.
- `src/lib/nav.ts` — `ViewId` lives here; gains the nav context and its `useNavigate()` hook.
- `src/views/SessionsView.tsx` — `continueInChat` (line ~512) and the `Inspector` `Actions` group (line ~1206).
- `src/views/ChatsView.tsx` — `selected` is internal today (line ~67); accepts an incoming chat id.
- `src/views/sessions/SessionDetail.tsx` — gains the action.
- `src-tauri/src/native/sessions/continue_chat.rs` — the title written for the new chat.

**Acceptance Criteria:**
- [ ] Clicking **Continue in chat** switches to the Chats view with the new chat selected and its composer ready to type — no manual navigation, no scrolling to find a note.
- [ ] The continued chat is titled from the source session (its `display_title`), not `New Chat`, so it is identifiable in the Chats list.
- [ ] Sending the first message in that chat resumes the Claude conversation — the CLI is invoked with `--resume <source session id>` and the reply shows prior context.
- [ ] A failed continue (404 `session not found`, 500) shows an error the user can see without scrolling, and does not navigate.
- [ ] The same action exists in the full session view (`SessionDetail.tsx`) and behaves identically.
- [ ] Continuing the same session twice in one visit does not leave two stray chats behind from accidental double-clicks (the button stays disabled while in flight, and success navigates away).

**Out of Scope:**
- Replaying the source transcript into the chat view. The continued chat legitimately has no messages of its own — the history lives in the CLI session — so this issue only makes the chat *identifiable*, not populated.
- The right-click context menu and moving the inspector actions above the fold — that is #486.
- Backend idempotency (returning an existing chat for a session already continued). Deliberately not done; see **Context**.

## WHY

**Business Value:** Continuing a Claude Code session in Agento is one of the app's headline reasons to exist — it is the bridge from "I have 1,200 past sessions" to "I can carry one on". Today that bridge appears broken, so the feature is effectively unavailable.

**Problem Statement:** The button works and looks like it does nothing. Reproduced against a dev build on a real corpus:

```
POST /api/claude-sessions/e04d111f-.../continue  ->  201 {"chat_id":"56531d55-..."}
GET  /api/chats  ->  top row: "New Chat", sdk_session_id=e04d111f-..., cwd + model inherited
```

The row is created correctly and `native/chat/runner.rs:340` does pass `sdk_session_id` as `--resume`, so the conversation *would* resume on the first message. But `SessionsView.tsx:512` resolves the POST and only sets `continuedChat`, rendered as a small grey `.sess-note` (`Continued as chat <id> — open Chats to resume it`) appended below the button, at the bottom of a scrolling inspector. No view change, no selection, no composer. The error note renders in the same place, so a 404 is visually identical to a success.

**Expected Impact:** A feature that currently reads as broken becomes a one-click path. It also removes a class of invisible failure: today a user cannot tell a refused continue from a successful one. Secondary: no more accumulation of untitled `New Chat` rows from repeat clicks on a button that appeared not to work.

**Context:**
- `CLAUDE.md` records this under Status as a known gap: *"Cross-view navigation: 'Continue in chat' on a session can only report the new `chat_id`; it cannot switch to the Chats view. Needs a nav context in `App.tsx`."* The approach below follows that stated direction rather than prop-drilling a callback.
- The backend needs no route change: `POST /api/claude-sessions/{id}/continue` answers `201 {"chat_id"}` correctly, and `sessions/continue_chat.rs` already writes both statements in one transaction. The only backend edit is the title.
- **Judgement call (non-interactive run, flagged):** repeat-click protection is handled by navigating away on success plus the existing in-flight `disabled`, *not* by making the endpoint idempotent. Reusing an existing chat for the same `sdk_session_id` would mean two chats claiming one CLI session, and after the first turn `persist.rs` rewrites `sdk_session_id` from the stream — so branching a session twice is a legitimate thing a user may want. If the maintainer prefers idempotency, say so and it can be folded in.
- Related: #479 is in flight against `src/views/sessions/SessionDetail.tsx` (session journey/timeline). Expect to rebase around it; the action added here is additive to that file.
- Related: #486 covers where these actions can be invoked from (row context menu, actions above the fold). The two share the action handlers and should be reviewed together, but neither blocks the other.

## HOW

**Technical Approach:** Extend the existing `navigate(id: ViewId)` in `App.tsx` — which already keeps a back/forward history — to carry an optional target, and expose it through a small context in `src/lib/nav.ts` so a nested view can call it without threading callbacks through `SessionsView`. `ChatsView` takes the target as a prop, following the `newChatNonce` precedent already in the tree. Views are conditionally rendered, so navigating to Chats mounts it fresh and its `useResource` fetch already includes the just-created chat — no cache invalidation needed.

**Implementation Details:**
- `src/lib/nav.ts` — add `NavTarget = { view: ViewId; chatId?: string }`, a `NavContext` and `useNavigate()`. Keep `ViewId`, `SECTIONS` and `VIEW_TITLES` where they are.
- `src/App.tsx` — `navigate` gains an optional second argument stored beside `view` (plus a nonce, so continuing the same session twice re-fires the selection). Wrap the view switch in the provider. Pass the target down: `<ChatsView inspectorOpen={inspectorOpen} newChatNonce={newChatNonce} openChatId={...} openChatNonce={...} />`.
- `src/views/ChatsView.tsx` — `useEffect` on `(openChatId, openChatNonce)` that calls `setSelected(openChatId)` and clears `drafting`, so the composer is the focused surface. Guard against selecting an id the list has not loaded yet.
- `src/views/SessionsView.tsx` — `continueInChat` calls `navigate("chats", { chatId: res.chat_id })` on success; drop `continuedChat` and its render. Keep `actionError` and render it where it is visible from the button.
- `src/views/sessions/SessionDetail.tsx` — add the action beside the existing header controls, calling the same handler (lift it, or duplicate the three-line `api.post` — prefer lifting so #486 has one place to wire).
- `src-tauri/src/native/sessions/continue_chat.rs` — the existing `UPDATE chat_sessions SET sdk_session_id = ?1, updated_at = ?2` already runs in the transaction; extend it to set `title` from `detail.summary`'s display title (fall back to the current default when it is empty). Note this is a deliberate divergence from the inherited behaviour, which is now allowed — the Go implementation is gone.

**Testing Strategy:**
- Rust: a unit/integration test over `continue_session` asserting the created row carries the source session's title and its `sdk_session_id`, and that an empty source title still yields the default. Check `parity/write_routes.json` still records the route as `native`; the response body (`{"chat_id"}`) does not change.
- Frontend: there is no TypeScript test harness in this repo (`CLAUDE.md` — `lib/typeAssert.ts` is the only value-level guard), so the nav path is verified by hand through the running app: continue a session, land in Chats, send a message, confirm the reply carries prior context. Use the `local-verify` / `ui-verify` skills; a dev build against `~/.agento-desktop-dev` is enough.
- Confirm reverting the navigation makes the manual check fail — i.e. the check is a real regression guard, not a smoke test.

**Documentation / Configuration:** Remove the "Cross-view navigation" bullet from `CLAUDE.md`'s **Known gaps** once it lands, and note the nav context under the frontend layout section. No new dependencies, no env vars.

**Files to Create / Modify:**
- Modify: `src/App.tsx`, `src/lib/nav.ts`, `src/views/SessionsView.tsx`, `src/views/ChatsView.tsx`, `src/views/sessions/SessionDetail.tsx`, `src-tauri/src/native/sessions/continue_chat.rs`, `CLAUDE.md`
- Create: none required (the nav context can live in `src/lib/nav.ts`)

**Risks:**
- *Nav context becomes a dumping ground* → keep it to `(view, optional target)`; it is not a router and should not grow query state.
- *Selecting a chat the list has not fetched* → the target is applied after the chats resource resolves, and ignored if no row matches.
- *Title change breaks a stored-shape assumption* → the response body is unchanged (`{"chat_id"}`); only the `title` column differs. Re-run the write tests.
- *Conflict with #479* on `SessionDetail.tsx` → whichever lands second rebases; the change here is a small additive control.

**Complexity:** Medium (6 files, well under 500 LOC, one small backend edit and a contained frontend pattern)

---
**Refined by:** AI Issue Refiner · **Date:** 2026-08-26 · **Status:** Ready

vibexp-agent:
Closes #485

## What

**Continue in chat** now takes the user to the resumed conversation instead of
silently creating a chat nobody is shown.

## Why

The backend was always healthy — `POST /api/claude-sessions/{id}/continue`
answers `201 {"chat_id"}` and `chat/runner.rs` already passes `sdk_session_id`
as `--resume`. Everything missing was on the frontend: `continueInChat`
resolved the POST and only set a note string, rendered as a small grey
`.sess-note` appended below the button at the bottom of a scrolling inspector.
No view change, no selection, no composer. The **error** rendered in the same
off-screen spot, so a 404 was visually indistinguishable from a success.

## How

- **`src/lib/nav.ts`** — `NavTarget` (a view id plus *one* optional row id),
  `NavProvider` and `useNavigate()`. Deliberately a hand-off, not a router.
  Two mechanisms now exist and which to use is decided by where the caller is
  rendered: a view `App.tsx` renders directly keeps `onNavigate` as a prop (the
  three gateway views); the context is for callers several levels down, where
  the prop would have to be threaded through every parent.
- **`src/App.tsx`** — `navigate` takes an optional target, stored beside `view`
  with a nonce so the same hand-off twice still fires. It is **cleared by any
  navigation that carries none**, or a consumed chat id would be re-applied on
  every later visit to Chats.
- **`src/views/ChatsView.tsx`** — `openChatId` / `openChatNonce`, mirroring
  `newChatNonce`. The id is applied without waiting for `/chats`: `detail`
  fetches by id, so the transcript and composer are live immediately and the row
  highlights when the list arrives. It claims `autoSelected` so the "select the
  most recent conversation" effect cannot overwrite the hand-off.
- **`src/views/SessionsView.tsx`** — navigates on success, `continuedChat` and
  its render are gone. The remaining note is only ever the failure, directly
  under the button pressed.
- **`src/views/sessions/SessionDetail.tsx`** — gains the same action, wired to
  the *same* handler rather than a second copy of the `api.post` (two copies is
  where one of them stops navigating), with its failure in the toolbar's own
  band.
- **`src-tauri/src/native/sessions/continue_chat.rs`** — the chat is titled from
  the source session's `display_title`, through `persist::truncate_title`'s
  existing 60-rune cut rather than a second spelling of it (`display_title`
  falls through to an unbounded transcript preview). An empty one keeps the
  store's `New Chat` default. A deliberate divergence from the inherited
  behaviour, which is now allowed.
- **`CLAUDE.md`** — the "Cross-view navigation" **Known gap** is removed and the
  nav context is documented on the `nav.ts` line.

The response body is unchanged (`{"chat_id"}`), so no golden moves and
`parity/write_routes.json` still records the route as `native`.

## Acceptance

- [x] Clicking **Continue in chat** switches to Chats with the new chat selected
      and its composer ready.
- [x] The continued chat is titled from the source session's `display_title`.
- [x] Sending the first message resumes the Claude conversation — untouched;
      `runner.rs` already passes `--resume`.
- [x] A failed continue shows an error without scrolling, and does not navigate.
- [x] The same action exists in `SessionDetail.tsx`.
- [x] Repeat clicks are bounded: `busy` disables the button in flight, and
      success unmounts the view.

## Tests

Three unit tests over `continue_session` in `continue_chat.rs`: the title is the
source session's, a long one is cut at 60 runes, and an untitled source keeps
`New Chat`. **Verified as a regression guard** — reverting the `title = ?1`
column out of the UPDATE fails the first two with
`left: "New Chat"`, and leaves the third passing (it asserts the fallback).

There is no TypeScript test harness in this repo, so the nav path is covered by
`tsc` and review rather than by a test.

## Out of scope

Replaying the source transcript into the chat view; the row context menu and
moving the inspector actions above the fold (#486); backend idempotency for a
session continued twice (see the issue's **Context** — branching a session twice
is legitimate, and after the first turn `persist.rs` rewrites `sdk_session_id`
from the stream).

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -176,7 +176,19 @@ src/
     hooks.ts     useResource / useDebounced / usePoll / describeError
     format.ts    compactNumber, usd, duration, relativeTime, toneFor, …
     stats.ts     cross-view counters for sidebar + status bar
-    nav.ts       sidebar sections, view ids, titles
+    nav.ts       sidebar sections, view ids, titles — and the cross-view
+                 hand-off (#485): `NavTarget`, `NavProvider` and `useNavigate`.
+                 **Two mechanisms, and which one to use is decided by where the
+                 caller is rendered.** A view `App.tsx` renders directly takes
+                 `onNavigate={navigate}` as a prop (the three gateway views);
+                 the context is for a caller several levels down, where the prop
+                 would have to be threaded through every parent — the Sessions
+                 inspector and `SessionDetail`. `NavTarget` is a view id plus
+                 *one optional row id*, and it must stay that: it is a hand-off,
+                 not a router, and query state, filters and scroll positions do
+                 not belong in it. `App` clears the target on any navigation
+                 that carries none, or a consumed chat id would be re-applied on
+                 every later visit to that section
     icons.tsx    16px / 1.5-stroke icon set
     tauri.ts     window + menu bridge; degrades to a plain browser tab
     clipboard.ts copyText, with the execCommand fallback WebKitGTK needs
@@ -3954,9 +3966,6 @@ because each one cost real time to find:
   serde's recursion limit, a site URL `url::Url` rejects. Each is documented at
   its site. They were invisible while something else could answer; they are
   user-visible 500s now, and each is a small piece of work to resolve properly.
-- Cross-view navigation: "Continue in chat" on a session can only report the
-  new `chat_id`; it cannot switch to the Chats view. Needs a nav context in
-  `App.tsx`.
 - `useAppStats` counters refresh on a 30s poll and on window focus, not on
   mutation, so a create in one view lags in the sidebar briefly.
 - Session table is not virtualised; 900+ rows render eagerly after "Load more".
--- a/src-tauri/src/native/chat/persist.rs
+++ b/src-tauri/src/native/chat/persist.rs
@@ -127,7 +127,7 @@ fn append_message(
 ///
 /// Runes, not bytes — a title of accented text would otherwise be cut mid
 /// character and stored as invalid UTF-8.
-fn truncate_title(content: &str, max: usize) -> String {
+pub(crate) fn truncate_title(content: &str, max: usize) -> String {
     let count = content.chars().count();
     if count <= max {
         return content.to_string();
--- a/src-tauri/src/native/sessions/continue_chat.rs
+++ b/src-tauri/src/native/sessions/continue_chat.rs
@@ -79,11 +79,12 @@ pub fn continue_session(db_path: &std::path::Path, session_id: &str) -> Result<A
     //
     // This is a deliberate divergence from the inherited behaviour, which left
     // every continued chat titled "New Chat".
-    let title = crate::native::chat::persist::truncate_title(&detail.summary.display_title, 60);
-    let title = if title.is_empty() {
+    let from_source =
+        crate::native::chat::persist::truncate_title(&detail.summary.display_title, 60);
+    let title = if from_source.is_empty() {
         session.title.as_str()
     } else {
-        title.as_str()
+        from_source.as_str()
     };
 
     // `chatService.UpdateSession`, which stamps its own `updated_at` — so this
--- a/src-tauri/src/native/sessions/continue_chat.rs
+++ b/src-tauri/src/native/sessions/continue_chat.rs
@@ -69,13 +69,35 @@ pub fn continue_session(db_path: &std::path::Path, session_id: &str) -> Result<A
         },
     )?;
 
+    // The chat is titled from the session it continues, so it is identifiable
+    // in the Chats list instead of joining the pile of untitled "New Chat" rows
+    // (#485). `display_title` falls all the way through to the transcript's
+    // first-message preview, which is unbounded, so it goes through the *same*
+    // 60-rune cut a chat's own derived title takes — one spelling of "how long
+    // a chat title may be", not two. An empty one keeps `insert_session`'s
+    // default rather than storing a blank.
+    //
+    // This is a deliberate divergence from the inherited behaviour, which left
+    // every continued chat titled "New Chat".
+    let title = crate::native::chat::persist::truncate_title(&detail.summary.display_title, 60);
+    let title = if title.is_empty() {
+        session.title.as_str()
+    } else {
+        title.as_str()
+    };
+
     // `chatService.UpdateSession`, which stamps its own `updated_at` — so this
     // is a second statement with a second clock reading rather than a value
-    // folded into the insert above. It writes every column, but only these two
-    // can differ from what was just inserted.
+    // folded into the insert above. It writes every column, but only these
+    // three can differ from what was just inserted.
     tx.execute(
-        "UPDATE chat_sessions SET sdk_session_id = ?1, updated_at = ?2 WHERE id = ?3",
-        rusqlite::params![session_id, crate::native::gotime::now_go_text(), session.id],
+        "UPDATE chat_sessions SET title = ?1, sdk_session_id = ?2, updated_at = ?3 WHERE id = ?4",
+        rusqlite::params![
+            title,
+            session_id,
+            crate::native::gotime::now_go_text(),
+            session.id
+        ],
     )
     .map_err(|e| WriteError::Fallback(format!("linking claude session: {e}")))?;
 
@@ -142,10 +164,10 @@ mod tests {
             .to_string();
 
         let conn = rusqlite::Connection::open(&db).expect("open");
-        let (count, sdk, cwd, model, slug): (i64, String, String, String, String) = conn
-            .query_row(
+        let (count, sdk, cwd, model, slug, title): (i64, String, String, String, String, String) =
+            conn.query_row(
                 "SELECT (SELECT COUNT(*) FROM chat_sessions), sdk_session_id,
-                        working_directory, model, agent_slug
+                        working_directory, model, agent_slug, title
                  FROM chat_sessions WHERE id = ?1",
                 [&chat_id],
                 |row| {
@@ -155,6 +177,7 @@ mod tests {
                         row.get(2)?,
                         row.get(3)?,
                         row.get(4)?,
+                        row.get(5)?,
                     ))
                 },
             )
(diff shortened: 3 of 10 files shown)


vibexp-agent:
Self-review round found two things in the first commit, both fixed in `d3fe523`:

- **The composer never took focus.** `autoFocus` cannot do it — React applies it on mount and this textarea survives a change of chat — so `Composer` takes a `focusNonce` and focuses on a bump, deliberately never on mount.
- **`actionError` was one string shared by the favourite and continue actions**, and the full session view rendered it under its "Continue in chat" button — a failed favourite would have surfaced as text under a button nobody pressed. It now carries which action failed.

CI green on `d3fe523` (6/6). Rebase onto `main` was a no-op — `origin/main` is still `9e3da7e`, so the merged SHA is the one CI validated.

PR #488, opened 2026-08-26
feat: right-click actions on a session row, and pin the inspector actions (#486)

### LINKED ISSUE #486: Sessions: right-click actions on a row, and move the inspector actions above the fold

vibexp-agent:
Closes #486

## What

Two ways to reach the session actions without hunting for them.

**A right-click menu on a session row** — View session, Add / Remove favourite,
Continue in chat, Copy session ID, Copy project path. Right-clicking selects the
row first, so the inspector agrees with what the menu is about to act on.

**The inspector's actions pinned above the fold** — the bottom `Actions` group is
now a compact icon strip between `.inspector__head` and `.inspector__scroll`, so
it cannot scroll away on a session with a full metadata block.

Both surfaces call the same `toggleFavorite` / `continueInChat` / `setOpenId`.

## How

- `src/components/ui.tsx` — a new `ContextMenu({ at, items, onClose })`, written
  beside `Dropdown` rather than by refactoring `Dropdown`'s internals into a
  shared popover (see the issue's flagged judgement call; extract it if a third
  consumer appears).
- `src/views/SessionsView.tsx` — `onContextMenu` on the row, a `menu` state
  holding the point and the **row id**, and the strip.
- `src/styles/controls.css` — `.ctxmenu` positioning and `.menu__item--active`.
- `src/styles/sessions.css` — `.sess-strip`, replacing `.sess-actions`.

## Three things the running app corrected

Verified through the WebKit inspector against the real Tauri webview
(`.claude/skills/ui-verify`), which is what found all three — none is visible
from the source:

1. **A `keydown` on a `<body>` portal does not reach React's delegated
   listener.** Written first as an `onKeyDown` prop the way `Dropdown` has it,
   Escape and the arrow keys did nothing at all. Keys are taken from `window` in
   the **capture** phase instead, which also puts the menu ahead of the sessions
   list's own global Enter handler — so `stopPropagation` on a claimed key is
   what stops Enter both opening the session and picking the item.
2. **`focus()` on a `visibility: hidden` element is a silent no-op.** The first
   pass renders hidden so the box can be measured and clamped; focusing in that
   same layout effect never took. Focus now happens once `pos` has made the menu
   visible.
3. **`is_favorite` is `omitempty`.** A session that is not a favourite has no
   such key, so `aria-pressed={selected.is_favorite}` made React drop the
   attribute — a toggle announcing its state in only one of its two states.
   Coerced with `!!`.

## Verification

There is no TypeScript test harness in this repo, so this is by hand in the
running app. `npm run build` (`tsc --noEmit` + `vite build`) is green; nothing
under `src-tauri/` is touched.

Confirmed in the webview:

- right-click selects the row and opens the menu at the cursor with all five
  items; `defaultPrevented` is true over a row and **false over an `<input>`**,
  where the webview's own Copy/Paste menu must still win
- arrows / Home / End / wrap-around move the active item, Enter fires it,
  Escape closes; menu focused on open
- clamped inside the viewport when opened at the bottom-right corner
  (`right: 1279 / 1280`, `bottom: 818 / 820`)
- closes on outside mousedown, on scroll, and on choosing an item
- favourite toggles from the menu and the label flips `Add` ↔ `Remove`; View
  session opens the transcript
- the strip fits the 220px `Splitter` minimum (199px row, no overflow)

**One criterion could not be proven on this machine, and it is not a
regression:** the clipboard. WebKitGTK refuses `writeText` outside a user
gesture it recognises, and this box has no `xdotool`/`xclip` for real OS input —
so under synthetic clicks the two new copy items fail *and so does the existing,
shipped `CopyButton` in the inspector*, identically. The menu items call the
same `copyText` from the same synchronous position inside a real `<button>`
click handler, so there is no behavioural difference between the new call sites
and the proven one. A failure is surfaced rather than swallowed: `copyValue`
renders "Could not copy the … to the clipboard." in the strip, which is what
that check actually exercised.

## Deviation from the issue's HOW

`.ctxmenu`'s rules are in `styles/controls.css` beside `.dropdown__menu`, not in
`styles/sessions.css` as the HOW suggests. `ContextMenu` is a shared control in
`components/ui.tsx`; positioning it from one view's stylesheet would leave the
next consumer unpositioned. The strip's own rules are in `sessions.css`, per the
per-view convention.

## Out of scope

What Continue in chat does once invoked (#485, merged), context menus on other
views' rows, multi-select, and a native OS context menu.

### DIFF
--- a/src/components/ui.tsx
+++ b/src/components/ui.tsx
@@ -406,6 +406,11 @@ export function ContextMenu({
       left: Math.max(m, Math.min(at.x, window.innerWidth - width - m)),
       top: Math.max(m, Math.min(at.y, window.innerHeight - height - m)),
     });
+    // Reopening does not always remount: right-clicking a second row while the
+    // menu is open closes and reopens it in one batch, so this component is
+    // reused and a stale `active` would carry over — leaving a highlighted item
+    // the user never chose, which Enter would then fire.
+    setActive(-1);
   }, [at.x, at.y]);
 
   // Focus only once `pos` has made the menu visible: a `visibility: hidden`
@@ -470,11 +475,19 @@ export function ContextMenu({
 
   useEffect(() => {
     const onKey = (e: KeyboardEvent) => {
+      // A chorded key is somebody else's: this listener is at the window's
+      // capture phase, so claiming Ctrl+Enter or Cmd+ArrowDown here would take
+      // it from every other handler in the app while a menu happens to be open.
+      if (e.metaKey || e.ctrlKey || e.altKey) return;
       switch (e.key) {
         case "Escape":
-        case "Tab":
           onClose();
           break;
+        // Tab closes and is then left alone, the way `Dropdown` treats it, so
+        // focus still moves on to the next control rather than being trapped.
+        case "Tab":
+          onClose();
+          return;
         case "ArrowDown":
           move(1);
           break;
--- a/src/components/ui.tsx
+++ b/src/components/ui.tsx
@@ -342,6 +342,205 @@ export function Dropdown({
   );
 }
 
+/* --- Context menu --------------------------------------------------------- */
+
+export interface ContextMenuItem {
+  label: string;
+  icon?: IconName;
+  danger?: boolean;
+  disabled?: boolean;
+  onSelect(): void;
+}
+
+/**
+ * A menu anchored to a **point** rather than to a control — what a right-click
+ * on a row opens.
+ *
+ * It is written beside `Dropdown` rather than by extracting a shared popover
+ * out of it. `Dropdown` is a working, keyboard-accessible listbox used across
+ * several views, and refactoring its internals is a larger and riskier change
+ * than the one consumer here justifies; extract the popover if a third one
+ * appears.
+ *
+ * Two things differ from `Dropdown`, and both follow from there being no
+ * trigger element:
+ *
+ * - **Placement is measured, not estimated.** `Dropdown` clamps against its
+ *   trigger's rect, which it has before the menu exists. Here the only input is
+ *   a pointer position, so the first pass renders at the origin, *hidden*, and
+ *   a layout effect clamps against the box the browser actually laid out —
+ *   before it paints, so there is no flash. Rendering that first pass at `at`
+ *   instead would measure a shrink-to-fit box squeezed by whatever room is left
+ *   to the right of the cursor, i.e. the wrong width near the edge this exists
+ *   to handle.
+ * - **Keys are taken from `window` in the capture phase, not from an
+ *   `onKeyDown` prop.** `Dropdown` can use the prop because its handler sits on
+ *   the trigger, which is inside the React root; this menu is portaled to
+ *   `<body>`, and a `keydown` there does not reach React's delegated listener —
+ *   measured in the running app, where Escape and the arrow keys did nothing
+ *   at all. Capturing at the window also puts this ahead of the sessions
+ *   list's own global Enter handler, so `stopPropagation` on a key the menu
+ *   claims is what stops Enter opening the session *and* picking the item. The
+ *   menu still takes focus on open, which is where a menu's focus belongs.
+ */
+export function ContextMenu({
+  at,
+  items,
+  onClose,
+}: {
+  /** Viewport coordinates — `clientX` / `clientY` of the opening event. */
+  at: { x: number; y: number };
+  items: ContextMenuItem[];
+  onClose(): void;
+}) {
+  const menuRef = useRef<HTMLDivElement>(null);
+  const [pos, setPos] = useState<{ left: number; top: number }>();
+  const [active, setActive] = useState(-1);
+
+  useLayoutEffect(() => {
+    const el = menuRef.current;
+    if (!el) return;
+    const { width, height } = el.getBoundingClientRect();
+    const m = 4;
+    setPos({
+      left: Math.max(m, Math.min(at.x, window.innerWidth - width - m)),
+      top: Math.max(m, Math.min(at.y, window.innerHeight - height - m)),
+    });
+  }, [at.x, at.y]);
+
+  // Focus only once `pos` has made the menu visible: a `visibility: hidden`
+  // element is not focusable, so focusing in the measuring effect above is
+  // silently a no-op — which is exactly what it did until the running app was
+  // asked what `document.activeElement` was.
+  useEffect(() => {
+    if (pos) menuRef.current?.focus({ preventScroll: true });
+  }, [pos]);
+
+  // "Outside" is outside the menu alone: the point it was opened at belongs to
+  // whatever the user right-clicked, which must not swallow the dismissal.
+  // Any scroll or resize under an open menu would strand it, so close.
+  useEffect(() => {
+    const onDown = (e: MouseEvent) => {
+      if (menuRef.current?.contains(e.target as Node)) return;
+      onClose();
+    };
+    const onScroll = (e: Event) => {
+      if (menuRef.current?.contains(e.target as Node)) return;
+      onClose();
+    };
+    document.addEventListener("mousedown", onDown);
+    window.addEventListener("scroll", onScroll, true);
+    window.addEventListener("resize", onClose);
+    window.addEventListener("blur", onClose);
+    return () => {
+      document.removeEventListener("mousedown", onDown);
+      window.removeEventListener("scroll", onScroll, true);
+      window.removeEventListener("resize", onClose);
+      window.removeEventListener("blur", onClose);
+    };
+  }, [onClose]);
+
+  /** Walk to the next enabled item, wrapping; a menu of only disabled items stays put. */
+  const move = useCallback(
+    (dir: 1 | -1) => {
+      setActive((i) => {
+        const n = items.length;
+        if (n === 0) return -1;
+        let next = i < 0 && dir === 1 ? -1 : Math.max(0, i);
+        for (let k = 0; k < n; k++) {
+          next = (next + dir + n) % n;
+          if (!items[next].disabled) return next;
+        }
+        return i;
+      });
+    },
+    [items]
+  );
+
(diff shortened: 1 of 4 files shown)


vibexp-agent:
## Not auto-merged — CI never ran

GitHub Actions was in a **major outage** for the whole of this run (`Actions: major_outage` on githubstatus, two incidents open). No `CI` workflow run was ever created for this PR — CodeQL sat `queued` for ~35 minutes and a close/reopen produced nothing. This job does not merge on a check that never ran, so the PR is handed off ready for review.

**What did pass, locally:**
- `npm run build` (`tsc --noEmit` + `vite build`) green on the final commit.
- Nothing under `src-tauri/` is touched, so `cargo fmt`/`clippy`/`test` are unaffected by this diff — but they are still unverified for this branch and CI should confirm them.
- The behaviour was verified against the real Tauri webview; see the PR body.

**Automated review:** one round, three findings, all applied in `b185aae` before hand-off — the stale `active` index on reopen (the component is reused when you right-click a second row while a menu is open, so Enter could fire an item nobody chose), chorded keys being swallowed by a window-capture listener, and `Tab` being `preventDefault`ed so focus was trapped. Each was re-verified in the running app.

Re-run CI when Actions recovers; merge is a one-liner from there.

PR #491, opened 2026-08-27
fix: show a resumed session's history, and stop continuing it twice (#490)

### LINKED ISSUE #490: A resumed session opens an empty transcript — show the history, and stop continuing the same session twice
## Problem

Continuing a Claude session opens a chat with an **empty transcript**. The title and SDK session id are right, the model genuinely has the history — but the user sees nothing, so there is no way to read what was said before deciding what to send next. Practically, a conversation you cannot see is a conversation you cannot continue.

Reproduced on v1.1.0 against a real corpus:

```
POST /api/claude-sessions/e475bf30-.../continue
GET  /api/chats/<new>   ->   title: 'Create daily morning briefing'
                             sdk_session_id: e475bf30-...
                             messages: 0            <-- the whole bug
```

Then one turn into that chat:

```
> "Without using any tools: what was the very first instruction I gave you in this conversation?"
< "Your first instruction was that you're an autonomous agent running on a daily
   morning schedule in Berlin time, tasked with generating a catch-up briefing…"
```

So `--resume` works and no new Claude session is created: the turn was **appended to the same transcript** (`e475bf30-….jsonl`, same `sessionId`, no new file, no fork), and `sdk_session_id` was unchanged afterwards.

## Root cause

`native/sessions/continue_chat.rs` creates a chat row that only *points* at the Claude session — it copies nothing and records no relationship beyond `sdk_session_id`. `GET /api/chats/{id}` builds `messages` from the `chat_messages` table, which is empty until Agento itself takes a turn, and `ChatsView` renders only that. The prior conversation lives in the transcript JSONL, which nothing but the CLI reads.

This is not a v1.1.0 regression — the Go web UI behaved identically (`handleContinue` did `navigate('/chats/' + chat_id)` and its chat page had no transcript loading). #485 restored the navigation, which is what made the emptiness visible for the first time. It was named in #485 as an explicit *Out of Scope* bullet; this issue reverses that call, because without it the feature does not do what its name says.

## Second defect: continuing twice creates a second chat

`continue_session` inserts unconditionally, so continuing the same session again mints another chat row pointing at the same Claude session. Continue means *resume*, and one Claude session should have one chat.

## What good looks like

- Opening a continued chat shows the previous conversation, scrolled to the end, with the composer beneath it — the same thing the Sessions view already renders, read-only.
- Continuing a session that was already continued reopens **that** chat rather than making another.

## Notes for whoever picks this up

- `GET /api/claude-sessions/{id}` already re-reads a session live from its transcript and returns normalized blocks including sub-agents (#483) — no new backend route is needed to get the history.
- **Trap, found by measurement:** the CLI appends resumed turns to the *same* transcript file. So after the first Agento turn the source transcript already contains the local messages, and rendering the whole transcript plus `chat_messages` would double-render the latest turn. The boundary has to be recorded at continue time, or the transcript treated as the single source for continued chats.
- Prefer a dedicated column (e.g. `continued_from_session_id`) over keying anything on `sdk_session_id`: `persist.rs` rewrites that column from whatever the stream reports, so it is not a stable key. The next migration number is 37.
- Careful with the dedupe lookup: a chat Agento created itself has `sdk_session_id` equal to its **own chat id** (the turn pins the CLI session to it), and those sessions appear in the corpus too — so "continue" on one of them should find the existing chat rather than clone it.

---
## WHAT

**Summary:** A chat created by **Continue in chat** shows the conversation it resumes, and continuing an already-continued session reopens that chat instead of creating another.

**Scope:**
- Record what a chat was continued from, and where the inherited history ends, at continue time.
- Render that history read-only above the chat's own messages, using the existing session-detail read.
- Make `POST /api/claude-sessions/{id}/continue` reuse an existing chat for the same source session.

**Affected Components:**
- `src-tauri/src/native/migrate.rs` + `parity/migrations_vectors.json` — migration **37**, three columns on `chat_sessions`.
- `src-tauri/src/native/sessions/continue_chat.rs` — the lookup-before-insert, and the new columns' values.
- `src-tauri/src/native/chats.rs` — `GET /api/chats/{id}` carries the continuation fields on `session`.
- `src/views/ChatsView.tsx` — fetch and render the inherited transcript above the live messages.
- `src/lib/types.ts` — the new `ChatSession` fields, snake_case, verbatim.
- `src/views/sessions/SessionDetail.tsx` — the transcript renderer to reuse rather than re-implement.

**Acceptance Criteria:**
- [ ] Opening a continued chat shows the prior conversation, in order, scrolled to the end, with the composer beneath it — no empty state, no "No messages yet".
- [ ] After sending a message in a continued chat, the new turn appears **once** — not duplicated against the inherited history (see the append trap in **Context**).
- [ ] Continuing a session that already has a chat returns that chat's id and creates no second row; the response stays `201 {"chat_id"}` with no shape change.
- [ ] Continuing a session with no existing chat behaves exactly as today (new chat, titled from the session, `sdk_session_id` set).
- [ ] A continued chat whose source transcript is unreadable or gone still opens and is usable — the history area reports it honestly rather than failing the chat.
- [ ] The inspector's `Messages` count and usage figures do not silently claim the inherited turns as this chat's own.

**Out of Scope:**
- Importing the transcript into `chat_messages`. It duplicates megabytes per session and would corrupt the chat's own usage and cost figures, which are stored, not derived.
- Editing, searching or filtering within the inherited history.
- Backfilling continuation metadata onto chats created before migration 37 — they keep today's behaviour and can be continued afresh.

## WHY

**Business Value:** Resuming a past Claude Code session is one of the reasons Agento exists. A resumed conversation you cannot read is one you cannot sensibly continue — the user has to decide what to send next with no view of what was already said.

**Problem Statement:** Reproduced on v1.1.0 against a real corpus:

```
POST /api/claude-sessions/e475bf30-.../continue
GET  /api/chats/<new>  ->  title: 'Create daily morning briefing'
                           sdk_session_id: e475bf30-...
                           messages: 0
```

The chat row is correct and the resume genuinely works — one probe turn into that chat recalled the original conversation's first instruction, the turn was appended to the **same** transcript (`e475bf30-….jsonl`, same `sessionId`, no fork, no new file), and `sdk_session_id` was unchanged afterwards. What is missing is only the display: `continue_chat.rs` copies nothing and records no relationship, `GET /api/chats/{id}` builds `messages` from `chat_messages` (empty), and `ChatsView` renders only that. Separately, `continue_session` inserts unconditionally, so continuing the same session twice mints a second chat pointing at one Claude session.

**Expected Impact:** The feature does what its name says. Users can read the thread they are resuming and reply in context; the Chats list stops accumulating duplicate rows for one conversation.

**Context:**
- **Maintainer decision (supersedes the call flagged in #485):** continue is idempotent per source session — one Claude session, one chat. #485 recommended the opposite (allowing several chats to branch one session) and flagged it; that is overruled here.
- **Trap, established by measurement:** the CLI **appends resumed turns to the same transcript file**. So after the first Agento turn the source transcript already contains the local messages, and rendering the whole transcript plus `chat_messages` double-renders the newest turn. The boundary must be captured at continue time.
- **Do not key anything on `sdk_session_id`:** `chat/persist.rs` rewrites that column from whatever the stream reports, so it is not a stable identity for the source session.
- **A chat Agento created itself has `sdk_session_id` equal to its own chat id** (the turn pins the CLI session to it) and those sessions are indexed in the corpus too — so "continue" invoked on one of them should find the existing chat rather than clone it. The dedupe lookup gets this right for free if it matches on the source session's row key.
- The corpus keys sessions on the **pair** `(session_id, project_path)`, not the id alone (`CLAUDE.md`, the #362 family) — the stored reference must be the pair.
- `GET /api/claude-sessions/{id}` already re-reads a session live from its transcript and returns normalized blocks including sub-agents (#483), so no new backend route is needed for the history.
- Related: #485 (navigation, shipped in v1.1.0) and #486 (where the actions can be invoked from).

## HOW

**Technical Approach:** Store the continuation as explicit state rather than inferring it. Migration 37 adds `continued_from_session_id`, `continued_from_project_path` and `continued_from_message_count` to `chat_sessions`; `continue_chat.rs` fills them inside the transaction it already opens, after first looking for an existing chat for that pair. The chat view fetches the source session through the existing detail endpoint and renders its first `continued_from_message_count` messages above the chat's own — a fixed prefix, so the turns the CLI later appends to that same file cannot be rendered twice.

**Implementation Details:**
- `native/migrate.rs` / `parity/migrations_vectors.json` — migration 37: three nullable columns, no backfill (`CLAUDE.md`: the migrations are append-only and `migrations_vectors.json` **is** the schema; 31 onward are authored there directly).
- `native/sessions/continue_chat.rs`
  - Before inserting, `SELECT id FROM chat_sessions WHERE continued_from_session_id = ?1 AND continued_from_project_path = ?2 ORDER BY created_at DESC LIMIT 1` — hit returns `201 {"chat_id"}` for that row and writes nothing. Legacy duplicates make the `ORDER BY` meaningful; keep it.
  - On insert, extend the existing `UPDATE chat_sessions SET title = …, sdk_session_id = …` (already in the transaction) with the three columns. `continued_from_message_count` is the length of the normalized message list `detail::get` just returned — it is already loaded, so this costs nothing extra.
- `native/chats.rs` — carry the three fields on the `session` object of `GET /api/chats/{id}` (and the list, if cheap). Field order is wire order; use `skip_serializing_if` so existing responses do not gain empty keys.
- `src/lib/types.ts` — mirror the fields verbatim, `string | null` per the repo's convention.
- `src/views/ChatsView.tsx` — when `continued_from_session_id` is set, `api.get('/claude-sessions/' + id)` alongside the chat, render `messages.slice(0, continued_from_message_count)` in a read-only region above the live messages, with a divider naming the source session. Keep the existing scroll-to-end behaviour so the composer is what the user lands on. A failed fetch renders an inline note in that region and leaves the chat usable.
- `src/views/sessions/SessionDetail.tsx` — lift the message/block renderer so both views use one implementation; do not fork the tool-result and sub-agent handling that #483 just landed.

**Testing Strategy:**
- Rust: `continue_session` called twice for one session returns the same `chat_id` and leaves one row; the first call stores the pair and a message count matching `detail::get`; a session with no transcript still 404s as today. Add a migration-37 case to the existing migration assertions.
- Rust: a chat continued and then advanced by a turn keeps `continued_from_message_count` unchanged (it is a boundary, not a running total).
- Frontend: no TypeScript test harness exists in this repo, so verify by hand through the running app (`local-verify` / `ui-verify`): continue a session with a substantial transcript, confirm the history renders and the newest turn appears exactly once after sending a message, and confirm reverting the prefix slice reproduces the double-render.
- Confirm reverting the fix makes the Rust dedupe test fail — if it does not, the test is not a regression guard.

**Documentation / Configuration:** Update `CLAUDE.md` where `sessions/continue_chat.rs` is described (it currently documents the two writes and the transaction) to record the dedupe and the three columns. No new dependencies, no configuration.

**Files to Create / Modify:**
- Modify: `src-tauri/src/native/migrate.rs`, `parity/migrations_vectors.json`, `src-tauri/src/native/sessions/continue_chat.rs`, `src-tauri/src/native/chats.rs`, `src/lib/types.ts`, `src/views/ChatsView.tsx`, `src/views/sessions/SessionDetail.tsx`, `CLAUDE.md`
- Create: none

**Risks:**
- *Double-rendering the newest turn* → the stored prefix count is the guard; it is the first thing to test, and the transcript-append behaviour above is why.
- *Large transcript on every chat open* → the detail read is a live re-read; if it proves slow on the biggest sessions, render the prefix lazily (collapsed "show earlier" region) rather than importing rows.
- *A compacted or rewritten transcript shifting the boundary* → clamp the slice to the list length; never index past it.
- *Dedupe hiding a legitimate second chat* → out of scope by decision above; if branching is wanted later it should be an explicit action, not an accident.

**Complexity:** Medium (7 files, one migration, well under 500 LOC)

---
**Refined by:** AI Issue Refiner · **Date:** 2026-08-27 · **Status:** Ready

vibexp-agent:
Closes #490

## What

A chat created by **Continue in chat** now shows the conversation it resumes, and continuing an already-continued session reopens that chat instead of minting a second one.

## Why

`continue_chat.rs` created a chat that only *pointed* at a Claude session through `sdk_session_id`, and `GET /api/chats/{id}` builds `messages` from `chat_messages` — empty until Agento itself takes a turn. So a resumed conversation opened blank: the model genuinely had the history and the user could not read a word of it, which is not a state anyone can sensibly reply from. Separately, `continue_session` inserted unconditionally, so a second click minted another chat for one conversation.

## How

**Migration 37** adds three columns to `chat_sessions`: `continued_from_session_id`, `continued_from_project_path`, `continued_from_message_count`. `NOT NULL DEFAULT` rather than nullable, following migration 30 — the one additive shape SQLite applies without rebuilding the table, and it leaves "not a continuation" one spelling instead of two. No backfill.

**The pair, not the id.** The corpus keys a session on `(session_id, project_path)`, so that is what is stored — and deliberately *not* `sdk_session_id`, which `chat/persist.rs` rewrites from whatever the stream reports.

**The count is a boundary, not a total, and nothing advances it.** The CLI appends a resumed turn to the *same* transcript file, so after the first Agento turn the source already holds the local messages; the view renders exactly the recorded prefix and its own `chat_messages` after it. The slice is taken from the **unfiltered** endpoint list, before the display filters, because that is what the count was read from.

**Continue is idempotent per source session.** `continue_session` looks a chat up inside its `Immediate` transaction before inserting. The lookup also matches a chat whose *own id* is the session id: a turn pins the CLI session to the chat id, so an Agento-created chat is indexed in the corpus under that id, and "continue" on one of those reopens the chat it already is.

**One renderer.** `SessionDetail`s message/block renderer moves to `src/views/sessions/SessionTranscript.tsx`, so the chat view reuses it rather than forking the tool-result and sub-agent handling #483 landed. The inherited history renders inside the live transcript's own scroller, so the view keeps one scrollbar and the reader lands at the composer.

## Acceptance criteria

- [x] A continued chat shows the prior conversation, in order, scrolled to the end, composer beneath — no empty state
- [x] After sending, the new turn appears **once**, not duplicated against the inherited history
- [x] Continuing an already-continued session returns that chat's id, creates no second row, `201 {"chat_id"}` unchanged
- [x] Continuing a fresh session behaves exactly as before
- [x] A continued chat whose source transcript is unreadable or gone still opens and is usable — the history region reports it
- [x] The inspector's `Messages` count and usage figures do not claim the inherited turns

## Tests

- `the_source_pair_and_the_message_boundary_are_recorded` — the pair and the count, asserted against `detail::get`'s own list rather than a literal
- `continuing_the_same_session_twice_returns_the_one_chat` — reverting the lookup fails it on both assertions (verified)
- `a_transcript_that_grew_afterwards_does_not_move_the_boundary` — appends a real turn to the same file, asserts the fixture grew first so it cannot pass vacuously
- `continuing_a_chat_agento_created_reopens_it_rather_than_cloning`
- `the_continuation_survives_the_read_path` (`chats::get`) and `a_patch_carries_the_continuation_fields_back` (`get_session_tx`) — two differently ordered positional readers, two tests
- `an_ordinary_chat_carries_no_continuation_keys` — every pre-migration-37 response stays byte-identical
- Verified end to end against a scratch instance under a swapped `HOME`: two `POST .../continue` calls return one `chat_id`, one row, and the fields land on the wire.

## Deviations from the HOW

- The columns are `NOT NULL DEFAULT` rather than nullable (reasoning above); the lookup and the wire shape are unaffected.
- The dedupe lookup also matches `id = ?session_id`, which the HOW said the pair would cover "for free" — it does not, because an Agento-created chat records no source.
- `src-tauri/src/native/search/mod.rs` was not in the file list. Its three upgrade tests pinned their boundary to `expected_version()`, so appending any migration turned them into fresh-install tests; they now pin `SEARCH_KEY_MIGRATION = 36`, the migration they actually exercise.

Deferred to #492: the history is fetched by session id alone while the pair is what is stored.


vibexp-agent:
Deferred to #492 — the inherited history is fetched by session id alone while migration 37 stores the pair, so an id existing under two project paths can resolve to the wrong transcript. Narrow (UUIDs), and closing it properly is either a view-side pair check or a lookup-by-pair route; not folded into this PR.

Issue #273, opened 2026-08-14
Drop WhatsApp from the desktop app cleanly


vibexp-agent:
## Scope

`whatsmeow` is **not being ported** (decided — no Rust equivalent). This issue is the fallout, not the port.

- Do **not** port `internal/integrations/whatsapp`, and do not keep a Go sidecar, subprocess or shim for it.
- The desktop UI must not offer WhatsApp as an integration type — no pairing flow, no QR screen, no picker entry. An option that cannot work is worse than an absent one.
- **Existing stored rows must degrade, not crash.** A user who paired WhatsApp under the Go server has an `integrations` row of that type; the Rust app must ignore it or show it unavailable rather than failing the whole integrations list on a type it cannot construct.
- Agents whose tool allowlists name WhatsApp tools keep those entries harmlessly; the tools simply never resolve.
- `main` keeps WhatsApp indefinitely. The desktop app is what loses it, and that is the accepted trade.

## BLOCKED BY: integration reads.
---
**Branch:** target `desktop`, never `main`. One PR, merged before the next task starts.
**Bar:** byte-identical JSON. Any field-name, key-order, escaping or float-spelling drift is a regression.
**Verify:**
```bash
cd desktop && eval "$(./scripts/parity-instance.sh start)"
(cd src-tauri && cargo test --test live_parity -- --ignored --nocapture)
./scripts/parity-instance.sh stop
```
**Read first:** VibeXP artifact `agento-desktop-rust-port-handover-2026-08-14`, then `desktop/CLAUDE.md`.
Never diff against the server on :8990 (stale binary). Seed the scratch instance if the endpoint has no local data, or an empty response diffs clean and proves nothing.

**Blocked by:** #272

_Unblocked issues in this epic can be worked in parallel._

vibexp-agent:
Done in #292 (squash-merged to `desktop` as 3466718).

WhatsApp is gone from the desktop UI: the `PROVIDERS` catalog entry (which is what the picker renders from), the device-pairing panel and its four `/whatsapp/*` calls, the 416-line QR encoder written solely for it, and the `.qr` CSS that was duplicated across two stylesheets.

Stored rows degrade rather than crash — nothing models an integration type, so a row paired under the Go server still lists and opens, and the detail pane now says WhatsApp is unavailable here instead of the misleading "created by a newer version of Agento".

[...]

PR #498, opened 2026-08-27
feat: read mcps.yaml in the native chat runner, and stop its mere presence refusing turns (#375)

### LINKED ISSUE #375: desktop: support external MCP servers from mcps.yaml in the native chat runner
Since #278 removed the Go sidecar, a chat whose agent names an MCP capability that only `mcps.yaml` can resolve **answers a 500 naming the gap** (before any subprocess exists) instead of forwarding to Go, which used to read `<data dir>/mcps.yaml` via `resolveServerConfig` and pass external servers to the CLI.

`native/chat/runner.rs::mcp_plan` still refuses:
1. any `capabilities.mcp` name with no integration row (Go would have consulted `mcps.yaml`)
2. any `mcp` capability at all when `<data dir>/mcps.yaml` exists on disk (a name could resolve to a different server in each implementation — that concern is moot now that there is one implementation)

Work:
- parse `mcps.yaml` (`MCPS_FILE` override included) into the SDK's external-server config types (`McpStdioServer` etc. already exist in `claude/options.rs`)
- resolve `capabilities.mcp` names registry-first/yaml-first exactly as Go's `resolveServerConfig` ordered it, and drop the blanket "mcps.yaml exists" refusal
- an unresolvable name keeps answering an error before spawn

Until then the desktop app supports integration-backed MCP tools (all six ported types) and the local tools server, but not hand-configured external MCP servers. The desktop app has no UI for `mcps.yaml`, so the gap only affects users who authored one for `agento web` and point the desktop app at the same data dir.

Branch: target `desktop`.

---
## WHAT
**Summary:** Read `<data dir>/mcps.yaml` in the native chat runner so hand-configured external MCP servers resolve — and, first and separately, drop the blanket refusal that makes the *mere presence* of that file disable every MCP-backed agent, including integration-backed ones that work perfectly.

**Scope:**
- **Part A (the regression, small):** remove the `mcps_yaml_present` early `Err` in `native/chat/runner.rs::mcp_plan`. Its stated reason — "a name could resolve to a different server in each implementation" — describes a two-implementation world that ended with #391. Today an unrelated file on disk breaks working agents.
- **Part B (the feature):** parse `mcps.yaml` into the SDK's external-server config types and resolve `capabilities.mcp` names in `resolveServerConfig`'s order (yaml registry first, then integrations), passing external servers through `Options::mcp_servers`.
- Keep the existing refusals that are still correct: a name resolving to neither the yaml registry nor a hostable integration still refuses the turn **before** any subprocess exists, and a `whatsapp` name still refuses.
- Honour the `MCPS_FILE` environment override alongside the default `<data dir>/mcps.yaml`.

**Affected Components:**
- `src-tauri/src/native/chat/runner.rs` — `mcp_plan` (line ~418), `mcps_yaml_exists` (~462), `start_integration_servers` (~487), and the module header, which documents all three refusals at length.
- `src-tauri/src/claude/options.rs` — `McpStdioServer` (81), `McpHttpServer` (97), `McpSseServer` (107) already exist and serialize correctly; `Options::mcp_servers` is the `BTreeMap<String, Value>` they go into. No new types expected.
- `src-tauri/src/native/chat/turn.rs`, `src-tauri/src/native/chat/mod.rs` — module headers state "`mcps.yaml` is not read"; both need updating.
- `src-tauri/src/native/schedule/executor.rs` — a scheduled run shares `build_options`, so both parts change its behaviour too.
- `src-tauri/Cargo.toml` — **no YAML parser is currently a dependency**; one must be added.

**Acceptance Criteria:**
- [ ] With an `mcps.yaml` on disk and an agent whose `capabilities.mcp` names only hosted **integrations**, the turn runs — today it is refused with a 500. This is the whole of Part A and is independently shippable.
- [ ] A `capabilities.mcp` name defined in `mcps.yaml` as a stdio server reaches the CLI in `--mcp-config` with the command, args and env the file declares.
- [ ] The same for the `http` and `sse` shapes.
- [ ] Resolution order matches `resolveServerConfig`: a name present in **both** `mcps.yaml` and the integrations resolves to the yaml entry.
- [ ] A name in neither still refuses the turn before spawn — a 500 naming the gap in chat, a failed `job_history` row in a scheduled run.
- [ ] A malformed or unreadable `mcps.yaml` refuses the turn with a message naming the file, rather than being silently treated as absent.
- [ ] `MCPS_FILE` overrides the default location.
- [ ] `cargo test` green, including a `tests/chat_turn.rs` case driving a yaml-declared server through the fake CLI.

**Out of Scope:**
- Any UI for authoring or editing `mcps.yaml` — the desktop app has none and this issue does not add one.
- Hosting, health-checking or lifecycle management of an external server; it is somebody else's subprocess, exactly as `claude/mcp.rs`'s header says.
- The `whatsapp` refusal (#273 — dropped, not deferred).

## WHY
**Business Value:** Two different things, and Part A is the one that matters most. A leftover `mcps.yaml` — the normal state for anyone who ran `agento web` against the same data directory — currently disables **every** MCP-backed agent in the desktop app, including the six integration types that are fully implemented and would work. The user sees a 500 mentioning a file they may not remember writing, and nothing in the UI connects the two.

**Problem Statement:** `mcp_plan` refuses in two shapes that are now wrong for different reasons. The blanket `mcps_yaml_present` check was written to avoid a name resolving differently in the Go server than in the Rust port — a hazard that required both to exist. The per-name refusal ("no integration row") is honest but leaves external servers unsupported, so an agent configured for `agento web` cannot run here at all.

**Expected Impact:** Part A restores integration-backed agents for affected users at near-zero risk. Part B closes the last functional gap between the retired server and the desktop app for MCP tooling. No change for the common case: an agent naming no MCP server never touches any of this.

**Context:**
- **The report's premise holds and its own reasoning already anticipates the split.** It states the two-implementation concern is "moot now that there is one implementation" — that is Part A, and it is worth landing on its own rather than waiting behind the yaml parser.
- **Two corrections from the current tree.** The body says a refused turn "answers a 500 naming the gap … instead of forwarding to Go"; there is no forwarding anywhere any more (#391 deleted the Go tree), so the 500 is simply the answer. And `Branch: target `desktop`` is obsolete — PRs target `main` since #392.
- **One thing the body understates:** it frames the gap as affecting "only users who authored [an `mcps.yaml`] for `agento web`". True for Part B; for Part A the file need not be *used* at all — its existence alone is the trigger, and it breaks agents that have nothing to do with it.
- **A dependency decision this issue must make:** the crate tree has no YAML parser. `serde_yaml` is unmaintained (archived by its author); `serde_yml` and `serde_norway` are the live forks. Pick one deliberately and record why — this is a supply-chain addition to a desktop app, not a detail.
- `CLAUDE.md` documents the refusal under *What the runner still refuses* and *Phase 5*; both need updating.

## HOW
**Technical Approach:** Land it as two commits (or two PRs) in order. First delete the `mcps_yaml_present` arm and its `mcps_yaml_exists` helper, updating the four module headers that explain it — a small, self-contained behaviour restoration. Then add the registry: a `native/chat/mcps_yaml.rs` that loads and deserializes the file into the existing SDK config types, consulted by `mcp_plan` ahead of `registry::can_host`, with external servers carried on the plan alongside the integration specs and merged into `Options::mcp_servers` in `start_integration_servers`.

**Implementation Details:**
- `mcp_plan` gains a resolved registry parameter rather than reading the file itself — the function already takes `db_path` and `mcps_yaml_present` as parameters precisely so the decision can be driven over a scratch fixture, and that property must survive.
- `McpPlan` grows an `external: BTreeMap<String, serde_json::Value>` beside `servers`. A `BTreeMap` for the reason `options.rs:322` already gives: `--mcp-config` carries this map and an unordered one makes the argument differ run to run.
- `start_integration_servers` inserts the external entries into `opts.mcp_servers` and appends their qualified `mcp__<id>__<tool>` names to `allowed`, using the same allowlist rule the integration path uses — an empty `tools` list means every tool.
- Match the reference `mcps.yaml` schema field for field; transcribe it from `LoadMCPRegistry`'s struct tags rather than inferring it from an example file, and pin the mapping with a fixture.
- **One deliberate divergence to preserve:** `start_integration_servers`' header records that a start failure refuses the turn where Go discarded the error and ran degraded. Keep that rule for external servers — an unparseable registry entry refuses rather than being skipped.
- `native/chat/runner.rs`, `turn.rs`, `mod.rs` and `schedule/executor.rs` headers all describe the current refusals; update each.

**Testing Strategy:**
- Unit, in `runner.rs`: `mcp_plan` over a scratch database and an in-memory registry — a yaml-only name, an integration-only name, a name in both (yaml wins), a name in neither (refused), and the Part A case of an agent naming only integrations with a registry present.
- Parser: a fixture `mcps.yaml` covering stdio/http/sse, asserted against the exact `--mcp-config` JSON produced, so a field rename is caught at the bytes.
- Integration, `tests/chat_turn.rs`: a turn whose agent names a yaml stdio server, driven through the fake CLI, asserting the `--mcp-config` argument.
- A test that a malformed file refuses with a message naming the path.

- `CLAUDE.md` — *What the runner still refuses* and the Phase 5 section both list the `mcps.yaml` refusals; rewrite once Part B lands, and after Part A note that only unresolvable names refuse.
- `docs/user-guide.md` — a short note that `mcps.yaml` in the data directory is read, given there is no UI for it.
- New dependency: one YAML parser, justified in the PR description.

**Files to Create / Modify:**
- Create: `src-tauri/src/native/chat/mcps_yaml.rs`, a test fixture `mcps.yaml`.
- Modify: `src-tauri/src/native/chat/runner.rs`, `turn.rs`, `mod.rs`, `src-tauri/src/native/schedule/executor.rs`, `src-tauri/Cargo.toml`, `src-tauri/Cargo.lock`, `src-tauri/tests/chat_turn.rs`, `CLAUDE.md`, `docs/user-guide.md`.

**Risks:**
- *A new YAML dependency in a shipped desktop app* → pick a maintained fork, state the choice in the PR, and keep the parse behind one module with no other caller.
- *An external server is an arbitrary subprocess with the user's environment* → that is what the file already means and what the CLI already does with it; no new privilege, but the docs note should say so plainly.
- *Silently running a turn with a tool set that failed to start* → explicitly refused; this is the divergence from Go the runner already chose, applied to the new path.
- *Part B slipping and Part A never shipping* → land Part A first; it needs no parser and no new dependency.

**Complexity:** Medium (Part A: Low, one file plus headers · Part B: Medium, a new module, a new dependency and multi-file wiring)

---
**Refined by:** AI Issue Refiner · **Date:** 2026-08-27 · **Status:** Ready

vibexp-agent:
Closes #375

## What

Two things, landed together because the second is what makes the first
complete — but the first is a **regression fix** and stands on its own.

**Part A — the regression.** `native/chat/runner.rs::mcp_plan` refused *any*
`capabilities.mcp` when `<data dir>/mcps.yaml` merely **existed** on disk. That
check guarded against a name resolving to a different server in the Go server
than in this port — a hazard that needs both to exist, and #391 deleted the Go
tree. Until now, one leftover file (the normal state for anyone who ever ran
`agento web` against the same data directory) disabled **every** MCP-backed
agent on the machine, including the six integration types that are fully
implemented here and would have worked. The file did not have to be *used*; its
existence alone was the trigger.

**Part B — the feature.** New `native/chat/mcps_yaml.rs` ports
`LoadMCPRegistry` / `convertMCPEntry` / `interpolateEnv`. `mcp_plan` now walks
`capabilities.mcp` in `resolveServerConfig`'s order — **registry first, then the
integrations** — so an external server reaches the CLI in `--mcp-config` with the
command, args, env or headers the file declares, and a name in *both* is the
yaml entry.

## Why the Go source is quoted rather than inferred

`internal/config/mcp.go` was deleted with the rest of the Go tree, so the schema
was recovered from git history (`git show daa8e69^:internal/config/mcp.go`) and
transcribed field for field rather than guessed from an example file, as the
issue's HOW asked. Everything below that reads as a quirk is Go's:

- the `transport` key and the `type` on the wire are **different spellings** and
  only `stdio` and `sse` agree — `streamable_http` becomes `http`;
- an `${ENV:VAR}` that is **set but empty** is the same refusal as an unset one
  (Go tests `value == ""`), and an unterminated `${ENV:` is left alone;
- a missing file is an empty registry with no error, while an unreadable or
  unparseable one is an error naming the path.

## What is still refused

A `capabilities.mcp` name that resolves to **neither** the registry nor a
hostable integration — which is where `whatsapp` lands, the type being dropped
rather than deferred (#273). A malformed registry refuses too, rather than
reading as "no external servers" and silently dropping a tool set: that is
`start_integration_servers`' existing divergence from Go (which discards the
error and runs degraded) applied to the new path. Both refusals happen **before
any subprocess exists** — a 500 carrying the reason in a chat, a failed
`job_history` row in a scheduled run.

## Two ordering rules that keep Part A's fix from being undone

- **`mcp_plan` takes the registry's *path*, not a loaded registry**, and loads it
  only after establishing that the agent names at least one MCP server. Loading
  in the caller would make a broken file refuse turns that never consult it —
  the exact shape of the bug Part A removes. (This is the one place the
  implementation departs from the HOW, which suggested a pre-resolved registry
  parameter; the stated property — "the decision can be driven over a scratch
  fixture" — is preserved, since a path is as injectable as `db_path` already is.)
- **The database is opened only for a name the registry did not answer**, so an
  all-external agent reads no `integrations` rows at all, matching
  `resolveServerConfig`'s short-circuit. `McpPlan.db_path` is `Option`, which is
  what makes "this turn read no database" an assertion rather than an assumption.

`MCPS_FILE` overrides the path. Go's `web` path had no override (the
`--mcps-file` flag was `agento ask`'s), so this is the desktop equivalent of
that flag — and the only way to point a test at a fixture, because a debug
build's `paths::data_dir` ignores the environment by design.

## The new dependency

`serde_norway = "0.9"` — **2 crates** added to the lock (`serde_norway`,
`unsafe-libyaml-norway`), one caller, justified in a comment at the site.

`serde_yaml` was archived by its author, so the choice is between its three live
forks. `serde_yml` has the most downloads and is rejected anyway — its maintainer
relicensed and bulk-published generated churn over the fork, which is precisely
the supply-chain shape a shipped desktop app should not take on. `serde_yaml_ng`
is credible but MIT-only and last published May 2024. `serde_norway` is dual
MIT/Apache-2.0 like the rest of this tree, was published more recently (Dec
2024), carries its own maintained `unsafe-libyaml-norway` rather than depending
on the archived parser, and is what the Nix/Zellij ecosystem settled on.

## One deliberate divergence from Go, documented at its site

Go's `interpolateEnv` rescans the whole string from the start after each
substitution, so `FOO=${ENV:FOO}` **spins forever** — in Go too. A hang inside
`build_options` is a chat that never answers and a scheduled run that never
records a row, so the loop is bounded at 1000 substitutions and the bound is an
error. Every terminating input behaves exactly as Go's does.

## Tests

Seven new, and **every one was verified to fail on the revert** (the blanket
`mcps_yaml_present` arm restored → 5 unit + 1 integration go red):

- `mcps_yaml.rs`: the three transports pinned as whole `--mcp-config`
  documents (an empty `env`/`headers` is *absent*, not `{}`); `null` as the zero
  value at the field and one level down (`args:` with nothing under it is how a
  person actually leaves it empty); an empty/comment-only file; a missing file
  vs a broken one; the unknown-transport sentence; `${ENV:…}` including the
  set-but-empty and unterminated cases; the self-reference bound; `MCPS_FILE`.
- `runner.rs`: **Part A's own guard** — a registry naming none of the agent's
  servers does not refuse it, over three file shapes including an empty one; a
  yaml name resolving externally with `db_path: None`, proving no database was
  read; a name in both resolving to the yaml entry; an unreadable registry
  refusing *and* not refusing an agent that names no MCP server; an external
  server registered without binding a listener, walked in the same allowlist
  pass as an integration.
- `tests/chat_turn.rs`: the end-to-end criterion — an agent whose server exists
  only in `mcps.yaml` runs, and the fake CLI reports back the `--mcp-config`
  entry, the `--allowedTools` value and `--strict-mcp-config` from its own argv.

## Acceptance criteria

- [x] An `mcps.yaml` on disk + an agent naming only hosted integrations → the turn runs (Part A)
- [x] A stdio server from the file reaches `--mcp-config` with its command, args and env
- [x] The same for `http` and `sse`
- [x] A name in both resolves to the yaml entry
- [x] A name in neither refuses before spawn
- [x] A malformed/unreadable file refuses with a message naming it
- [x] `MCPS_FILE` overrides the default location
- [x] `cargo test` green, including a `chat_turn.rs` case through the fake CLI

## Local gates

`npm run build` · `cargo fmt --check` · `cargo clippy --all-targets -- -D warnings` ·
`cargo test` (1442 lib + all integration suites) · `cargo build --release` — all green.

## Docs

`CLAUDE.md` — *What the runner still refuses* rewritten as *how the runner
resolves an MCP name*, plus the executor's copy of the same claim and the layout
tree. `docs/user-guide.md` — a new **External MCP servers (`mcps.yaml`)** section
under Agents → Tools, since there is no UI for the file, stating plainly that a
`stdio` entry is a program Agento does not sandbox or supervise.

## Out of scope

No UI for authoring `mcps.yaml`; no hosting or lifecycle management of an
external server; the `whatsapp` drop (#273) is untouched.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -294,8 +294,12 @@ src-tauri/src/
     chat/        the SSE turn and the three routes that steer it (#276)
       live.rs    the process-local live-session registry — why the four move together
       runner.rs  an agent's config as SDK options; starts the local tools
-                 server (#310) and one per github integration (#311), and
-                 refuses only what it still cannot supply
+                 server (#310) and one per github integration (#311), passes
+                 an mcps.yaml server through (#375), and refuses only a name
+                 that resolves to neither
+      mcps_yaml.rs `<data dir>/mcps.yaml` — LoadMCPRegistry, the three
+                 transports and the `${ENV:…}` substitution (#375). The one
+                 YAML this app reads and the only caller of `serde_norway`
       turn.rs    spawn, stream, and the AskUserQuestion continuation
       persist.rs what a finished turn writes, and what an interrupted one does not
       sse.rs     the frame bytes: raw pass-through vs the two synthetic events
@@ -2160,20 +2164,42 @@ means a model mid-`create_issue` when the user hits Save is the ordinary case.
 `a_tool_call_in_flight_survives_the_handles_drop` pins it, and it fails with the
 two lines swapped.
 
-**What the runner still refuses.** `chat/runner.rs::build_options` serves an
-agent whose `capabilities.mcp` names hosted integrations, starting one filtered
-server per name and handing the handles to the turn beside the local-tools one.
-Three things it cannot serve, and each is a *refusal* rather than a fallback —
-a chat reports it, and a scheduled run records a failed `job_history` row: a
-name whose row is a type this build does not host (i.e. `whatsapp`), a name
-with **no** integration row at all, and — broader than it strictly needs to be
-— *any* `mcp` capability when `<data dir>/mcps.yaml` exists, because the
-original resolution order consults that file **before** the integrations and
-this build reads none. The server is registered under the **bare integration id**,
-not `github::server_name`'s `github-<id>`: the latter is `mcp.NewServer`'s
-implementation name and never appears on a tool, while the id is the
-`mcpServers` key and so the prefix on every qualified name already in an agent's
+**How the runner resolves an MCP name, and the one thing it still refuses**
+(#375). `chat/runner.rs::mcp_plan` walks `capabilities.mcp` in
+`resolveServerConfig`'s order: **the `mcps.yaml` registry first**
+(`native/chat/mcps_yaml.rs`), then the integrations. A registry hit is an
+*external* server — somebody else's subprocess or URL, handed to the CLI in
+`--mcp-config` with nothing bound here; an integration is a filtered in-process
+server started per turn. Either is registered under **the name the agent
+wrote** — the bare integration id, not `github::server_name`'s `github-<id>`:
+the latter is `mcp.NewServer`'s implementation name and never appears on a tool,
+while the key is the prefix on every qualified name already in an agent's
 allowlist.
+
+The order is Go's and it decides a real case: **a name in both is the yaml
+entry**, so a user who shadows an integration id in their own file gets their
+own server. One shape is refused, and it is a *refusal* rather than a fallback —
+a chat reports it, a scheduled run records a failed `job_history` row: **a name
+in neither**, which `whatsapp` reaches by construction since the type is dropped
+(#273). A malformed or unreadable `mcps.yaml` refuses too, rather than reading
+as "no external servers" and silently dropping a tool set.
+
+**What #375 removed is the check that was broader than it needed to be:** the
+mere *presence* of `<data dir>/mcps.yaml` used to refuse every `mcp` capability,
+including names that resolved perfectly to hosted integrations. It was guarding
+against a name resolving differently in the Go server than here — a hazard that
+needed both to exist, and #391 deleted the Go tree. Until it went, one leftover
+file (the normal state for anyone who ever ran `agento web` against the same
+data directory) disabled every MCP-backed agent on the machine.
+
+Two ordering rules keep that fix from being undone by accident. `mcp_plan` takes
+the registry's **path**, not a loaded registry, and loads it only after
+establishing that the agent names at least one MCP server — so a typo in the
+file cannot refuse a turn that never consults it. And the database is opened
+only for a name the registry did not answer, so an all-external agent reads no
+`integrations` rows at all. `MCPS_FILE` overrides the path, which is the only
+way to point a test at a fixture: a debug build's `paths::data_dir` ignores the
+environment by design.
 
 **Reading a credential is this module's job and nobody else's.** The rule in
 `native/integrations.rs` — `credentials` is never selected, `auth` collapses to
@@ -2992,7 +3018,9 @@ database".
 `Err` and let Go answer" is not available, because there is no request. So every
 path ends in a `job_history` row. That includes the one case Go has no
 equivalent for: `chat/runner.rs::build_options` can still refuse an agent whose
-tools this build cannot host (a name in `mcps.yaml`, or a `whatsapp` row), and
+tools this build cannot supply (a `capabilities.mcp` name that neither
+`mcps.yaml` nor a hostable integration resolves — a `whatsapp` row reaches that
+by construction — or one whose `mcps.yaml` could not be read), and
 where a chat forwards, a scheduled run records a **failed** row reading
 `agent tools unavailable in this build: …` and publishes the failed event.
 Silence is the one outcome that is not allowed, because a job history with no row
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -2202,18 +2202,41 @@ unprefixed `MCPS_FILE` honoured behind it, because #375's acceptance criteria
 named that spelling), which is the only way to point a test at a fixture: a
 debug build's `paths::data_dir` ignores the environment by design.
 
-Three things in `mcps_yaml.rs` were **measured against `gopkg.in/yaml.v3`**
-rather than inferred from the JSON side, and each is wrong in a way nothing
-would report. A null *sequence element* is **dropped** where a null *map value*
-is `""` — so `args` is not a `GoList`, because `["--f", ~]` is
-`docs-mcp --f` and not `docs-mcp --f ""`. Merge keys (`<<: *anchor`) are
-expanded by `yaml.v3` during decode and need `Value::apply_merge` here, or a
-perfectly valid file refuses every MCP-backed agent with *"unknown transport"*.
-And **no decode failure quotes what it was decoding**: this file holds `Bearer`
-tokens and a refusal's text reaches a chat body, `job_history.error_message` and
-the exported log, so a syntax error reports its position and a shape error
-reports the server name — `native/integrations/registry.rs`'s rule, same
-reasoning.
+**Four things in `mcps_yaml.rs` were measured against `gopkg.in/yaml.v3` v3.0.1
+rather than inferred from the JSON side**, and each is wrong in a way nothing
+would report. Measure before changing any of them; the JSON intuition is wrong
+about three.
+
+- **Any scalar is a string.** `d.scalar` fills a `string` field from the node's
+  own text whatever it resolved to, so `env: {PORT: 8080}`, `command: 123` and
+  `DEBUG: true` all decode. `serde` type-checks and rejects them — which is
+  *Part A's regression re-entered through the parser*, because `mcp_plan` loads
+  the registry for **any** agent naming any MCP server, so one unquoted port
+  number would refuse every MCP-backed agent on the machine including ones with
+  no entry in the file. `YamlString` is the visitor that accepts them.
+- **A null *sequence element* is dropped where a null *map value* is `""`.** So
+  `args` is deliberately **not** a `GoList`: `["--f", ~]` is `docs-mcp --f`, not
+  `docs-mcp --f ""`. `env`/`headers` keep `GoMap`, which is right for maps.
+- **Merge keys (`<<: *anchor`) are expanded during decode** and need
+  `Value::apply_merge` here, or a perfectly valid file refuses every MCP-backed
+  agent with *"unknown transport"*.
+- **A duplicate server name is an error in Go and last-one-wins here** — the one
+  divergence left standing, as [#499](https://github.com/shaharia-lab/agento/issues/499),
+  because no `serde_yaml` fork exposes a flag for it.
+
+Two rules about what a failure here may *say*, both because this file holds
+credentials and a refusal's text reaches a chat body, `job_history.error_message`
+and the exported app log. **No decode failure quotes what it was decoding**: a
+syntax error reports its position and a shape error reports the server name —
+`native/integrations/registry.rs`'s rule, same reasoning. And **`McpSource`'s
+`Debug` is hand-written to withhold an external config**, because `${ENV:…}` is
+resolved at load, so the plan holds the live token the *file never contained*;
+`McpPlan` is formatted in every panic message in that module.
+
+The accepted cost of `apply_merge` is that a scalar is resolved before any field
+reads it, so a spelling that does not survive a `Value` round trip does not
+survive here: `1.50` is `"1.5"` and `0x10` is `"16"` where Go keeps the raw text.
+Integers, booleans and strings are exact. Pinned, not reconciled.
 
 **Reading a credential is this module's job and nobody else's.** The rule in
 `native/integrations.rs` — `credentials` is never selected, `auth` collapses to
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -2197,9 +2197,23 @@ the registry's **path**, not a loaded registry, and loads it only after
 establishing that the agent names at least one MCP server — so a typo in the
 file cannot refuse a turn that never consults it. And the database is opened
 only for a name the registry did not answer, so an all-external agent reads no
-`integrations` rows at all. `MCPS_FILE` overrides the path, which is the only
-way to point a test at a fixture: a debug build's `paths::data_dir` ignores the
-environment by design.
+`integrations` rows at all. `AGENTO_MCPS_FILE` overrides the path (with the
+unprefixed `MCPS_FILE` honoured behind it, because #375's acceptance criteria
+named that spelling), which is the only way to point a test at a fixture: a
+debug build's `paths::data_dir` ignores the environment by design.
+
+Three things in `mcps_yaml.rs` were **measured against `gopkg.in/yaml.v3`**
+rather than inferred from the JSON side, and each is wrong in a way nothing
+would report. A null *sequence element* is **dropped** where a null *map value*
+is `""` — so `args` is not a `GoList`, because `["--f", ~]` is
+`docs-mcp --f` and not `docs-mcp --f ""`. Merge keys (`<<: *anchor`) are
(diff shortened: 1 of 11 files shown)


vibexp-agent:
Round 1 review applied. All four decode claims were re-measured against real `gopkg.in/yaml.v3` v3.0.1 before acting — all four confirmed.

| # | Outcome |
|---|---|
| 1 · serde error quotes the offending scalar | **Fixed.** Syntax errors report line/column, shape errors report the server name. Neither quotes a value. |
| 2 · `GoList` zero-fills a null sequence element | **Fixed.** `yaml.v3` *drops* it (measured), so `args` is `Vec<Option<String>>` + `flatten`. `env`/`headers` keep `GoMap` — measurement confirms that half was right. |
| 3 · merge keys silently ignored | **Fixed.** `Value::apply_merge()` before the entry decode; the anchor's own entry stays a server, as it does in Go. |
| 4 · `MCPS_FILE` set outside `env_lock` | **Fixed.** `run_turn_on` takes the registry path and sets/removes it inside the guard. |
| 5 · unprefixed env var | **Fixed, with one deviation.** `AGENTO_MCPS_FILE` is the name and wins; the unprefixed `MCPS_FILE` is honoured behind it, because #375's acceptance criteria name that spelling. Documented at the site. |
| 6 · duplicate server name | **Ticketed as #499**, linked both ways. Needs a custom `Deserialize` — no serde_yaml fork exposes a flag. |

Also fixed the cosmetic note: `tests/scheduled_run.rs:278`'s comment.

Three new tests (`a_merge_key_is_expanded_the_way_yaml_v3_expands_it`, `a_decode_failure_never_echoes_the_value_it_failed_on`, and the corrected null-element assertion), each verified to fail on its own revert. Local gates green again: fmt, `clippy --all-targets -D warnings`, 1444 lib tests + every integration suite, `npm run build`, `cargo build --release`.

vibexp-agent:
Round 2 applied. Both decode claims re-measured against real `gopkg.in/yaml.v3` v3.0.1 before acting — both confirmed.

| # | Outcome |
|---|---|
| 1 · bare scalars rejected | **Fixed**, and the review is right that this was Part A's regression re-entered through the parser. Measured: `env: {PORT: 8080}`, `{DEBUG: true}`, `{RATE: 1.50}`, `args: [--port, 8080]`, `command: 123` all decode in Go with no error. `YamlString` accepts any scalar. Residual divergence pinned per the review's own suggestion: `1.50`→`"1.5"`, `0x10`→`"16"` (`apply_merge` forces the `Value` round trip); integers, booleans and strings are exact. |
| 2 · `Debug` carries a resolved token | **Fixed.** Hand-written `Debug` printing `External(..)`; the comment that licensed the derive is corrected to say the opposite of what it said. |
| 3 · substitution-bound error quotes the value | **Fixed**, and the wording no longer misreports 1001 legitimate placeholders as a self-reference. |
| 4 · user-guide blast radius | **Fixed** — a broken registry stops every agent naming *any* MCP server, not only those with an entry in it. |
| 5 · merge-key failure has no position | **Taken** rather than left as polish, but not by forwarding the serde message — that is the rule finding 1 of round 1 established. `<<` is the only thing `apply_merge` can fail on, so it gets its own fixed sentence naming it. |

Two new tests — `a_bare_scalar_is_read_as_its_text_the_way_yaml_v3_reads_one` (every row measured against Go, including the two divergent ones) and `the_plans_debug_does_not_carry_an_interpolated_secret` — each verified to fail on its own revert. Note the first revert-check needed care: removing `visit_i64` alone did not break it, because serde_norway routes positive integers through `visit_u64`.

Local gates green: fmt, `clippy --all-targets -D warnings`, 1446 lib tests + every integration suite, `npm run build`, `cargo build --release`. `CLAUDE.md` now carries all four measured yaml.v3 rules and the two rules about what a failure here may say.

vibexp-agent:
## Not auto-merged — handing off

CI is green on `578fdaa` (8/8, including Windows unit tests and all five CodeQL jobs) and every local gate passes, but **this run will not merge it**, for one reason: the automated review ran its full budget of **3 rounds and never returned APPROVE**. Round 3's verdict was CHANGES_REQUESTED over a single 🟠 finding, which is applied in `578fdaa` — so the last commit on this branch is the one change no review has seen.

That is exactly the case the auto-merge boundary reserves for a human, so it stays here rather than going in unread.

### Review history

| Round | Verdict | Findings | Outcome |
|---|---|---|---|
| 1 | CHANGES_REQUESTED | 6 | 5 fixed in `aa0a6ac`; 1 ticketed as #499 |
| 2 | CHANGES_REQUESTED | 5 | all 5 fixed in `10fbb7c` |
| 3 | CHANGES_REQUESTED | 1 | fixed in `578fdaa` — **not re-reviewed** |

Every finding across the three rounds was latent rather than live — no shipped behaviour was wrong, and rounds 2 and 3 each found one thing the previous round had missed one struct away. Round 3 independently re-measured every decode claim against real `gopkg.in/yaml.v3` v3.0.1 and recovered the deleted Go source itself rather than trusting this PR's transcription; it found the port faithful.

### What a reviewer should look at first

`578fdaa` alone — one `impl Debug for Registry` printing `servers.keys()` instead of `servers` (the map holds post-interpolation `Authorization` headers), plus one assertion folded into the existing secret test, verified to fail on the revert.

### State

- Branch `feat/issue-375-mcps-yaml-registry`, 4 commits, ready for review.
- Local gates on `578fdaa`: `cargo fmt --check`, `cargo clippy --all-targets -- -D warnings`, 1446 lib tests + every integration suite, `npm run build`, `cargo build --release` — all green.
- Deferred: #499 (duplicate server name in `mcps.yaml` is last-one-wins here, an error in Go), cross-linked both ways.

Issue #517, opened 2026-08-30
Integrations: the connect and edit screens present the same fields under different headings, order and help text


vibexp-agent:
## What

The connect screen and the edit screen for the same integration present the same fields under different headings, in a different order, with different help text.

| | connect (`ConnectForm`) | edit (`IntegrationDetail`) |
|---|---|---|
| order | About → Connection → Services and tools | Authorisation → Triggers → Webhook → **Settings** → Services and tools → Credentials |
| **Name** lives under | *Connection* | *Settings* |
| Name help text | "How this connection is labelled in Agento…" | *(none)* |
| credentials live under | *Connection*, under Name | *Credentials*, at the bottom |
| Services help | "Only the tools you leave on are exposed to agents. You can change this later." | *(none)* |
| Enabled switch | *(absent — always created enabled)* | *Settings* |

So the user learns a layout on the way in and meets a different one on the way back. Nothing about a saved integration makes its Name a "Setting" where it was a "Connection" a minute earlier, and the "only the tools you leave on are exposed" promise — the one sentence explaining what the checkbox grid means — is shown only to people who have never used the app before.

## Why

The two components were written independently (`IntegrationsView.tsx:303` and `:603`) and share only `CredentialFields` and `ServiceEditor`. Every wrapper around them drifted.

## Notes for whoever picks this up

- Depends on the credentials issue: once an omitted credential preserves the stored one, the *Credentials* section folds back into the connection block and the ordering question becomes answerable.
- The genuinely edit-only sections (Authorisation, Triggers, Webhook, the Enabled switch) should sit *after* the shared block rather than before it, so both screens open on the same thing.
- Section titles and help text should come from one place rather than being typed twice.

---
## WHAT

**Summary:** Make the connect screen and the edit screen for one integration present the same fields under the same headings, in the same order, with the same help text — by rendering the shared part from one component instead of two hand-written copies.

**Scope:**
- One shared "connection" block (Name + credentials + auth method) rendered by both `ConnectForm` and `IntegrationDetail`.
- One shared "services and tools" block, keeping its explanatory help line on both screens.
- Edit-only sections (Authorisation, Trigger rules, Webhook, the Enabled switch) move **after** the shared block, so both screens open on the same thing.
- Section titles and help strings live in one place rather than being typed twice.

**Affected Components:**
[...]

PR #526, opened 2026-08-31
fix(ui): one action grammar for every form, and no + on a submit (#516)

### LINKED ISSUE #516: Integrations: the connect form's submit is "+ Create" with a plus icon, and the primary action moves between the toolbar and the savebar
Reported from the app with a screenshot of the *Connect GitHub* screen.

## What

Two problems with the same button.

**1. It is labelled `+ Create` with a plus icon.** Everywhere else in Agento the `+` prefix marks a *list* action that opens a new blank thing — "New task", "New rule", "New profile", the new-agent button. It has never marked a form submit. On the connect screen it reads as "add another one" rather than "save this form", which is what the user reported: *"this create button with + icon can be confusing"*.

**2. It has no partner.** Tasks and Agents pair the primary action with a **Discard** (creating) / **Revert** (editing) button; the gateway pairs it with **Cancel**. The connect form has neither — once fields are typed there is no way back except selecting a different provider in the sidebar.

## Why

Compare the three views:

| view | creating | editing | where |
|---|---|---|---|
| Tasks (`TasksView.tsx:398-430`) | `Discard` + `Create` | `Revert` + `Save` | top toolbar |
| Agents (`AgentsView.tsx:690-716`) | `Discard` + `Create agent` | `Revert` + `Save` | bottom savebar |
| Gateway (`ProvidersView.tsx`, `ModelsView.tsx`) | `Cancel` + `Save` | `Revert` + `Save` | bottom savebar |
| Integrations | *(nothing)* + `+ Create` | `Revert` + `Save` | **create: top toolbar; edit: bottom savebar** |

Integrations is the only view where the primary action *moves* depending on whether the record exists — top-right on create, bottom savebar on edit — and the only one whose submit carries a `+`.

## Notes for whoever picks this up

- Pick one placement per view and apply it to both states; the savebar (`.savebar`, appearing when dirty) is the majority pattern and already exists in `IntegrationDetail`.
- Drop the `plus` icon from the submit; keep `+` for list-level "new" actions.
- Add the missing Discard/Cancel to the connect form.
- The verb pair itself should be settled repo-wide — `Create` vs `Create agent`, `Discard` vs `Cancel` — so the choice here is not a fourth dialect.

---
## WHAT

**Summary:** Give Integrations the same primary-action grammar as the rest of the app — one placement for both create and edit, a plain verb with no `+` icon, and a Discard/Cancel partner on the connect form.

**Scope:**
- Drop the `plus` icon from `ConnectForm`'s submit and rename it so it does not read as a list "add" action.
- Move the connect form's action into the same place the edit screen uses (the `.savebar`), so the primary action does not move between the two states of one view.
- Add the missing Discard/Cancel to the connect form.
- Settle the verb pair once (`Create` vs `Create agent`, `Discard` vs `Cancel`) so this becomes the fourth *user* of a pattern, not a fourth dialect.

**Affected Components:**
- `src/views/IntegrationsView.tsx` — `ConnectForm`'s toolbar button (L346–349) and the form body; `IntegrationDetail`'s savebar (L815–831) is the reference.
- `src/styles/integrations.css` / the shared `.savebar` rules — only if the savebar needs to render a create-state message.
- `src/views/TasksView.tsx`, `src/views/AgentsView.tsx`, `src/views/gateway/*View.tsx` — read-only references unless the verb decision changes their labels.

**Acceptance Criteria:**
- [ ] No submit button in the app carries a `+` icon; `+` appears only on list-level "new X" actions (`New task`, `New rule`, `New profile`).
- [ ] The connect form and the edit form put their primary action in the same place, and it does not move as the user goes from creating to editing one integration.
- [ ] The connect form offers a way back — a Discard/Cancel that clears typed fields (including any typed credential) without leaving the screen.
- [ ] The disabled/enabled logic is unchanged: the submit stays disabled until the name and required credentials are present, and shows the in-flight label while busy.
- [ ] The chosen verb pair is written down (a comment at the shared component or in `CLAUDE.md`'s conventions), and Tasks/Agents/Gateway either already match it or are changed to match in this PR.

**Out of Scope:** what the form *contains* and in what order — that is #517. No behavioural change to create or save requests.

## WHY

**Business Value:** Reported by the user directly: *"it shows a button called `+ Create`… this create button with + icon can be confusing for people."* First-run connect is the screen a new user meets first, so a misleading affordance costs most there.

**Problem Statement:** Integrations is the only view whose primary action both carries a `+` and *moves*: top-right toolbar while creating, bottom savebar while editing. It is also the only create form with no way to abandon typed input.

| view | creating | editing | placement |
|---|---|---|---|
| Tasks (`TasksView.tsx:398-430`) | `Discard` + `Create` | `Revert` + `Save` | toolbar |
| Agents (`AgentsView.tsx:690-716`) | `Discard` + `Create agent` | `Revert` + `Save` | savebar |
| Gateway (`ProvidersView.tsx`, `ModelsView.tsx`) | `Cancel` + `Save` | `Revert` + `Save` | savebar |
| **Integrations** | *(none)* + `+ Create` | `Revert` + `Save` | **toolbar, then savebar** |

**Expected Impact:** UX consistency and lower first-run confusion; no functional risk. Removes one of the three dialects the table shows.

**Context:**
- Related: #517 (same two components, layout rather than actions), #515 (removes the credential re-type that currently gates the edit-side save).
- The savebar is the majority placement (3 of 4 views) and already exists in `IntegrationDetail`, so it is the cheaper convergence point.

## HOW

**Technical Approach:** `ConnectForm` and `IntegrationDetail` are two independent components sharing only `CredentialFields` and `ServiceEditor`. Rather than editing the button in place, lift the action strip into one small component used by both, so a future divergence needs a deliberate edit. `ModelsView.tsx:534-560` already models exactly this — one savebar whose partner button is `Cancel` when creating (an `onCancel` prop is present) and `Revert` when editing.

**Implementation Details:**
- `IntegrationsView.tsx` — extract the savebar from `IntegrationDetail` (L815–831) into a local component taking `{ dirty, busy, canSave, message, onRevert, onSave, creating }`; render it from `ConnectForm` too, with the toolbar button removed.
- `ConnectForm` gains a `revert()` mirroring `IntegrationDetail::revert` (L670) — reset `name` to `provider.label`, `values` to `{}`, `services` to `emptyServices(provider, true)`, and clear the error.
- Keep the existing `ready` gate as the savebar's `canSave`, and reuse the savebar's `__text` slot for the "why can't I save" hint the edit side already has.
- If the decision is to converge on the **toolbar** instead, the same extraction applies in reverse — either is acceptable, but pick one and apply it to both components in this PR.

**Testing Strategy:** No backend surface. `npm run build` for the typecheck; visual verification through the `ui-verify` skill — connect screen, type a name, confirm the savebar appears, Discard clears, Create still creates and selects the new integration (`onCreated`).

**Documentation / Configuration:** Add the chosen button grammar to `CLAUDE.md`'s *Conventions* section (it already legislates CSS reuse and text selection but says nothing about action verbs or placement). No deps.

**Files to Create / Modify:** `src/views/IntegrationsView.tsx`; possibly `src/styles/integrations.css` and `CLAUDE.md`; other views only if the verb sweep touches them.

**Risks:**
- *Scope creep into a repo-wide verb sweep* — mitigation: the sweep is limited to the labels in the table above; anything else stays out.
- *Losing the disabled-state hint* — mitigation: reuse `savebar__text`, which already carries the reason on the edit side.

**Complexity:** Low

---
**Refined by:** AI Issue Refiner · **Date:** 2026-08-31 · **Status:** Ready

vibexp-agent:
Closes #516

## What

The Integrations connect screen submitted through a `+ Create` in the **top toolbar** while its edit screen saved from the **bottom savebar** — the only view in the app whose primary action moved between the two states of one record, and the only submit carrying a `+`, which everywhere else marks a list-level *New X* that opens a blank thing. It also had no partner: once a credential was typed there was no way back except selecting a different provider.

## How

- **`IntegrationsView.tsx`** — a local `SaveBar` rendered by *both* `ConnectForm` and `IntegrationDetail`, so the action cannot move again without a deliberate edit. The connect form gains `discard()`, which resets name, auth mode, typed credentials and service selection **in place** (it never leaves the screen — there is nowhere to go back to). Its strip is always rendered, where the edit screen's still appears only once something changed.
- **`src/lib/formVerbs.ts`** (new) — the four literals plus compile-time pins through `lib/typeAssert.ts`. `Create`/`Save` follow whether the record exists; `Discard`/`Revert` follow whether there is anything stored to restore. The in-flight label follows the primary.
- **The sweep that makes the rule true, rather than just declared**: Tasks and Agents call the shared verbs (`Create agent` → `Create`); the gateway's `Cancel` → `Discard` and its `Save` → `Create` while creating (with `Save anyway` → `Create anyway` to match); the `+` comes off the two other `Create …` submits, in `SecurityPane` and the gateway `OverviewView`.
- **`CLAUDE.md` → Conventions** — the grammar written down, which is what binds the views that do not import the component.

## Judgement calls

- **Savebar, not toolbar.** Both were allowed by the spec; the savebar is the majority placement (3 of 4 views) and already existed in `IntegrationDetail`, so converging there moved one component instead of two.
- **`Discard` + `Create`, not `Cancel` + `Save`.** All four views' create-partner already does the same thing — abandon a record that was never stored — so the choice was naming, not behaviour. `Create`/`Save` keeps the information that the record does not exist yet, which a uniform `Save` would throw away; `Cancel` reads as *leave this screen*, which the connect form's button explicitly does not do.
- **The two token buttons are in scope.** AC1 is repo-wide ("no submit button in the app carries a `+` icon"), and `Create token` / `Create gateway token` were the only other `+`-bearing submits. Every remaining `+` is a `New …`/`Add …`.

## Proof

There is no TypeScript test harness in this repo — `npm run build` is `tsc --noEmit && vite build` and that is the whole frontend gate — so the guard is the repo's own `Eq`/`Expect` idiom rather than a test file. **Reverting a label fails the build**, verified: respelling `PARTNER_DISCARD` back to `"Cancel"` gives `src/lib/formVerbs.ts(76,40): error TS2344: Type 'false' does not satisfy the constraint 'true'`.

`npm run build` green; `pre-commit run --all-files` green. No backend surface touched.

## Acceptance criteria

- [x] No submit carries a `+`; every remaining `+` is a list-level `New …`/`Add …`
- [x] Connect and edit put the primary action in the same place, and it does not move
- [x] The connect form offers a Discard that clears typed fields, including a typed credential, without leaving the screen
- [x] Disabled/enabled logic unchanged (`ready` is still the gate) and the in-flight label is shown
- [x] The verb pair is written down (`CLAUDE.md` + `lib/formVerbs.ts`) and Tasks/Agents/Gateway are changed to match

## Out of scope

Form *contents* and their order (#517). No change to any create or save request.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -525,6 +525,34 @@ drag across it means "select this row", so selectable cells would leave a word
 highlighted behind the view they opened. `::selection` is focus-aware over the
 existing `--bg-select*` tokens, matching `.window--focused`.
 
+**A form's actions are a fixed grammar, and the `+` is not part of it** (#516).
+Four rules, and each of them was broken by exactly one view before it was
+written down here:
+
+- **The primary verb follows existence**: `Create` while the record does not
+  exist yet, `Save` once it does — with the in-flight label following it
+  (`Creating…` / `Saving…`). Not `Create agent`, not `Save` for something that
+  was never stored.
+- **Its partner follows the same split**: `Discard` throws away a record that
+  was never stored, `Revert` restores one that was. Two words because they undo
+  two different things — a single `Cancel` claims to undo a save, which is why
+  the gateway's spelling of it went rather than spreading.
+- **A `+` icon marks a list-level *New X* / *Add X*** — an action that opens a
+  blank thing (`New task`, `New rule`, `New profile`, `Add fallback`). **Never a
+  submit.** `+ Create` on the Integrations connect screen is what this issue was
+  reported for: it reads as "add another one" rather than "save this form".
+- **Placement is one strip per view, used by both states.** Integrations was the
+  only view whose primary action *moved* — toolbar while creating, savebar while
+  editing. Which strip a view uses is its own choice (Tasks is the toolbar,
+  everything else is the `.savebar` at the foot of the form); moving between
+  them inside one view is not.
+
+`IntegrationsView.tsx`'s `SaveBar` is the grammar as code and the component to
+copy from; nothing forces a view to import it, so this paragraph is the
+enforcement. Note `.savebar` is declared in `styles/integrations.css` **and**
+`styles/settings.css` and is available everywhere only because `App.tsx` imports
+both views statically.
+
 **The window's origin has to be in the capability's scope, or every IPC
 command is denied — silently.** This is the single most expensive thing to
 re-discover in this shell, because it fails *only in release builds* and it
--- a/src/lib/formVerbs.ts
+++ b/src/lib/formVerbs.ts
@@ -1,0 +1,77 @@
+/* ============================================================================
+   The words on a form's two action buttons (#516).
+
+   Agento had three dialects for one gesture. Tasks said `Discard` + `Create`,
+   Agents said `Discard` + `Create agent`, the gateway said `Cancel` + `Save`,
+   and Integrations said `+ Create` with no partner at all — the plus icon that
+   was reported as confusing, because everywhere else in the app a `+` opens a
+   blank *new thing* rather than committing a form.
+
+   Two rules, and they are the whole module:
+
+   * **The submit follows existence.** `Create` while the record does not exist
+     yet, `Save` once it does — and the in-flight label follows it, so a form
+     never says "Saving…" for something that was never stored.
+   * **Its partner follows the same split.** `Discard` throws away a record
+     that was never stored; `Revert` restores one that was. Two words because
+     they undo two different things, which is why the gateway's single `Cancel`
+     went rather than spreading: it claimed to undo a save.
+
+   Why a module rather than a paragraph: the four views that speak this grammar
+   share no component, so the only thing that can hold them together is one
+   spelling of each literal plus a compile-time pin on it. There is no
+   TypeScript test harness here (`npm run build` is `tsc --noEmit && vite
+   build`, and that is the whole frontend gate), so `lib/typeAssert.ts`'s idiom
+   is what stands in for the test — respell a literal below and CI fails.
+
+   The `+` icon rule cannot be expressed here, because it is the *absence* of an
+   `<Icon name="plus" />` beside these labels. It is written down in `CLAUDE.md`
+   under *Conventions*, with the rest of the grammar.
+   ========================================================================== */
+
+import type { Eq, Expect } from "./typeAssert";
+
+/** The submit, while the record does not exist yet. */
+export const SUBMIT_CREATE = "Create";
+/** The submit, once the record exists. */
+export const SUBMIT_SAVE = "Save";
+/** The submit while a create is in flight. */
+export const SUBMIT_CREATING = "Creating…";
+/** The submit while a save is in flight. */
+export const SUBMIT_SAVING = "Saving…";
+/** The partner, while the record does not exist yet: throw the draft away. */
+export const PARTNER_DISCARD = "Discard";
+/** The partner, once the record exists: restore what is stored. */
+export const PARTNER_REVERT = "Revert";
+
+/**
+ * What the primary button reads, given the two states that decide it.
+ *
+ * `busy` is checked first deliberately: a form mid-request is neither
+ * saveable nor discardable, and the in-flight word is the only honest label
+ * for a button that is doing something.
+ */
+export function submitLabel(creating: boolean, busy: boolean): string {
+  if (busy) return creating ? SUBMIT_CREATING : SUBMIT_SAVING;
+  return creating ? SUBMIT_CREATE : SUBMIT_SAVE;
+}
+
+/** What the partner button reads. */
+export function partnerLabel(creating: boolean): string {
+  return creating ? PARTNER_DISCARD : PARTNER_REVERT;
+}
+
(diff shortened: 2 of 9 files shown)


vibexp-agent:
### Review round 1 — two findings, both applied

| # | Sev | Location | Issue | Fix |
|---|-----|----------|-------|-----|
| 1 | 🟠 | `src/views/TasksView.tsx` | `busy` there is shared with pause/resume, so threading it into the submit made the button read "Saving…" during an unrelated request | `submitLabel(creating, false)` — the disabled state already covers it. `AgentsView`'s `saving` *is* save-specific and is still passed |
| 2 | 🟡 | `ModelsView.tsx`, `ProvidersView.tsx` | `partnerLabel(true)`/`partnerLabel(false)` where `creating` is in scope and means the same thing | Read the state variable; comment records that `onCancel` is passed exactly when the record is undefined |

Applied in `4a0e8b4`, then rebased onto `293d7df` (#520 added a `modeChanged` message branch to the same savebar — kept, inside the new structure). CI green on the rebased head, 8/8.

Issue #503, opened 2026-08-28
fix(desktop): Claude Code detection misses most real installs — a GUI app never sees the user's shell PATH


vibexp-agent:
## The report

A user on macOS with Claude Code installed and working (`claude -v` → `2.1.231 (Claude Code)` in their shell) opens Agento and gets the startup banner:

> ⚠ Claude Code is not installed, so agents cannot run. Install it with `npm i -g @anthropic-ai/claude-code`, sign in with `claude`, then restart Agento.

This is not only a wrong banner. The **same lookup drives the spawn**, so their agents genuinely cannot run — `claude_executable()` (`src-tauri/src/native/chat/runner.rs:827`) calls `find_claude_cli()` and falls back to the bare name `"claude"`, which fails to resolve for exactly the same reason the detection did. The banner is honest; the lookup is too narrow.

## Root cause

`find_claude_cli()` — `src-tauri/src/lib.rs:150` — resolves the CLI from two sources only:

1. every directory in `$PATH`;
2. six hardcoded home-relative directories: `.local/bin`, `.npm-global/bin`, `.bun/bin`, `.volta/bin`, `bin`, `AppData/Roaming/npm`.

**A GUI application does not inherit the user's shell PATH.** On macOS an app launched from Finder, the Dock or Spotlight inherits launchd's environment, which is `/usr/bin:/bin:/usr/sbin:/sbin` and nothing else — no Homebrew, no nvm, no npm prefix, none of the user's `.zshrc`. On Linux a `.desktop` launch inherits the session manager's environment, which is only marginally better. So for a GUI launch, step 1 contributes essentially nothing and the whole decision rests on the six-entry list in step 2.

That list misses most of the ways Claude Code is actually installed today:

| location | how it gets there | covered? |
|---|---|---|
| `~/.local/bin/claude` | the native installer (`curl … \| bash`) | ✅ |
| `~/.claude/local/claude` | `claude migrate-installer` — the documented local install | ❌ |
| `~/.nvm/versions/node/<ver>/bin/claude` | `npm i -g` under nvm | ❌ |
| `/opt/homebrew/bin/claude`, `/usr/local/bin/claude` | Homebrew / system npm prefix on macOS | ❌ (not on launchd's PATH) |
| `~/Library/pnpm/claude` | pnpm global | ❌ |
| `~/.asdf/shims/claude`, fnm, `n` | version managers | ❌ |
| `~/.yarn/bin/claude` | yarn global | ❌ |

[...]

PR #534, opened 2026-08-31
fix: re-resolve the Claude Code CLI when the cached path stops being spawnable (#533)

### LINKED ISSUE #533: A Claude Code self-update mid-session bricks every chat until Agento is restarted
## What happened

Chats fail with:

```
WARN turn  chat "…": starting agent session failed: claude: binary not found: "/home/shaharia/.local/bin/claude"
WARN proxy POST /api/chats/…/messages 500 6ms native-stream
```

…while `claude --version` in a terminal answers `2.1.252 (Claude Code)` and the app's own startup line had reported the path as found:

```
[20:35:40][agento_lib][INFO] claude cli resolved path="/home/shaharia/.local/bin/claude" source=login-shell
```

## Root cause

`~/.local/bin/claude` is a symlink into `~/.local/share/claude/versions/<version>`. Claude Code self-updated at 21:56 and swapped that version directory; during the swap the symlink was dangling, so `execve` returned `ENOENT` and the SDK reported `binary not found`. The failing turns were at 21:48–21:49, inside that window.

`claude_cli::executable()` resolves once and caches the answer in a process-wide `OnceLock` (`src-tauri/src/claude_cli.rs`), so the running instance kept spawning the stale path for the rest of its life. Restarting the app resolved it again and everything worked.

This is not a #503 regression — the resolution order is correct and the path was correct when it was resolved. The gap is that the cached answer is never revisited.

## Why it matters

Claude Code updates itself frequently and Agento is a desktop app people leave open. Every chat and every scheduled run fails until the user happens to restart, and the failure names a path that looks perfectly fine in a terminal — so it reads as an Agento bug with no obvious action.

## Suggested direction

On a spawn failure of kind `NotFound`, re-run `claude_cli::resolve` once and retry, rather than trusting the cache. Points to settle in refinement: whether the cache is invalidated (so the banner and later turns agree), whether the retry is bounded to one attempt per turn, and whether the scheduler's executor path gets it too.
---
## WHAT
**Summary:** Re-resolve the Claude Code CLI when the cached path stops being spawnable, instead of trusting a `OnceLock` answer for the life of the process.

**Scope:**
- Make `claude_cli`'s cache invalidatable and re-resolvable, remembering the stored override so a refresh cannot silently drop it.
- Revalidate the cached path cheaply before each spawn, and re-resolve when it no longer holds.
- Bound re-resolution so a genuinely missing CLI does not run the login-shell probe on every turn.
- Report the refreshed answer through `host_info` (startup banner / Settings), so the banner and the spawn keep agreeing — the invariant #503 established.

**Affected Components:**
- `src-tauri/src/claude_cli.rs` — `CACHE` (`OnceLock<Option<Resolution>>`), `prime`, `cached`, `executable`; add invalidation + revalidation.
- `src-tauri/src/native/chat/runner.rs` — `claude_executable()` (delegates; no logic of its own, must stay that way).
- `src-tauri/src/lib.rs` — `host_info` (lines ~123–124) reads `cached()`; must surface a refreshed resolution.
- `src-tauri/src/native/chat/turn.rs` — the `Session::new` error arm that logs `starting agent session failed`.
- `src-tauri/src/claude/process.rs` / `errors.rs` — `Error::CliNotFound`, the signal a retry would branch on.

**Acceptance Criteria:**
- [ ] With the app running and the CLI already resolved, replacing `~/.local/bin/claude` with a dangling symlink and sending a chat message succeeds without restarting Agento.
- [ ] Re-resolution passes the stored `claude_executable_path` — never `None`, which is `cached()`'s ordering safety net and not a second way to read the setting.
- [ ] `AGENTO_CLAUDE_EXECUTABLE` still wins verbatim and is never stat-gated or re-resolved away (rule 1 is taken on trust by design).
- [ ] `GET host_info` (banner + Settings) reports the refreshed path and source after a recovery, not the stale one.
- [ ] The happy path adds no subprocess per turn: at most one `stat`; the login-shell probe runs only when revalidation fails.
- [ ] A CLI that is genuinely gone does not re-probe on every turn — re-resolution is rate-limited, and the failure message is unchanged.
- [ ] Scheduled runs (`schedule/executor.rs`, via the same `claude_executable()`) inherit the behaviour with no second implementation.

**Out of Scope:**
- Watching the filesystem for CLI changes, or any background re-detection timer.
- Changing the five-rule resolution order or what each rule verifies (#503).
- A "re-detect now" button in Settings.
- Bundling or managing the Claude Code install.

## WHY
**Business Value:** Claude Code updates itself often, and Agento is a desktop app people leave open for days. Today one self-update silently disables every chat and every scheduled run until someone thinks to restart — and the error names a path that works perfectly in a terminal, so it reads as an Agento bug with no discoverable action.

**Problem Statement:** `claude_cli::executable()` resolves once and stores the answer in a process-wide `OnceLock` (`src-tauri/src/claude_cli.rs:136`). `~/.local/bin/claude` is a symlink into `~/.local/share/claude/versions/<version>`; a self-update swaps that directory, so the cached path is briefly dangling and then points at a version that no longer exists. `execve` returns `ENOENT`, `process.rs:199` maps it to `Error::CliNotFound`, and every turn answers:

```
WARN turn  chat "…": starting agent session failed: claude: binary not found: "/home/shaharia/.local/bin/claude"
WARN proxy POST /api/chats/…/messages 500 6ms native-stream
```

Observed 2026-08-31: resolution at 20:35:40 (`source=login-shell`) was correct, the symlink was rewritten at 21:56, turns at 21:48–21:49 failed, and a restart at 22:00:45 resolved the same path successfully. Nothing recovers without that restart.

**Expected Impact:** Removes a whole class of "chats stopped working and I changed nothing" reports — one that is unfalsifiable from the user's side, because `which claude`, `claude --version` and the app's own startup banner all agree the CLI is fine. Cost is one `stat` per turn against a spawn that already costs a process.

**Context:**
- Not a regression: `claude_cli.rs` is byte-identical to `v1.2.0`, where the same install works. The resolution order is right; only the cache's lifetime is wrong.
- Related: #503 (the resolution this builds on), #504 (its PR).
- The header of `claude_cli.rs` rejects background priming because `cached()` would race it and the loser fills the lock **without** the stored override. Any refresh added here must not reintroduce that race — which is why the override has to be remembered rather than re-derived.

## HOW
**Technical Approach:** Keep the resolution order untouched and change only how long its answer is trusted. Replace the `OnceLock<Option<Resolution>>` with a lock holding both the resolution and the stored override it was produced with, and add a cheap pre-spawn revalidation in `executable()`: if the cached path is no longer an executable file, re-resolve once (with the remembered override) and cache the new answer. A rate limit keeps a genuinely absent CLI from paying for a login-shell probe on every turn.

**Implementation Details:**
- `src-tauri/src/claude_cli.rs`
  - `static CACHE: RwLock<Option<Cached>>` where `Cached { resolution: Option<Resolution>, stored_override: Option<String>, resolved_at: Instant }`. `prime` fills it; `cached()` keeps its "resolve with no override if nothing primed" safety net.
  - `cached()` / `prime()` return an owned `Option<Resolution>` (a clone) rather than `Option<&'static Resolution>`, since the value can now change; update `lib.rs:123–124` accordingly.
  - `executable()`: rule 1 (`AGENTO_CLAUDE_EXECUTABLE`) unchanged and first. Otherwise take the cached path and check `is_executable_file`; on failure, and if `resolved_at` is older than a small cooldown (e.g. 5s), re-run `resolve(stored_override)`, store, log one `info` line in the existing `claude cli resolved path=… source=…` shape so the log shows the recovery.
  - Reuse the existing `is_executable_file` helper — it follows symlinks, which is exactly what catches this case; do **not** run `verify()` (a `--version` spawn) on the hot path.
- Optional second layer, decide during implementation: in `native/chat/turn.rs`'s `Session::new` error arm, branch on `Error::CliNotFound`, invalidate, rebuild `options.claude_executable` and retry **once**. This covers the residual `ENOENT` a `stat` cannot see (a present file whose `#!` interpreter is missing). If it is taken, `schedule/executor.rs` needs the same, so prefer a shared helper over two copies.
- `native/chat/runner.rs::claude_executable()` stays a pure delegation — the "banner and spawn are one answer" property is structural.

**Testing Strategy:**
- Unit (`claude_cli.rs`): a primed resolution whose path is then deleted re-resolves on the next `executable()`; the re-resolution receives the stored override (assert the override path wins, not the detected one); `AGENTO_CLAUDE_EXECUTABLE` is still returned verbatim without a stat; a missing CLI does not re-resolve twice inside the cooldown (count probes).
- Integration (`src-tauri/tests/chat_turn.rs`): point the runner at a scripted fake CLI via `AGENTO_CLAUDE_EXECUTABLE`-free resolution, swap the file for a dangling symlink between two turns, and assert the second turn still streams. Assert the test fails when the revalidation is reverted.
- Manual: reproduce with `ln -sf /nonexistent ~/.local/bin/claude` against a running instance.

**Documentation / Configuration:** `CLAUDE.md` → *The one external dependency* — the "Resolution is once per launch" sentence becomes "once per launch, revalidated before each spawn"; keep the note that saving the setting takes effect at next start if that stays true. No new dependency, no new env var.

**Files to Create / Modify:**
- Modify: `src-tauri/src/claude_cli.rs`, `src-tauri/src/lib.rs`, `CLAUDE.md`; conditionally `src-tauri/src/native/chat/turn.rs`, `src-tauri/src/native/schedule/executor.rs`, `src-tauri/tests/chat_turn.rs`.
- Create: none.

**Risks:**
- *Probe storms when the CLI is genuinely gone* — every turn would pay 3s + 2s of subprocess timeouts → the cooldown on `resolved_at` is the mitigation, and is part of the acceptance criteria rather than an optimisation.
- *Losing the stored override on refresh* — the exact defect the module header warns about → the override is stored beside the resolution and asserted by a test.
- *Lock contention / poisoning on a hot path* — `RwLock` read on every turn is negligible next to a process spawn, but avoid holding the write lock across `resolve()`'s subprocesses; resolve outside the lock, then store.
- *A stale banner after recovery* — `host_info` must read the same refreshed value, or the app reports one path while spawning another, which is the failure #503 exists to prevent.

**Complexity:** Medium (2–3 files, well under 500 LOC, existing patterns)

---
**Refined by:** AI Issue Refiner · **Date:** 2026-09-01 · **Status:** Ready

vibexp-agent:
Closes #533

## What

`claude_cli::executable()` resolved the CLI once into a process-wide `OnceLock` and never revisited it. Claude Code self-updates, and the native install is `~/.local/bin/claude` symlinked into `~/.local/share/claude/versions/<version>` — a swap the update performs. For the length of that swap the symlink dangles, `execve` answers `ENOENT`, and **every chat and every scheduled run failed for the rest of the process's life**, naming a path that works perfectly in a terminal.

The walk still runs once. Its *answer* is now revalidated before every spawn.

## How

- `src-tauri/src/claude_cli.rs` — `OnceLock<Option<Resolution>>` becomes `RwLock<Option<Cached>>`, holding the resolution, **the stored override it was produced with**, and when the last refresh finished. `executable()` `stat`s the cached path through the existing `is_executable_file` (which follows symlinks, so a dangling one is exactly what it catches) and re-walks the order when that fails. `prime`/`cached` return an owned `Option<Resolution>` — the value can change, so a `&'static` into the cache would be a lie about its lifetime.
- `src-tauri/src/lib.rs` — `host_info` reads `cached()` **once** and splits it, so a refresh landing between two reads cannot report one rule's path beside another rule's name.
- `CLAUDE.md` — *The one external dependency*: "Resolution is once per launch" becomes the walk-once / revalidate-per-spawn rule and its four sub-rules.

Five properties, each an acceptance criterion rather than an optimisation:

| | |
|---|---|
| the happy path | one `stat`, no subprocess — the `--version` round trip is the expensive half and a path that stats clean is one the walk already verified |
| the override | remembered *beside* the resolution, so a refresh cannot silently demote a user's configured path to whatever detection finds |
| `AGENTO_CLAUDE_EXECUTABLE` | verbatim, never stat-gated, never refreshed away — rule 1 is taken on trust by design |
| a CLI that is gone | rate-limited to one walk per `REFRESH_COOLDOWN` (10 s), claimed under the write lock so two concurrent turns produce one walk; inside the cooldown, whatever the last walk concluded is returned untouched |
| a walk that finds nothing | **overwrites the cache with nothing**, deliberately — `cached()` is what the banner reads, and a banner claiming an install detection could not find is the #503 defect wearing a different hat. The spawn falls back to the bare name and fails with `binary not found: "claude"`, which is what a machine that never had the CLI has always reported. The price, stated here because it is the most surprising consequence: a *transient* break recovers on the next due walk (≤10 s) rather than on the next free `stat` |

The cooldown is keyed on the last **refresh**, not on when the resolution was made, so the first failure always recovers however recently the process started.

Scheduled runs inherit this with no second implementation: `executor.rs` → `agent_run::run_headless` → `runner::build_options` → `runner::claude_executable()`, which is `claude_cli::executable()` and nothing else.

## Deviation from the HOW

The HOW offered an **optional** second layer — branching on `Error::CliNotFound` in `turn.rs`/`executor.rs` and retrying once. Not taken: it covers only the residual `ENOENT` a `stat` cannot see, it is in no acceptance criterion, and it would need a shared retry helper across two more files. The reported failure is precisely the shape `is_executable_file` catches.

## Tests

- `src-tauri/tests/claude_cli_refresh.rs` (new, one test, not `#[ignore]`d) — builds the real install shape (a symlink into a versioned directory), primes with a stored override, deletes the version directory, and asserts the next `executable()` recovers on the same process. The fake `$SHELL` appends a line per invocation, so the walk count is observable: the happy path adds none, and three calls inside the cooldown with nothing installed add none either. One test per binary deliberately — the cache is process-wide and primes exactly once.
- `claude_cli.rs::the_first_refresh_is_always_due_and_the_next_one_waits` — the cooldown boundary, constructed rather than raced (`refresh_due` takes `now` as a parameter).

**Verified against three reverts**, each failing the test:

| revert | fails on |
|---|---|
| revalidation always passes | `a dangling path must be resolved again rather than spawned` |
| no cooldown | the stale path is re-walked away, so a missing CLI re-probes per turn |
| refresh passes `None` instead of the override | the recovery lands on `"claude"` rather than the configured wrapper |

## Gates

`npm run build` · `cargo fmt --check` · `cargo clippy --all-targets -- -D warnings` · `cargo test --no-fail-fast` (1503 lib + every integration binary green) · `cargo build --profile ci`.

---

## Review

Three automated review rounds. Round 1 moved the refresh off the tokio worker (`spawn_blocking`) and corrected three stale `OnceLock` claims in `CLAUDE.md`. Round 2 found that hand-off pinned by **no** test — it passed on the revert — and the guard added for it (`tests/claude_cli_offload.rs`) needed two non-obvious things to be real: `build_options` **spawned** rather than awaited inline (`block_on` drives the test body on the calling thread, leaving the single worker free), and `sleep` resolved absolutely **before** `PATH` is emptied. Reverted it reports 0 ticks across the 1500 ms walk; kept, green. Round 3 re-proved that revert independently and found the code clean on every dimension; its one remaining item was a scope error in the deferred issue's own body, now fixed.

Deferred: #535 — all three `claude_cli` suites assume no Claude Code in `/usr/local/bin`, `/opt/homebrew/bin` or `/opt/local/bin`. Pre-existing, and fixing it means changing what rule 5 scans, which #533 puts out of scope.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -3003,9 +3003,8 @@ either way is the easy mistake here.
 `claude_cli::executable()` and then the bare name — see *The one external
 dependency* for the whole order and why a `PATH` scan was never enough. The
 override is re-read per call rather than taken from the cache, and that is for
-the tests rather than the app: the cache is a `OnceLock`, so a test binary whose
-cases each point at a *different* scripted CLI would otherwise all run the
-first one's.
+the tests rather than the app: a test binary whose cases each point at a
+*different* scripted CLI would otherwise all run the first one's.
 
 ### The scan (#289)
 
@@ -3430,7 +3429,13 @@ option-building — `TurnSettings::stored` and `registry::can_host`, shared by a
 three callers, plus `runner::load`, which is the chat turn's alone. Every one is
 `open_read_only`, and a WAL reader does not wait on a writer, which is why they
 are left: the test's contention is a write lock, so it says nothing about them
-either way. A *write* added there would be a different matter. A panic *inside* a handed-off
+either way. A *write* added there would be a different matter. **So is a
+subprocess**, and #533 nearly added one: `runner::claude_executable` can now
+re-walk the CLI order, which spawns a login shell bounded at 3 s plus a
+`--version` bounded at 2 s through a `std::thread::sleep` poll loop — twice the
+hold this test was written to catch. It is `spawn_blocking`ed at that one call
+site for exactly this rule; the list above is "non-blocking reads", not
+"whatever option-building happens to do". A panic *inside* a handed-off
 section is the one thing `db::blocking` cannot make safe: the executor's `finish`
 may have written the session results and not the job row, leaving a `running`
 row nothing will finish, which is why the rule is "every path ends in a job
@@ -4363,7 +4368,7 @@ runtime, which is not a packaging change.
 (#503). `AGENTO_CLAUDE_EXECUTABLE` → the stored `claude_executable_path` setting
 (migration 38) → **`$SHELL -lic 'command -v claude'`** → the `PATH` this process
 was launched with → a list of known install locations. First hit wins; the
-answer is cached in a `OnceLock` for the process.
+answer is cached, and **revalidated before every spawn** (#533, below).
 
 - **The login-shell probe is not decoration, and it must not be "simplified"
   away.** A GUI application does not inherit the user's shell `PATH`: a macOS
@@ -4416,8 +4421,18 @@ answer is cached in a `OnceLock` for the process.
     on every turn. The slot is claimed under the write lock *before* the walk, so
     two concurrent turns produce one walk; the cooldown is keyed on the last
     **refresh**, not on when the resolution was made, so the first failure always
-    recovers however recently the process started. Inside the cooldown the stale
-    path is returned and the failure message is unchanged.
+    recovers however recently the process started. Inside the cooldown, whatever
+    the last walk concluded is returned untouched.
+  - **A walk that finds nothing overwrites the cache with nothing**, and that is
+    the deliberate half of the previous rule rather than an oversight in it. The
+    old path is not kept: `cached()` is what the banner reads, and a banner
+    claiming an install detection could not find is the #503 defect wearing a
+    different hat. The spawn falls back to the bare name and fails with
+    `binary not found: "claude"` — which is what a machine that never had the
+    CLI has always reported, so the failure a user sees is unchanged rather than
+    newly worded. The cost is that recovery from a *transient* break waits for
+    the next due walk instead of the next free `stat`, and a ≤10 s wait is the
+    right price for a banner that cannot lie.
   - **`cached()` reads the refreshed value**, so the banner and Settings report
     the recovery — #503's "the banner and the spawn are one answer", kept true
     through a refresh. `host_info` reads it **once** and splits, or a refresh
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -4394,11 +4394,36 @@ answer is cached in a `OnceLock` for the process.
   are *the worst case a user waits for a window*, not a generous ceiling. A
   pathological shell degrades to "found by a later rule", never to a window that
   does not open. Priming on a background thread was rejected: `cached()` would
-  race it, and the loser fills the `OnceLock` **without the stored override**,
+  race it, and the loser fills the cache **without the stored override**,
   so a configured path would be ignored on some launches and not others.
-- **Resolution is once per launch.** Saving the setting takes effect at the next
-  start, which the Settings help says. Re-detecting live was considered and left
-  out.
+- **The walk is once per launch; its answer is revalidated before every spawn**
+  (#533). Claude Code self-updates, and the native install is a symlink into a
+  versioned directory that the update swaps — so for the length of that swap the
+  resolved path dangles, `execve` answers `ENOENT`, and a `OnceLock` filled at
+  startup meant **every chat and every scheduled run failed until the app was
+  restarted**, naming a path that works perfectly in a terminal. `executable()`
+  therefore `stat`s the cached path (`is_executable_file`, which follows
+  symlinks and is exactly the check a dangling one fails) and re-runs the walk
+  when it no longer holds. Four rules, each of them an acceptance criterion:
+  - **The refresh is given the stored override the first walk was given**, held
+    beside the resolution in the cache rather than re-derived — losing it would
+    silently demote a user's configured path to whatever detection finds, which
+    is the same defect background priming was rejected over.
+  - **`AGENTO_CLAUDE_EXECUTABLE` is returned verbatim and is never stat-gated**,
+    because rule 1 is taken on trust by design.
+  - **Re-resolution is rate-limited** (`REFRESH_COOLDOWN`, 10 s), or a CLI that
+    is genuinely gone pays a login-shell probe plus a `--version` per candidate
+    on every turn. The slot is claimed under the write lock *before* the walk, so
+    two concurrent turns produce one walk; the cooldown is keyed on the last
+    **refresh**, not on when the resolution was made, so the first failure always
+    recovers however recently the process started. Inside the cooldown the stale
+    path is returned and the failure message is unchanged.
+  - **`cached()` reads the refreshed value**, so the banner and Settings report
+    the recovery — #503's "the banner and the spawn are one answer", kept true
+    through a refresh. `host_info` reads it **once** and splits, or a refresh
+    between two reads would report one rule's path beside another rule's name.
+  There is no filesystem watcher and no background re-detection timer. Saving
+  the setting still takes effect at the next start, which the Settings help says.
 
 ---
 
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -4409,7 +4409,7 @@ answer is cached, and **revalidated before every spawn** (#533, below).
   restarted**, naming a path that works perfectly in a terminal. `executable()`
   therefore `stat`s the cached path (`is_executable_file`, which follows
   symlinks and is exactly the check a dangling one fails) and re-runs the walk
-  when it no longer holds. Four rules, each of them an acceptance criterion:
+  when it no longer holds. Five rules, each of them an acceptance criterion:
   - **The refresh is given the stored override the first walk was given**, held
     beside the resolution in the cache rather than re-derived — losing it would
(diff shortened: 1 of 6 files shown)


vibexp-agent:
Deferred to #535 — the `claude_cli` suites assume no Claude Code in `/usr/local/bin`, `/opt/homebrew/bin` or `/opt/local/bin`. Pre-existing (`claude_cli_detection.rs` has the same exposure), and fixing it means changing what rule 5 scans, which #533 puts out of scope.

PR #549, opened 2026-09-02
fix: the Limits section says "no limit" instead of showing one (#543)

### LINKED ISSUE #543: Tasks: the Limits section shows "no limit" as a 0 and a date mask, with no way to clear an end date
## What

The *Limits* section of the Tasks form communicates "no limit" badly in both of its rows:

- **Stop at** is a bare `<input type="datetime-local">`. When unset, the browser paints its own mask (`dd/mm/yyyy, --:--`), which reads as a value rather than as "not set" — and once a date *is* set there is no control to clear it back to unset.
- **Stop after** is a number input whose "no limit" value is `0`, explained only by help text. A task that runs forever displays the same `0` as a task nobody has configured yet.

## Why

Reported from the app: *"the stop at value in the Limits should be 'not set' or something but currently it's showing me a value ... how can we reset this because user might get confused about it ... if user select anything it shows the value otherwise it should be empty."*

This is the same class of inverted-default problem `CLAUDE.md` already calls out for the gateway's retention horizon — *`0` means keep everything, and it is therefore the **longest** horizon rather than the shortest*. Every comparison and every label over such a value has to read the right way round, and a spinner sitting on `0` reads as "zero runs", the exact inverse of "unlimited".

Code: `src/views/TasksView.tsx` ~lines 583-614. `toLocalStamp` already returns `""` correctly for a null `stop_after_time` (line 104) — the value is genuinely empty and the mask is the browser's, so this is presentation, not state.

**Note the dependency:** neither of these fields is currently persisted at all — see the dropped-fields bug. Fixing the UX on a field the API discards would produce a control that looks correct and still forgets. That issue should land first, or this one will be untestable end to end.

## How

- **Stop at**: render an explicit *Not set* state rather than an empty native picker — a "Set an end date" affordance that reveals the picker, or the picker plus a clear (`×`) control that returns it to `null`. Clearing must send `stop_after_time: null`, not `""`.
- **Stop after**: make "no limit" a state rather than a magic number. Either a *Unlimited / Stop after N runs* pair where the number input only appears in the second, or a placeholder-and-empty representation that maps to `0` on the wire. The stored wire value stays `0` — this is a display change, not a schema change.
- Keep the help text honest with whatever is chosen; `"0 keeps the task running forever."` becomes wrong the moment `0` stops being shown.
- Applies to the *new task* path too: `blankTask` starts both at their unset values, and a brand-new form must read as "no limits" rather than "limit of zero".
- Same treatment is worth considering for the `run_at` picker at line 878, which is the same native control — but that one is genuinely required for a `one_off` task, so it is *not* the same problem and should not be given a clear button. Check, do not assume.

## Acceptance criteria

- An unconfigured task's Limits section reads as "no limits", with no number or date mask that could be mistaken for a configured value.
- A user who sets an end date can clear it back to unset without deleting and recreating the task.
- The wire format is unchanged: no limit is still `stop_after_count: 0` and `stop_after_time: null`.

---
## WHAT
**Summary:** The Tasks *Limits* section renders "no limit" as a `0` and as an empty native date mask, neither of which reads as unset — and an end date, once chosen, cannot be cleared.

**Scope:**
- *Stop after*: give "no limit" a state of its own rather than a magic `0`.
- *Stop at*: an explicit **Not set** representation, and a way to return to it.
- Presentation only — the wire format does not change.

**Affected Components:**
- `src/views/TasksView.tsx` L583-614 — the `Limits` `formsec`: the `stop_after_count` number input (L588-596) and the `stop_after_time` `datetime-local` (L602-612).
- `src/views/TasksView.tsx` L121-137 — `blankTask`, which seeds both at their unset values.
- Possibly `src/styles/` — a per-view stylesheet if the chosen control needs one.

**Acceptance Criteria:**
- [ ] A newly created task's *Limits* section reads as "no limits": no `0` and no date mask that could be mistaken for a configured value.
- [ ] A user who has set an end date can clear it back to unset without deleting and recreating the task.
- [ ] Clearing sends `stop_after_time: null` — never `""`.
- [ ] "No limit" is still `stop_after_count: 0` on the wire; no request or response shape changes.
- [ ] The help text is consistent with whatever is displayed (today's *"0 keeps the task running forever."* becomes wrong the moment `0` stops being shown).
- [ ] The `one_off` `run_at` picker (~L878) is **not** given a clear control — it is required, not optional.

**Out of Scope:**
- Persisting these fields at all — that is #540, and this issue is not testable end to end without it.
- Any backend or wire change.
- The `run_at` picker's own presentation beyond confirming it is left alone.

## WHY
**Business Value:** The two controls that bound an unattended task — how many runs and until when — are the ones a cautious user reaches for first. Both currently display a value when nothing is configured, so the user cannot tell a deliberate limit from an untouched default, and cannot undo an end date at all. That is precisely the state in which someone abandons the safety controls.

**Problem Statement:** Reported verbatim: *"the stop at value in the Limits should be 'not set' or something but currently it's showing me a value ... how can we reset this because user might get confused about it ... if user select anything it shows the value otherwise it should be empty."*

- **Stop at** is a bare `<input type="datetime-local">`. `toLocalStamp` already returns `""` correctly for a null `stop_after_time` (L104-112), so the value is genuinely empty — what the user sees is the *browser's* own mask (`dd/mm/yyyy, --:--`). This is presentation, not state. And there is no control to clear a chosen date.
- **Stop after** is a number spinner whose "no limit" value is `0`, explained only by help text. A task that runs forever shows the same `0` as a task nobody has configured.

**Expected Impact:** Removes a class of inverted-default misreading, and restores an undo that does not exist. Frontend-only, no wire risk.

**Context:**
- This is the same inverted-default problem `CLAUDE.md` already documents for the gateway's retention horizon: *`0` means keep everything, and it is therefore the **longest** horizon rather than the shortest* — every comparison and every label over such a value has to read the right way round. A spinner sitting on `0` reads as "zero runs", the exact inverse of "unlimited".
- **Blocked by: #540** — neither of these fields is persisted today, so a control fixed here would look correct and still forget. #540 must land first, or this cannot be verified end to end.
- Related: #541, #542

## HOW
**Technical Approach:** Two display changes in one `formsec`, both mapping to the existing wire values. Nothing is added to the request or the response: "no limit" stays `stop_after_count: 0` and `stop_after_time: null`, and only what the user sees changes.

**Implementation Details:**

*Stop after* — make "no limit" a state rather than a magic number. Recommended: an *Unlimited / Stop after N runs* pair (radio or toggle) where the number input only renders in the second branch, defaulting to something sensible when switched on. The lighter alternative is an empty-with-placeholder number input mapping `""` → `0` on the wire. Either way the stored value is unchanged.

*Stop at* — render an explicit **Not set** state instead of an empty native picker. Recommended: a *"Set an end date"* affordance that reveals the picker, plus a clear (`×`) control returning it to `null` once set. The clear must call `edit({ stop_after_time: null })` — the existing `onChange` at L605-611 already maps an empty input to `null`, so the failure mode to avoid is sending `""`, which `fromLocalStamp` produces for an invalid date.

*New-task path* — `blankTask` (L121-137) seeds `stop_after_count: 0` and `stop_after_time: null`, so a fresh form must read as "no limits" rather than "a limit of zero". Verify this explicitly; it is the first thing every user sees.

*Help text* — rewrite `"0 keeps the task running forever."` and `"Leave empty for no end date."` to match whatever is displayed. If `0` is no longer visible, the first sentence is actively misleading.

*The `run_at` picker (~L878)* — the same native control, but for a `one_off` task it is **required**, so it is not the same problem. Confirm by reading the code; do not give it a clear button and do not assume symmetry.

*Grammar* — per `CLAUDE.md` → *A form's actions are a fixed grammar*, a clear control is not a submit and must not read as one; and `+` marks a list-level *New X*, so it is not the right glyph for "set an end date" on an existing form either.

**Testing Strategy:** No TypeScript test harness exists in this repo, so:
- Type-level: `npm run build` must pass.
- Manual verification through the `ui-verify` skill against the real Tauri webview: create a task (both rows read "no limit") → set a count and a date → save → reopen (**requires #540**) → clear the date → save → reopen and confirm it is unset.
- Network check: confirm the cleared save sends `stop_after_time: null`, not `""` — via `read_network_requests`, or the app log.

- `docs/user-guide.md` if it describes the Limits controls.
- No new dependencies, env vars or migrations.

**Files to Create / Modify:**
- Modify: `src/views/TasksView.tsx`; a `src/styles/` file only if the chosen control needs one
- Create: none

**Risks:**
- *Shipped before #540* → the controls would look right and still forget; the blocked-by link is the mitigation, and a reviewer should check it is closed first.
- *Clearing sends `""` instead of `null`* → an empty string is not the same as no end date; covered by the network check.
- *The `one_off` `run_at` picker gets the same treatment by symmetry* → explicitly out of scope; it is required.
- *Help text left stale* → its own acceptance criterion.

**Complexity:** Low — one file, well under 100 LOC, no backend and no wire change.

---
**Refined by:** AI Issue Refiner · **Date:** 2026-09-02 · **Status:** Ready

vibexp-agent:
Closes #543

## What

The Tasks *Limits* section rendered both of its "no limit" values as values. `stop_after_count: 0` sat in a number spinner, which reads as *zero runs* — the inverse of unlimited — and `stop_after_time: null` was an empty `datetime-local`, which the browser paints with its own `dd/mm/yyyy, --:--` mask. Neither was distinguishable from a configured value, and an end date once chosen could not be cleared at all.

Each row is now a `Segmented` mode switch over the same stored value — **No limit / Run limit**, **No end date / End date** — with the number input and the picker rendering only in the second branch, so neither unset value is ever on screen. The switch is also the clear control the picker never had.

## Why

Reported from the app: *"the stop at value in the Limits should be 'not set' or something but currently it's showing me a value ... how can we reset this"*. This is the inverted-default class `CLAUDE.md` already documents for the gateway's retention horizon — `0` is the **longest** horizon, not the shortest, so every label over it has to read the right way round.

## How

The mode is **derived from the stored value, never held in state**. `setDraft` has eight call sites in this view (the 10 s re-seed poll, `select`, `startCreate`, `revert`, save-success, `toggleStatus`), and a second copy of the mode would have to resync at every one or silently disagree with the draft. That choice has four consequences, each of which is a guard in the diff:

- **Nothing may write the unset value while the input is mounted.** A `datetime-local` reports `""` for anything that is not a *complete* datetime, so backspacing one segment to retype it would clear the value the render guard reads and unmount the control. An empty value is an edit in progress; the switch beside it is the clear. The count input has the same guard as a floor of 1.
- **`Segmented` fires `onChange` on every click, including the already-selected option** (`ui.tsx`, no short-circuit — eight of its ten consumers are idempotent and never notice). Both handlers early-return when the clicked value equals the derived mode.
- **Seeding "Run limit" is a schedule-stopping act.** `should_auto_pause` is `stop_after_count > 0 && run_count >= stop_after_count`, checked *before* the run and answered by declining it **silently** — no `job_history` row. A flat `10` on a task with 42 runs behind it would stop it for good, so it seeds ten *further* runs.
- **Flicking a switch off stashes what it had**, so flicking it back restores rather than re-seeds. The stash is a memory in a `useRef`, keyed by task id and dropped on Discard — never a source of truth.

## Acceptance criteria

- [x] An unconfigured task's Limits reads as "no limits" — no `0`, no date mask
- [x] A set end date can be cleared without deleting and recreating the task
- [x] Clearing sends `stop_after_time: null`, never `""`
- [x] Wire unchanged — no limit is still `stop_after_count: 0` / `stop_after_time: null`; `blankTask` untouched
- [x] Help text consistent with what is displayed
- [x] The `one_off` `run_at` picker is **not** given a clear control — it is required

## Verification

No TypeScript test harness exists in this repo, so beyond `npm run build` (`tsc --noEmit`) this was verified against the **real Tauri webview** via the repo's `ui-verify` skill, on a dev instance built from this branch. Full results in <https://github.com/shaharia-lab/agento/pull/549#issuecomment-5516016773|the round-1 comment> and the round-2 commit message; the request bodies were read off the app's own `fetch`, showing `"stop_after_time": null` on the clearing save and `"stop_after_count": 0` on the no-limit save. The verification task was deleted afterwards.

## Deviation from the HOW

The HOW recommended a *"Set an end date"* affordance plus a separate `×` clear for **Stop at**, and an *Unlimited / Stop after N runs* pair for **Stop after**. Both rows use the segmented pair instead, so the two limits read as one idea rather than two mechanisms, no new CSS or icon button is needed, and the mode switch doubles as the clear. Every acceptance criterion is met either way.

## Noted, deliberately out of scope

A **past** end date is accepted with no warning and then silently pauses the task on its first fire — `validate_task` never mentions `stop_after_time`. Pre-existing, not worsened here (a newly revealed picker seeds a week into the future), and not #543's scope.

### DIFF
--- a/docs/user-guide.md
+++ b/docs/user-guide.md
@@ -299,8 +299,9 @@ Other options:
 - **Model**: override the agent's model for this task.
 - **Timeout**: runs longer than this are cancelled and recorded as failed.
 - **Save output**: keep the agent's reply in the run history.
-- **Stop after**: end the task after N runs. `0` means never.
-- **Stop at**: an end date.
+- **Stop after**: **No limit**, or a **Run limit** of N runs.
+- **Stop at**: **No end date**, or an **End date**. Switching back to *No end
+  date* clears a date you already chose.
 - **Enabled**: pause without deleting.
 
 The inspector shows the next run, the last run, the total number of runs and the
--- a/src/views/TasksView.tsx
+++ b/src/views/TasksView.tsx
@@ -118,6 +118,26 @@ function fromLocalStamp(v: string): string {
   return isFinite(d.getTime()) ? d.toISOString() : "";
 }
 
+/* --- The two "no limit" defaults ------------------------------------------
+   Both limits are stored as their *unset* value — `stop_after_count: 0` and
+   `stop_after_time: null` — and `0` is the longer horizon, not the shorter
+   one, the same way round as the gateway's `usage_retention_days`. The form
+   therefore never shows either unset value: switching a limit *on* has to
+   seed something a user would plausibly have typed, or the control lands on
+   the number it exists to stop meaning. -------------------------------------- */
+
+/** What "Run limit" starts at. Any value ≥ 1 works; ten is a short leash. */
+const FIRST_STOP_AFTER_COUNT = 10;
+
+/** What "End date" starts at: a week out, truncated to the minute the picker
+ *  shows — so the stored instant and the displayed one agree from the start
+ *  rather than only after the first edit. */
+function firstStopAfterTime(): string {
+  const d = new Date(Date.now() + 7 * 86_400_000);
+  d.setSeconds(0, 0);
+  return d.toISOString();
+}
+
 /* --- New-task template ---------------------------------------------------- */
 
 function blankTask(agentSlug: string): ScheduledTask {
@@ -695,35 +715,105 @@ export function TasksView({
 
                 <div className="formsec">
                   <div className="formsec__title">Limits</div>
-                  <FormRow label="Stop after" help="0 keeps the task running forever.">
+                  {/* Both rows are a mode switch over one stored value, and
+                      the *unset* value is never rendered as a value: a `0` in
+                      a spinner reads as "zero runs" and an empty
+                      `datetime-local` paints the browser's own
+                      `dd/mm/yyyy, --:--` mask, each the inverse of the "no
+                      limit" it means. The segmented pair is also the clear
+                      control the picker never had — selecting "No end date"
+                      writes `null`, so an end date can be undone without
+                      recreating the task. The wire is untouched. */}
+                  <FormRow
+                    label="Stop after"
+                    help={
+                      draft.stop_after_count > 0
+                        ? "The task pauses itself once it has run this many times."
+                        : "The task keeps running until you pause it."
+                    }
+                  >
                     <div className="inline">
-                      <label className="field field--num">
-                        <input
-                          type="number"
-                          min={0}
-                          value={draft.stop_after_count}
-                          onChange={(e) =>
-                            edit({ stop_after_count: Number(e.target.value) || 0 })
-                          }
-                        />
-                      </label>
-                      <span className="inline__label">runs</span>
-                    </div>
-                  </FormRow>
-                  <FormRow label="Stop at" help="Leave empty for no end date.">
-                    <label className="field field--stamp">
-                      <input
-                        type="datetime-local"
-                        value={toLocalStamp(draft.stop_after_time)}
-                        onChange={(e) =>
+                      <Segmented
+                        value={draft.stop_after_count > 0 ? "count" : "none"}
+                        options={[
+                          { value: "none", label: "No limit" },
+                          { value: "count", label: "Run limit" },
+                        ]}
+                        onChange={(v) =>
                           edit({
-                            stop_after_time: e.target.value
-                              ? fromLocalStamp(e.target.value)
-                              : null,
+                            stop_after_count:
+                              v === "count" ? FIRST_STOP_AFTER_COUNT : 0,
                           })
                         }
                       />
-                    </label>
+                      {draft.stop_after_count > 0 && (
+                        <>
+                          <label className="field field--num">
+                            {/* The floor is 1, not 0: `0` is now reached
+                                through "No limit" alone, and letting the
+                                number fall to it would collapse the input the
+                                user is typing into. */}
+                            <input
+                              type="number"
+                              min={1}
+                              value={draft.stop_after_count}
+                              onChange={(e) =>
+                                edit({
+                                  stop_after_count: Math.max(
+                                    1,
+                                    Number(e.target.value) || 1
+                                  ),
+                                })
+                              }
+                            />
+                          </label>
+                          <span className="inline__label">runs</span>
+                        </>
+                      )}
+                    </div>
+                  </FormRow>
+                  <FormRow
+                    label="Stop at"
+                    help={
+                      draft.stop_after_time
+                        ? "Local time; stored as UTC."
+                        : "The task has no end date."
+                    }
+                  >
+                    <div className="inline">
+                      <Segmented
+                        value={draft.stop_after_time ? "date" : "none"}
+                        options={[
+                          { value: "none", label: "No end date" },
+                          { value: "date", label: "End date" },
+                        ]}
+                        onChange={(v) =>
+                          edit({
+                            stop_after_time:
+                              v === "date" ? firstStopAfterTime() : null,
+                          })
+                        }
+                      />
+                      {draft.stop_after_time && (
+                        <label className="field field--stamp">
+                          <input
+                            type="datetime-local"
+                            value={toLocalStamp(draft.stop_after_time)}
+                            onChange={(e) => {
+                              /* Both ways of ending up with nothing collapse
+                                 to `null`: an emptied picker, and a value
+                                 `fromLocalStamp` cannot parse — which answers
+                                 `""`, and `""` is not "no end date" to the
+                                 API. */
+                              const iso = e.target.value
+                                ? fromLocalStamp(e.target.value)
+                                : "";
+                              edit({ stop_after_time: iso || null });
+                            }}
+                          />
+                        </label>
+                      )}
+                    </div>
                   </FormRow>
                 </div>
               </div>
--- a/src/views/TasksView.tsx
+++ b/src/views/TasksView.tsx
@@ -760,7 +760,7 @@ export function TasksView({
                     help={
                       draft.stop_after_count > 0
                         ? "The task pauses itself once it has run this many times."
-                        : "The task keeps running until you pause it."
+                        : "The number of runs is not limited."
                     }
                   >
                     <div className="inline">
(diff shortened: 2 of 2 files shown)


vibexp-agent:
**Round 1 — one finding dismissed, with evidence.**

**#3 ("the number input cannot be cleared to retype")** is not adopted, on two grounds:

1. **It is not a regression.** Before this PR the same input ran `Number(e.target.value) || 0`, so backspacing to empty snapped it to `0` and re-rendered `0`. It snaps to `1` now. The behaviour class is unchanged; only the floor moved, and it moved because `0` is no longer a value this row can show.
2. **The proposed fix is unreachable here.** `value={draft.stop_after_count || ""}` (the `timeout_minutes` pattern at `:697`) can only render `""` when the count is `0` — and the input renders *only* while the count is `> 0`, because the mode is derived from the value. Making it reachable means holding the mode in component state, which is exactly what finding #1 argues against by another route: `setDraft` is called from eight places (the 10 s re-seed poll, `select`, `startCreate`, `revert`, save-success, `toggleStatus`), and a second copy of the mode would have to resync at every one of them or silently disagree with the draft. A derived mode cannot desync. That trade is worth more than saving a select-all.

#1, #2, #4 and #5 are all addressed in `2596835`.

**#5 — the prescribed verification ran**, against the real Tauri webview via the repo's `ui-verify` skill, on a dev instance built from this branch:

| check | result |
|---|---|
| new task's Limits | `No limit` / `No end date` — no `0`, no date mask |
| both branches reveal | number + `runs`, and the picker seeded a week out |
| count `250` → `No limit` → `Run limit` | restores `250`, does not re-seed |
| emptied picker (native setter + `input`) | stays mounted, keeps `2026-09-09T22:31` |
| create → reopen | `250` read back from the server |
| set date → save → reopen | `09/09/2026, 10:32 PM`, `Local time; stored as UTC.` |
| clear → save → reopen | `No end date`, picker gone |
| **the clearing save's body** | `"stop_after_time": null` — JSON `null`, not `""` |
| **the no-limit save's body** | `"stop_after_count": 0` |

Bodies were read off the app's own `fetch`, not inferred. The verification task was deleted afterwards; the dev database is back to zero tasks.

One thing that check surfaced and this PR does **not** change: a task loaded from the server with no end date has `stop_after_time` *absent* from the response, so `JSON.stringify` omits the key on the way back out. That is pre-existing (the field is `string | null | undefined` in `types.ts`) and the backend decodes absent and `null` identically — `stop_after_time_is_none_when_absent_or_null_and_byte_identical_otherwise`. Only an explicit *clear* produces a literal `null`, and it does.

Issue #555, opened 2026-09-07
MCP: every hosted tool is invisible to the model — tools/list lacks the ttlMs/cacheScope the 2026-07-28 revision requires


vibexp-agent:
## What

Every in-process MCP tool Agento hosts — all six integrations (GitHub, Google, Slack, Telegram, Jira, Confluence) **and** the local `current_time` server — is invisible to the model under the current Claude Code CLI (2.1.263). The agent's turn runs, Agento logs `mcp server started … tools=["get_issue", "list_issues"]`, the CLI's `init` event reports the server as `"status":"connected"`, and the model answers *"No GitHub MCP tools are available in this environment"* after `ToolSearch` finds nothing. Nothing in Agento's log, the chat, or the inspector says why.

Reproduced end to end from the UI on 2026-09-07: create an agent (Agents → New Agent, tick `list_issues` + `get_issue` under GitHub), start a chat on it, ask for the five most recent open issues in `shaharia-lab/agento`. Three chats on two agents, same answer every time. The committed probe `cargo test --test mcp_e2e_probe -- --ignored` fails the same way against the local-tools server (`current_time` absent from `init.tools`).

## Why

The CLI negotiates MCP protocol revision **2026-07-28**, and our `tools/list` answer is not a valid result under that revision.

From `claude -p --debug-file …` against the exact `--mcp-config` the runner built (server URL and token read off the live process):

```
MCP server "1763a925-…": Connection established with capabilities: {"hasTools":true,…}
MCP server "1763a925-…": tools/list failed (Invalid result for tools/list: [
  { "expected": "number", "code": "invalid_type", "path": ["ttlMs"], "message": "Invalid input" },
  { "code": "invalid_value", "values": ["public","private"], "path": ["cacheScope"], "message": "Invalid input" }
]
[ERROR] MCP server "1763a925-…" Failed to fetch tools: …
```

The CLI validates the result against the 2026-07-28 schema, the validation fails, and it drops the whole tool list while still reporting the server connected.

The specification is unambiguous. `docs/specification/2026-07-28/changelog.mdx`, minor change 5: *"Require `ttlMs` and `cacheScope` fields on results returned by `tools/list`, `prompts/list`, `resources/list`, `resources/read`, and `resources/templates/list` via a new `CacheableResult` interface"* (SEP-2549). `schema/2026-07-28/schema.json` lists both under `CacheableResult.required`, with `cacheScope ∈ {"public","private"}` and `ttlMs ≥ 0`.

[...]

Issue #501, opened 2026-08-27
MCP: an integration can host zero tools (or every tool) silently, and nothing logs or tests it


vibexp-agent:
## What

An agent's MCP tool set can end up **empty or wider than configured**, and nothing anywhere says so. The turn starts, the server logs `serving on …`, the CLI connects and completes the handshake, `--allowedTools` names the tools — and the model has none of them.

Reported symptom: an agent configured with two Gmail tools from a `google` integration; the model was unaware of the tools. The app log shows both MCP servers serving and the CLI completing `initialize` against each, so nothing in the log distinguishes "2 tools hosted" from "0 tools hosted".

## Why

Three separate things, one failure mode:

**1. `filter_config_tools` drops a service that means "host everything".**
`src-tauri/src/native/integrations/registry.rs::filter_config_tools` intersects the agent's requested tool names with each enabled service's stored `tools` list:

```rust
let kept: Vec<String> = service.tools.iter().flat_map(|l| l.iter())
    .filter(|tool| want.contains(tool.as_str())).cloned().collect();
if kept.is_empty() { continue; }
```

A service stored as `{"enabled": true}` (or `"tools": []`) has no list, so `kept` is empty and the whole service is **dropped**. `build_allowed_set` then sees an empty map, `service_enabled` is false for every group, and the integration hosts **zero tools**. This is reachable through `POST /api/integrations` (the UI always writes an explicit list, so it is API- and legacy-row-shaped, not UI-shaped) and through any row written before the service tool lists existed.

**2. The inverse: an enabled service with `tools: []` exposes every tool.**
`build_allowed_set` treats an empty union as "host everything" (`len(allowed) == 0 || allowed[name]`, ported deliberately). So a user who enables the Gmail service and unchecks *every* tool, paired with an agent that requests no specific tools for that integration, gets **all 8 Google tools hosted** — directly contradicting the UI copy "Only the tools you leave on are exposed to agents". Reachable via the API today; the agent editor drops an empty selection (`setMcpTools`), so the UI half is closed by accident rather than by design.

**3. Neither outcome is observable, and neither is tested.**
- Nothing logs how many tools a started server actually registered. `start_filtered_server` / `start_google_mcp_server` log only "serving on <addr>".
- `--allowedTools` still names tools the server does not host. `start_integration_servers` deliberately does not check requested names against hosted ones (documented for the local server as the kinder failure) — but combined with (1) it means the *entire* tool set can vanish with no signal.
[...]

Issue #519, opened 2026-08-30
UI chrome drift: a second savebar implementation in Agents, and mixed authorise/authorize spelling


vibexp-agent:
## What

Two small drifts found while auditing the Integrations screens; neither is user-reported, both are the kind that get copied into the next view.

**1. Two savebars.** `IntegrationsView`, `TasksView`, `SettingsView` and both gateway views use the shared `.savebar` class with `btn` / `btn--primary`. `AgentsView` has its own `.agents-savebar` with `.agents-savebar__msg`, an icon, and `btn--lg` buttons — a second implementation of the same strip, at a different button size.

**2. Mixed spelling.** The UI copy carries both spellings of the same word: `authorise`/`Authorisation` (5 occurrences) and `authorize`/`Authorization` (3), alongside British `labelled` and `behaviour` elsewhere. Whichever is house style, it should be one.

## Notes for whoever picks this up

- Check whether `agents-savebar` exists for a reason (it carries an inline message the shared one renders differently) before deleting it — the right fix may be to widen `.savebar` to carry a message and drop the duplicate.
- The spelling sweep is UI copy only; identifiers, wire fields and Rust code are out of scope.

---
## WHAT

**Summary:** Collapse the two savebar implementations into one, and settle a single spelling for "authorise/authorize" across the UI copy.

**Scope:**
- `.agents-savebar` versus the shared `.savebar` — one implementation, one button size.
- The `authorise`/`authorize` split in user-visible strings.

**Affected Components:**
- `src/views/AgentsView.tsx` (L690–716) — the `.agents-savebar` markup.
- `src/styles/` — `.agents-savebar*` rules and the shared `.savebar` rules.
- `src/views/IntegrationsView.tsx`, `src/views/TasksView.tsx`, `src/views/SettingsView.tsx`, `src/views/gateway/ProvidersView.tsx`, `src/views/gateway/ModelsView.tsx` — the existing `.savebar` users, which must not shift visually.

**Acceptance Criteria:**
- [ ] One savebar component/class is used by every view; `.agents-savebar` and its modifiers no longer exist in the stylesheets or the JSX.
- [ ] The shared savebar can carry the inline message Agents needs (icon + text, with an error variant) — Agents loses no information in the merge.
- [ ] Button size is consistent: either every savebar uses `btn` or every one uses `btn--lg`, decided once.
[...]

PR #561, opened 2026-09-07
feat: say so when the CLI drops a hosted tool list (#556)

### LINKED ISSUE #556: MCP: say so when the CLI drops a hosted tool list — compare init.tools against what the runner hosted
## What

When the Claude Code CLI discards a hosted MCP server's tool list, Agento reports success everywhere a user or an agent could look: the log says `mcp server started server=… tools=["get_issue", "list_issues"]`, the CLI's `init` event says `"mcp_servers":[{"name":"…","status":"connected"}]`, the chat inspector's Live session shows a tool count, and the model then says the tools do not exist. #555 is the case that exposed it — every integration was dead for at least one CLI release with nothing in the product saying so. The next cause will be different (a schema change, a header the CLI starts requiring, a tool name the CLI filters) and equally silent.

## Why

`report_hosted_tools` (#501) closed the hop before this one: it reads the tool names off the *handle* rather than off the request, so a server hosting nothing is logged as such. The hop it cannot see is the CLI's: whether the tools we hosted were **given to the model**. That fact is on the wire already — the `system`/`init` frame the CLI emits at the start of every turn carries `tools`, the full list the model was given, with MCP tools as `mcp__<server>__<tool>` — and `turn.rs` streams that frame through to the client without reading it.

The two hops disagreeing *is* the defect signal, and it is available for free on the first frame of every turn.

## How

In `chat/turn.rs` (and the executor, which consumes the same event stream): on the `init` system event, compare `tools` against the qualified names of every server `build_options` started (the same `tool_names` `report_hosted_tools` reads). For each hosted server:

- **every hosted tool present** → nothing; today's `info` line already covers it;
- **none present, server reported `connected`** → a `warn` naming the server and the missing tools, and — because a scheduled run has no one reading the log — a line in the chat (a synthetic event the way `user_input_required` is) or the `job_history.error_message` for a run: *"GitHub tools were hosted but the Claude CLI did not offer them to the model; see the app log"*;
- **some present** → `warn` with the difference.

Keep it a warning rather than a refusal: `start_local_tools`' rule is that a missing tool is the kinder failure than a refused run, and this is the same shape.

Two details worth getting right:

- The tools list in `init` reflects `--allowedTools`/`--disallowedTools` filtering too, so compare against what the runner *asked for*, not against every tool the server could host — an allowlist that narrows is not a defect.
- The CLI's per-run debug log (`--debug-file`) carries the reason (`tools/list failed (Invalid result …)`), and Agento never sees it. Consider passing `--debug-file <data dir>/claude-debug/<turn>.log` on a turn that hits the mismatch path, or documenting the hand-run in `docs/troubleshooting.md` so the user can get the reason in one command (the recipe is in `.claude/skills/local-verify`).

A test: `tests/chat_turn.rs`'s fake CLI emits an `init` whose `tools` omits a hosted server's names and the test asserts the warning and the synthetic frame; and one where they are present and nothing is emitted.

---
## WHAT

**Summary:** Compare the CLI's `init` frame — the list of tools the model was actually given — against the tools Agento hosted for the turn, and say so, in the log and in the run, when they disagree.

**Scope:**
- On the `system`/`init` event of every turn, diff `SystemMessage::tools` against the qualified names of each MCP server the turn started.
- Disagreement produces a `warn` naming the server and the missing tools.
- A total miss on a server the CLI reported `"connected"` also produces a user-visible signal: a synthetic SSE frame on the chat path, and text on the `job_history` row for a scheduled or manual run.
- Fake-CLI tests covering the mismatch and the agreement cases.

**Affected Components:**
- `src-tauri/src/native/chat/turn.rs:547` — `handle_event` has no `"system"` arm today; add one. The data is already decoded: `Event::system` is a `SystemMessage` carrying `tools: Vec<String>` and `mcp_servers` with each server's `status` (`claude/messages.rs:896-947`). No new parsing.
- `src-tauri/src/native/chat/runner.rs:699` — `report_hosted_tools` and its two call sites (`:610`, `:664`) already hold the ground truth. `build_options` must hand back what it hosted per server, so the turn can diff against it; today the names are logged and dropped.
- `src-tauri/src/native/agent_run.rs:46` (`run_headless`) — the headless event loop the scheduler consumes via `executor.rs:673 run_agent`; the same check belongs here, ending in the `job_history` row rather than an SSE frame.
- `src-tauri/src/claude/options.rs:275-284` — `allowed_tools` / `disallowed_tools`, the filter that must be applied before the diff is meaningful.
- `src-tauri/tests/chat_turn.rs` — the scripted fake CLI.

**Acceptance Criteria:**
- [ ] A turn whose `init.tools` contains every qualified name of a hosted server emits **no new log line and no new frame** — today's `info` from `report_hosted_tools` stays the whole record.
- [ ] A turn whose `init.tools` contains none of a hosted server's names, with that server `"connected"` in `init.mcp_servers`, emits a `warn` naming the server and the missing tools, **and** one user-visible signal (chat frame / `job_history.error_message`).
- [ ] A partial miss emits a `warn` listing exactly the difference.
- [ ] The diff is taken against what the runner **asked for** after `--allowedTools`/`--disallowedTools` narrowing — an allowlist that legitimately narrows the set never warns.
- [ ] A mismatch never fails or refuses the turn; the run proceeds and produces its normal result.
- [ ] `tests/chat_turn.rs` covers all three cases (all present / none present / some present) against the scripted fake CLI.

**Out of Scope:**
- The `ttlMs`/`cacheScope` fix itself — that is #555. This issue is the signal, not the cause, and must be verifiable with #555 unfixed (that is in fact the easiest way to test it by hand).
- Any change to what is hosted, or to `--allowedTools` assembly.
- Retrying, reconnecting, or refusing a turn on mismatch.
- A UI surface beyond the existing chat event stream and the job-history text; no new view, no new inspector panel.

## WHY

**Business Value:** #555 killed every integration in the product for at least one CLI release, and the product's own instruments all reported success: the log said `mcp server started server=… tools=["get_issue", "list_issues"]`, the CLI's `init` said `"status":"connected"`, the chat inspector showed a tool count, and only the model's prose said otherwise. The next cause will be different — a schema change, a header the CLI starts requiring, a tool name it filters — and equally silent unless this lands.

**Problem Statement:** `report_hosted_tools` (#501) closed the hop before this one: it reads names off the **handle** rather than off the request, so a server hosting nothing is logged as hosting nothing. The hop it cannot see is the CLI's — whether the tools we hosted were *given to the model*. That fact is already on the wire and already decoded: `turn.rs` receives the `system`/`init` event with `tools` populated, forwards it to the client, and never reads it. Two hops disagreeing **is** the defect signal, available for free on the first frame of every turn.

**Expected Impact:** A class of failure that currently takes a hand-run of `claude -p --debug-file` to even name becomes a line in the app log, and a line the user sees. Diagnosis time for the next occurrence drops from a triage session to reading the log. Cost is one set-difference over two small string lists, once per turn.

**Context:**
- Related: #555 — the instance that exposed this. **Independent work; the two can be implemented in parallel**, and #556 is best tested with #555's patch reverted or against a stubbed `init`.
- Related: #501 — `report_hosted_tools`, whose doc comment already states that "the two disagreeing is exactly the defect this reports"; this issue extends that from the handle-vs-request pair to the handle-vs-model pair.
- Related: #282 — the one way to host a tool, which is why every integration and the local server share a single code path and therefore a single check.
- Verified against `main` @ `b8ebae2` on 2026-09-07: `turn.rs::handle_event` has arms for `"assistant"` and `"result"` only, `Event::system` and `SystemMessage::tools` exist and are tested (`messages.rs::init_carries_the_capabilities_the_handshake_does_not`), and `report_hosted_tools` is at `runner.rs:699` with the two call sites the issue names. Nothing in the report was stale.

## HOW

**Technical Approach:** Have `build_options` return, alongside the server handles it already returns, the qualified tool names it registered per server — it computes them already for `allowed`, via `registry::allowed_tool_names` and `tools::allowed_tool_names`. Carry that map into both event loops. On the `system`/`init` event, subtract `init.tools` from it and report the remainder. Keep it a **warning, never a refusal**: `start_local_tools`' own header records why a missing tool is the kinder failure than a refused run, and this is the same shape.

**Implementation Details:**
- `runner.rs` — extend the `build_options` return (or the `tool_servers` value it already hands back) with `Vec<(server_key, Vec<qualified_name>)>`. Prefer widening the existing handle type over a second parallel return, so the two cannot drift. The qualified form is `mcp__<server>__<tool>`, which `allowed_tool_names` already produces — reuse it rather than formatting the string again.
- New shared helper, next to `report_hosted_tools` in `runner.rs` so the rule lives beside its sibling: `fn report_tools_offered(hosted: &[(String, Vec<String>)], init: &SystemMessage) -> Vec<Mismatch>` — logs the `warn` lines and returns what each caller needs for its own user-visible surface. One function, two callers, no duplicated policy.
- `turn.rs::handle_event` — a `"system"` arm; on `subtype == messages::subtype::INIT` (the constant at `messages.rs:63`) with `event.system` present, call the helper. On a total miss for a `"connected"` server, emit a synthetic SSE frame the way `user_input_required` and `permission_request` are emitted (`sse::json_frame`) — a new event name, with the server and the missing names in the payload. Return `Flow::Continue` unconditionally.
- `agent_run.rs::run_headless` — the same call in the headless loop; the result travels out in `RunResult` so `executor.rs::finish` can put it on the `job_history` row. Wording along the lines of *"GitHub tools were hosted but the Claude CLI did not offer them to the model; see the app log"*. A scheduled run has nobody reading the log, and `schedule/CLAUDE.md`'s rule that every run path ends in a `job_history` row is why the signal has to reach that row.
- **The filtering detail that decides correctness:** `init.tools` reflects `--allowedTools` / `--disallowedTools` (`options.rs:780-785`), so diff against what the runner asked for, not against every tool a server could host. Getting this backwards makes a narrowing allowlist warn on every turn, which is worse than silence.
- Any new synthetic frame follows the wire rules — `gojson` encoding, HTML escaping, key order — as `turn.rs`'s own frame tests pin (`a_synthetic_frames_input_is_compacted_and_escaped_the_way_go_marshals_one`).

**Testing Strategy:**
- `src-tauri/tests/chat_turn.rs` — script the fake CLI to emit an `init` whose `tools` omits a hosted server's names; assert the `warn` and the synthetic frame. A second script where they are all present; assert neither. A third with a partial overlap; assert the difference is what is named. (Unix-only, like its siblings — `PermissionsExt` on the fake CLI.)
- `runner.rs` unit tests — the helper's set arithmetic directly, including the allowlist-narrowing case that must **not** warn.
- `src-tauri/tests/scheduled_run.rs` — a run whose `init` drops the tools writes the message onto its `job_history` row and still completes.

- `src-tauri/src/native/chat/CLAUDE.md` — the new synthetic event alongside `user_input_required` and `permission_request`, which that file enumerates as a contract.
- `docs/troubleshooting.md` — what the warning means and the one-command hand-run that gets the underlying reason: the CLI's per-run `--debug-file` log carries it (`tools/list failed (Invalid result …)`) and Agento never sees it. **Document the hand-run rather than wiring `--debug-file` into the mismatch path** — writing a per-turn debug log into the data directory is a new file lifecycle, a new disk-growth question and a new place for tokens to land, and it is not needed for the diagnosis. If a later issue wants it, that is its own decision.
- No new dependencies or env vars.

**Files to Create / Modify:**
- Modify: `src-tauri/src/native/chat/runner.rs`, `src-tauri/src/native/chat/turn.rs`, `src-tauri/src/native/agent_run.rs`, `src-tauri/src/native/schedule/executor.rs`, `src-tauri/tests/chat_turn.rs`, `src-tauri/tests/scheduled_run.rs`, `src-tauri/src/native/chat/CLAUDE.md`, `docs/troubleshooting.md`.
- Create: none.

**Risks:**
- *False positives from legitimate allowlist narrowing* → diff against the requested set after filtering; a dedicated unit test pins it.
- *A noisy `warn` on every turn devalues the signal* → the all-present case must emit nothing at all; that is an acceptance criterion, not a nicety.
- *Widening `build_options`' return touches the chat turn, the headless run and the tests at once* → carry the names on the existing server handle instead of adding a parallel value, so there is one source of truth.
- *A new SSE event name is a wire contract the UI must tolerate* → the client ignores unknown event names today; keep the payload flat and encode it through `gojson` like the other two synthetic frames.

**Complexity:** Medium (four source files plus tests, well under 500 LOC, no new architecture — the data is already decoded and the policy already has a home)

---
**Refined by:** AI Issue Refiner · **Date:** 2026-09-07 · **Status:** Ready

vibexp-agent:
Closes #556

## What

On the `system`/`init` event of every turn, diff `SystemMessage::tools` — the
list of tools the Claude Code CLI actually gave the model — against the
qualified names each MCP server the turn hosted put in front of it, and say so
when they disagree.

## Why

`report_hosted_tools` (#501) closed the hop before this one: it reads names off
the *handle* rather than off the request, so a server hosting nothing is logged
as hosting nothing. The hop it cannot see is the CLI's — whether what we hosted
reached the model. #555 is what that costs: every integration in the product was
dead for at least one CLI release while the log said
`mcp server started server=… tools=[…]`, `init` said `"status":"connected"`, and
the inspector showed a tool count. Only the model's prose disagreed.

## How

- **`runner.rs`** — `build_options` returns a third value, `Vec<HostedTools>`:
  per server, the **overlap** of what it registered and what the agent asked
  for, qualified. It is built at the two `report_hosted_tools` call sites (that
  function now returns the entry it just reported, so the reported and diffed
  sets are one computation), then narrowed by `narrow_to_command_line` against
  the assembled `--allowedTools` / `--disallowedTools`.

  It is deliberately **not** read back off `InProcessMcpServer`, which the
  issue suggested: the handle does not know the name the CLI prefixes tools
  with. An integration's server is *built* as `github-<id>` and *registered* as
  `<id>`, and only the second is the tool prefix — `registry::allowed_tool_names`
  says so at length. The doc comment on `build_options` records this.

- **`report_tools_offered`** is the shared policy, beside its #501 sibling: it
  emits the `warn` lines and hands each caller what it needs for its own
  user-visible surface. One function, two callers, no duplicated policy.

- **`turn.rs`** — a `"system"` arm on `handle_event`, and a third synthetic
  frame, `tools_not_offered` (flat, `gojson`-encoded, no embedded raw value).
  Returns `Flow::Continue` unconditionally.

- **`agent_run.rs` / `executor.rs`** — the same check in the headless loop; the
  finding travels out on `RunResult::tools_not_offered` and lands in
  `job_history.error_message`, with the status left at `success`. A scheduled
  run has nobody reading the log, and that row is the only record it leaves.

Three rules, each of which makes the signal worthless if broken, and each an
acceptance criterion:

- the diff is against what the runner **asked for after filtering** — a
  narrowing allowlist is not a defect;
- the **agreeing case emits nothing at all** — no line, no frame;
- a mismatch is a **warning, never a refusal** (`start_local_tools`' rule), and
  only a *whole* server's tools vanishing from a server the CLI called
  `connected` reaches the user. A partial miss stays in the log, because a
  filter this side does not model would otherwise put a scary sentence in front
  of a user on every turn.

## Tests

- `runner.rs` — eight unit tests over the set arithmetic: the overlap, the
  all-present silence, #555's whole-server shape, a partial miss naming exactly
  the difference, a server that never connected, **allowlist and denylist
  narrowing**, and an empty allowlist narrowing nothing. Each uses a `556-`
  prefixed server key, because the `testlog` buffer is process-wide and
  `cargo test` runs a binary's tests in parallel.
- `tests/chat_turn.rs` — two scripted-CLI turns against the real hosted
  `local-tools` server: an `init` listing its tool adds no frame; an `init`
  listing none of it emits the `warn`, the frame (asserted at the bytes) and
  still produces the turn's normal result.
- `tests/scheduled_run.rs` — a run whose `init` drops the tools completes with
  status `success` and the sentence on its `job_history` row.

**All three fail on the revert**, verified by hand: removing the frame emission
fails the chat test, passing `""` instead of the notice fails the scheduled
test, and making `narrow_to_command_line` a no-op fails both narrowing tests.

## Deviation from the HOW, one

The spec asks `tests/chat_turn.rs` to cover **three** cases (all present / none
present / some present). The partial case is not constructible there: the only
server an integration test can host is `local-tools`, which registers exactly
one tool (`current_time`), so "some but not all missing" has no shape — and an
integration server cannot be hosted from `tests/`, because `mcp_plan` resolves
names against the process-wide `paths::database_path()` rather than the test's
temp file (the same constraint `an_agent_whose_tools_this_build_cannot_host_is_a_recorded_failure`
already documents). The partial case is covered as a unit test in `runner.rs`,
where any hosted/requested/offered combination can be constructed; the two
reachable cases are covered end to end in `chat_turn.rs`.

## Out of scope, as the issue says

`--debug-file` is **not** wired into the mismatch path — `docs/troubleshooting.md`
documents the hand-run instead. No change to what is hosted or to
`--allowedTools` assembly. No retry, reconnect or refusal. No new UI surface.

## Docs

`src-tauri/src/native/chat/CLAUDE.md` (the synthetic-event contract goes from
two frames to three, plus the three rules above) and `docs/troubleshooting.md`
(what the message means, and the one-command hand-run that gets the CLI's own
reason).

### DIFF
--- a/docs/troubleshooting.md
+++ b/docs/troubleshooting.md
@@ -288,6 +288,32 @@ redirect to a local address, allow it and retry.
 Three things have to line up: the integration is connected, the service is
 enabled inside it, and the tool is ticked on the agent.
 
+### "…tools were hosted but the Claude CLI did not offer them to the model"
+
+Agento started the server and registered its tools, and the Claude Code CLI then
+gave the model none of them. The chat shows this as a notice on the turn; a
+scheduled task records it on the run's row in **Tasks → History**, which still
+reads `success` because the run itself completed. Everything else in the app
+reports the server as connected, which is why this message exists at all.
+
+The tools are genuinely unavailable to the model for that turn, so it will say
+the tool does not exist. Nothing on the Agento side can be reconfigured to fix
+it — it is the CLI that dropped the list — but two things narrow it down:
+
+- **The app log names the server and the missing tools**, on a line beginning
+  `mcp tools not offered to the model`. See [Reading the logs](#reading-the-logs).
+- **The reason is in the CLI's own debug log, which Agento never sees.** Run one
+  turn by hand with `--debug-file` and grep it:
+
+  ```bash
+  claude -p --debug-file /tmp/cli.debug "hello"
+  grep -i mcp /tmp/cli.debug
+  ```
+
+  A line like `tools/list failed (Invalid result for tools/list: …)` names the
+  field the CLI rejected. Upgrading the Claude Code CLI, or updating Agento, is
+  usually the fix — this is what a version skew between the two looks like.
+
 ### WhatsApp is listed but unusable
 
 Agento does not support WhatsApp. An integration created by an older version is
--- a/docs/troubleshooting.md
+++ b/docs/troubleshooting.md
@@ -292,7 +292,7 @@ enabled inside it, and the tool is ticked on the agent.
 
 Agento started the server and registered its tools, and the Claude Code CLI then
 gave the model none of them. The chat shows this as a notice on the turn; a
-scheduled task records it on the run's row in **Tasks → History**, which still
+scheduled task records it on the run's row in **Job History**, which still
 reads `success` because the run itself completed. Everything else in the app
 reports the server as connected, which is why this message exists at all.
 
--- a/src-tauri/src/native/agent_run.rs
+++ b/src-tauri/src/native/agent_run.rs
@@ -164,9 +164,7 @@ async fn collect_run_result(
                     crate::native::chat::runner::report_tools_offered(hosted_tools, system)
                 {
                     if dropped.whole_server {
-                        tools_not_offered.push(crate::native::chat::runner::tools_dropped_message(
-                            &dropped.server,
-                        ));
+                        tools_not_offered.push(dropped.message());
                     }
                 }
             }
--- a/src-tauri/src/native/agent_run.rs
+++ b/src-tauri/src/native/agent_run.rs
@@ -31,6 +31,12 @@ pub struct RunResult {
     pub output_tokens: i64,
     pub cache_creation_tokens: i64,
     pub cache_read_tokens: i64,
+    /// One sentence per MCP server whose hosted tools the CLI never handed the
+    /// model (#556). A scheduled run has nobody reading the app log, so this
+    /// travels out to `job_history` — see
+    /// [`crate::native::schedule::executor`]. Empty is the normal case, and a
+    /// non-empty one never fails the run.
+    pub tools_not_offered: Vec<String>,
 }
 
 /// `agent.RunAgent`: build the options, spawn the CLI, drain it, answer the
@@ -67,7 +73,7 @@ pub async fn run_headless(
     // The refusal this port has that Go does not — an agent whose tools cannot
     // be hosted here. The caller decides what to do with it; both callers
     // record it rather than dropping it.
-    let (options, tool_servers) =
+    let (options, tool_servers, hosted_tools) =
         tokio::time::timeout_at(deadline, runner::build_options(spec, None))
             .await
             .map_err(|_| DEADLINE_EXCEEDED.to_string())?
@@ -90,9 +96,10 @@ pub async fn run_headless(
         }
     };
 
-    let collected = tokio::time::timeout_at(deadline, collect_run_result(&mut stream))
-        .await
-        .unwrap_or_else(|_| Err(DEADLINE_EXCEEDED.to_string()));
+    let collected =
+        tokio::time::timeout_at(deadline, collect_run_result(&mut stream, &hosted_tools))
+            .await
+            .unwrap_or_else(|_| Err(DEADLINE_EXCEEDED.to_string()));
 
     stream.close();
     // The in-process tool listeners outlive the subprocess and stop when
@@ -140,11 +147,30 @@ pub const DEADLINE_EXCEEDED: &str = "context deadline exceeded";
 /// and returning early would race the scanner against a half-written file.
 async fn collect_run_result(
     stream: &mut crate::claude::client::Stream,
+    hosted_tools: &[crate::native::chat::runner::HostedTools],
 ) -> Result<RunResult, String> {
     let mut result: Option<RunResult> = None;
     let mut result_err: Option<String> = None;
+    let mut tools_not_offered: Vec<String> = Vec::new();
 
     while let Some(event) = stream.next_event().await {
+        // The `init` frame says what the model was actually given, which is the
+        // only place a dropped tool list is visible (#556). A scheduled run has
+        // no stream to put a synthetic frame on, so the finding travels out on
+        // the result and reaches the `job_history` row instead.
+        if let Some(system) = event.system.as_ref() {
+            if system.subtype == crate::claude::messages::system_subtype::INIT {
+                for dropped in
+                    crate::native::chat::runner::report_tools_offered(hosted_tools, system)
+                {
+                    if dropped.whole_server {
+                        tools_not_offered.push(crate::native::chat::runner::tools_dropped_message(
+                            &dropped.server,
+                        ));
+                    }
+                }
+            }
+        }
         let Some(r) = event.result.as_ref() else {
             continue;
         };
@@ -158,6 +184,7 @@ async fn collect_run_result(
                 output_tokens: r.usage.output_tokens,
                 cache_creation_tokens: r.usage.cache_creation_input_tokens,
                 cache_read_tokens: r.usage.cache_read_input_tokens,
+                tools_not_offered: Vec::new(),
             });
         }
     }
@@ -165,7 +192,12 @@ async fn collect_run_result(
     if let Some(err) = result_err {
         return Err(err);
     }
-    result.ok_or_else(|| "agent finished without returning a result".to_string())
+    // Attached after the loop, so it is carried by whichever `result` frame
+    // turned out to be the last one — `init` always precedes them all.
+    let mut result =
+        result.ok_or_else(|| "agent finished without returning a result".to_string())?;
+    result.tools_not_offered = tools_not_offered;
+    Ok(result)
 }
 
 /// `buildResultError`, message for message.
--- a/src-tauri/src/native/chat/CLAUDE.md
+++ b/src-tauri/src/native/chat/CLAUDE.md
@@ -107,9 +107,13 @@ every turn is worse than silence. **The agreeing case emits nothing at all** —
 no line, no frame. And a mismatch is a **warning, never a refusal**: the turn
 runs and answers, for `start_local_tools`' own reason. Only a *whole* server's
 tools vanishing from a server the CLI called `connected` reaches the user, as a
-`tools_not_offered` frame here and as text on the `job_history` row for a
-headless run (`agent_run::collect_run_result` → `executor::finish`), because a
-scheduled run has nobody reading the log. The reason behind a miss is in the
+`tools_not_offered` frame here — rendered by `useChatStream`'s own case as the
+amber `banner`, never the red `banner--error`, because the turn answered — and
+as text on the `job_history` row for a headless run
+(`agent_run::collect_run_result` → `executor::finish`), because a scheduled run
+has nobody reading the log. **That row is the first `success` row that can
+carry `error_message`**, so `JobsView` titles the group off the status rather
+than off the column being non-empty. The reason behind a miss is in the
 CLI's own `--debug-file` log, which Agento never sees; `docs/troubleshooting.md`
 carries the hand-run that gets it.
 
--- a/src-tauri/src/native/chat/CLAUDE.md
+++ b/src-tauri/src/native/chat/CLAUDE.md
@@ -113,7 +113,11 @@ as text on the `job_history` row for a headless run
 (`agent_run::collect_run_result` → `executor::finish`), because a scheduled run
 has nobody reading the log. **That row is the first `success` row that can
 carry `error_message`**, so `JobsView` titles the group off the status rather
-than off the column being non-empty. The reason behind a miss is in the
+than off the column being non-empty. **The sentence names the server's *label*,
+not its `mcpServers` key** — for an integration that key is a v4 UUID the user
+has never seen — which is why it is `ToolsDropped::message()` rather than a
+function taking a name: with two callers and a free function, one of them passed
+the wrong field. The reason behind a miss is in the
 CLI's own `--debug-file` log, which Agento never sees; `docs/troubleshooting.md`
 carries the hand-run that gets it.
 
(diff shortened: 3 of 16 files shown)


vibexp-agent:
## Not auto-merged — the 3-round automated review budget is spent

This PR was driven by the autonomous issue→PR job with `auto-merge` on. It is **ready for review and not merged**, because the pipeline's hard gate is *CI green **and** an automated APPROVE within 3 rounds*, and round 3 returned REQUEST CHANGES. Its findings are fixed in `94754ac`, but those fixes have not themselves been reviewed, so the merge is a human's call.

**CI is green on `94754ac`** — all 9 checks, including `Windows unit tests` and every CodeQL job.

### What the three rounds found, and what changed

| Round | Finding | Fixed in |
|---|---|---|
| 1 | The `tools_not_offered` frame reached the wire and was dropped by `useChatStream`'s `default:` arm — the chat path showed nothing, which is the acceptance criterion the issue exists for | `e097f30` |
| 1 | A `success` job row rendered a red **Error** block, contradicting its own status | `e097f30` |
| 1 | A third copy of the `mcp__<server>__<tool>` prefix rule | `e097f30` |
| 2 | The chat notice replaced instead of accumulating, so with two servers dropped only the last was named — and #555's shape was *every* integration going at once | `6669610` |
| 2 | `NoticeBar` was a verbatim copy of `ErrorBar`; collapsed to one `Bar` with an opt-in tone, per `SaveBar`'s own lesson (#519) | `6669610` |
| 2 | `docs/troubleshooting.md` sent the reader to "Tasks → History"; the sidebar calls it **Job History** | `6669610` |
| 2 | `let registered` shadowed a binding eight lines below it | `6669610` |
| 3 | **The user-facing sentence named the `mcpServers` key, which for an integration is a v4 UUID** — so for exactly the servers #555 killed it read `a3f2c9e1-… tools were hosted`. A `label` now travels beside the key from `mcp_plan` (`can_host` → `hostable_type`, the same single read); the log line and the frame's `server` field keep the key | `94754ac` |
| 3 | An `init` frame listing *no* tools is a frame we could not read, not a CLI that offered nothing — `SystemMessage` is `serde(default)` with `lenient` fields, so a renamed key decodes as an empty `Vec` with no `decode_err`, and every server would read as a whole-server miss on every turn | `94754ac` |
| 3 | `JobsView`'s "No output was saved for this run." fallback still derived meaning from `error_message` being non-empty | `94754ac` |

Every new rule has a test that was **verified to fail on the revert**, by hand: the frame emission, the `job_history` notice, both narrowing cases, the label plumbing (driven through `mcp_plan` + `start_integration_servers`, not a literal), and the unreadable-`init` guard.

### Deliberately not done

- **`tests/chat_turn.rs` covers two of the issue's three cases, not three.** A partial miss is not constructible there: `local-tools` hosts exactly one tool, and an integration server cannot be hosted from `tests/` because `mcp_plan` resolves against the process-wide `paths::database_path()`. The partial case is a `runner.rs` unit test.
- **The trigger (Telegram) run path drops `tools_not_offered`.** It calls the same `run_headless` but writes no `job_history` row by design, so it is the one remaining unattended path with no signal. Outside #556's stated scope — worth its own issue, which I have deliberately not opened.
- `stream.notice` is rendered without the `chatId === selected` guard, exactly as the pre-existing `ErrorBar` two lines below it is. Guarding one and not the other would introduce the inconsistency it was meant to remove.

Issue #569, opened 2026-09-07
Integrations UI: Inbound section on Slack, execution settings on the rule form, connection badge


vibexp-agent:
Part of epic #562. Blocked by: #566.

## What

The Integrations UI for the feature, on the Slack provider only:

- An **Inbound** section (`INBOUND_TITLE` beside the other section constants in `IntegrationsView.tsx:65-76`) rendered in the trigger-gated slot after the services and auth sections, with `supportsTriggers: true` on the Slack provider in `catalog.ts` so the trigger rules list appears too.
- An **App token** field in the connection section: `••• stored` + *Replace* when `has_app_token`, the `bot_token` treatment; help copy naming Socket Mode and linking the docs manifest.
- A **Socket Mode** switch that calls `PUT /api/integrations/{id}/inbound`; disabled (with the reason) while no app token is stored; beside it a status badge from `inbound_status` — `badge--green` Connected, amber Reconnecting, `badge--red` Error with `inbound_error` as the title — the `WebhookPanel` three-way shape.
- The trigger **rule form** gains the execution settings: agent (existing), model (the `NewChatBar` `MODELS` dropdown, locked with "from agent" when the agent names one), working directory (`DirField`), settings profile (the `NewChatBar` profile dropdown, hidden with no profiles), permission mode (`PERMISSION_MODES`), timeout minutes. Chip strip on the row shows the non-default ones. Channel ids field help says "Slack channel ids, comma-separated; empty = every channel the app is in".
- Copy: switch off is *Turn off*, not `Remove`/`Delete` (nothing is destroyed); deleting a rule stays `Delete <name>?`.

## Why

Decision 10 of the epic and the `src/CLAUDE.md` form grammar. The rule form is the one place a user sets what a channel runs with; the switch and badge are the only place they learn whether the socket is alive.

## How

- Keep `TriggerRules` self-contained (own `useResource`, own `run()` helper) as it is; extend `RuleDraft`/`BLANK_RULE` with the five fields as strings.
- `withCurrent()` from `NewChatBar` keeps a stored model/profile the picker does not enumerate.
- Poll the integration read while the section is visible so the badge follows reconnects (the sessions status poll cadence).

## Acceptance

- [ ] `ui-verify` screenshots: the section renders on Slack and on no other provider; the switch is disabled with no token and enabled after a token is stored.
- [ ] Saving a rule from the form round-trips all five settings (`curl` the rule back); the model picker is locked when the chosen agent has a model.
[...]

Issue #567, opened 2026-09-07
Slack Socket Mode worker: connect, ack, dedup, reconnect with backoff, stored status


vibexp-agent:
Part of epic #562. Blocked by: #566.

## What

The Socket Mode worker, `native/integrations/slack/socket.rs`: for every Slack integration with `inbound_enabled` and an app token, a tokio task that calls `apps.connections.open` with the `xapp-` token, opens the returned `wss://` URL, acknowledges every envelope by `envelope_id` before doing anything else, de-duplicates by `event_id`, hands `app_mention` events to a handler (#568 supplies it; this issue lands with a no-op that logs at `debug`), honours `disconnect` envelopes by reconnecting, reconnects after any drop with capped exponential backoff, and keeps `inbound_status` / `inbound_error` current.

## Why

Decisions 11 and 12 of the epic. There is no long-lived outbound connection anywhere in the tree today and no websocket crate in the lockfile, so the lifecycle, the dependency and the failure reporting are all first-of-their-kind here.

## How

- **Dependency**: `tokio-tungstenite` with `rustls-tls-native-roots`, argued in `Cargo.toml` the way the `rmcp` paragraph is — count the packages it adds, and say why the platform trust store rather than bundled roots (the corporate-proxy reason already recorded for `reqwest`).
- **Lifecycle**: owned by `integrations/registry.rs`, a handle whose drop cancels the task; `stop`/`reload` use the existing generation counter so a stop landing mid-connect drops the handle rather than leaving a socket holding a credential. Started beside `start_all` at boot, **after** `keys::install` and `tokens::load_revoked` — extend `a_gateway_start_requires_an_installed_keypair`'s source assertion to cover it.
- **Ack first**: Slack expects the ack within seconds; the handler runs after, on `tokio::spawn`, under the trigger dispatcher's semaphore.
- **Dedup**: `INSERT OR IGNORE` into `slack_processed_events` inside an immediate transaction plus a 48-hour sweep — `claim_update`'s shape.
- **Backoff**: 1 s doubling to a 60 s cap with jitter, **re-anchored on the wall clock** the way the scheduler's `advance_past_now` is, because `tokio::time::sleep` does not advance while the machine sleeps; a fresh `apps.connections.open` on every attempt (the URL is single-use).
[...]

Issue #562, opened 2026-09-07
Epic: Slack Socket Mode inbound — @agento mention starts a chat, thread replies resume it


vibexp-agent:
### Problem

Agento can only be driven from its own window or by a schedule. The one inbound path that exists, the Telegram webhook (`native/trigger/`), starts a fresh chat per message, stores nothing about where the message came from, cannot continue a conversation, and gives a trigger rule no execution settings (`dispatcher.rs:255-256` says so at the call site). A team working in Slack cannot ask an agent a question where they already are, and a desktop app has no public URL for Slack's Events API anyway.

### Solution

A Slack integration can opt into **Socket Mode** with an app-level token. Agento holds one websocket per enabled Slack integration, receives `app_mention` events, and treats a top-level mention as *start a chat with this channel's settings* and a mention inside a thread Agento already answers in as *continue that chat*. Each turn runs headless through the existing runner (`native/agent_run.rs`) and posts the final answer back into the thread as Slack mrkdwn. A mapping table ties `(integration, channel, thread)` to a chat so the thread and the Agento chat are one conversation.

### Design decisions

1. **Entry points.** Only `app_mention` events. A mention in a top-level message starts a chat and a thread under it. A mention inside a mapped thread resumes that chat. A mention inside a thread Agento did not start, a message without a mention, and anything from a bot or the app itself are ignored. Since both start and resume require a mention, `app_mention` is the only event subscription needed — no `message.channels` history scope.
2. **Who.** Anyone in a channel the app is a member of. Channel membership is the boundary.
3. **Configuration lives in `trigger_rules`**, which gains `model`, `working_directory`, `settings_profile_id`, `permission_mode` and `timeout_minutes` (empty = agent default). A Slack rule names channel ids in `filter_chat_ids`; a rule with none is the workspace-wide default. Telegram rules inherit the settings for free.
4. **Selection is most-specific first, and a disabled specific rule means silence.** A channel-specific rule beats the default; if that specific rule is disabled the channel is ignored rather than falling through — the per-channel off switch a workspace default otherwise removes. Telegram's first-enabled-by-creation selection is untouched.
5. **Headless like scheduled runs.** No permission handler; the rule's `permission_mode`, then the agent's, then bypass (`runner.rs:316-333`). No `AskUserQuestion`.
[...]

Issue #515, opened 2026-08-30
Integrations: the credential field moves to a bottom "Credentials" section on edit, because a save wipes stored secrets unless they are re-typed


vibexp-agent:
Reported from the app with a screenshot: connect a GitHub integration with a personal access token, then reopen it to edit. The token field is no longer where it was.

## What

On the **connect** screen (`ConnectForm`) the credential fields sit under *Connection*, directly beneath **Name** — one coherent "here is the connection" block.

On the **edit** screen (`IntegrationDetail`) the same fields are moved to a separate *Credentials* section at the very bottom of the page, below *Services and tools*, behind a warning banner:

> Agento never returns stored secrets, and saving replaces them. Re-enter the credentials below to save any change on this page — otherwise the stored ones would be cleared.

So renaming an integration, or unticking one tool, forces the user to re-type a token they cannot read back from anywhere. This applies to every token-based provider (GitHub, Slack bot token, Telegram, Jira, Confluence).

## Why

The UI is not being fussy — it is defending against a real backend defect. `PUT /api/integrations/{id}` replaces the `credentials` column wholesale while `GET` scrubs it, so a read-then-write round trip destroys the stored secret. `CLAUDE.md` records this under *Known bugs* as "data loss and the highest-value fix on this list", and `IntegrationsView.tsx` comments say the "re-enter to save" rule exists specifically so the app never silently wipes a secret.

The LLM gateway already solved exactly this problem, in this repo, in `PUT /api/gateway/providers/{id}`:

- `api_key: Option<String>` with three meanings — absent leaves the column out of the `SET` list, `Some("")` is a deliberate clear, `Some(k)` replaces;
- the UI renders a configured key as `••••••••••• stored` behind an explicit **Replace key** button (`views/gateway/ProvidersView.tsx`);
- `a_scrubbed_read_written_straight_back_preserves_the_stored_key` drives the real `GET` body back through the `PUT` and fails on the revert.

Integrations should adopt the same contract. Once an omitted credential preserves the stored one, the warning banner and the separate bottom section have nothing left to protect, and the token field can sit next to **Name** on both screens.

## Notes for whoever picks this up

- The `credentials` blob is a map, not one field — the "keep the stored one" rule has to be decided for the whole blob (or per key) and written down.
- `services`/`auth` behaviour must not change; `auth` is already preserved in SQL by `registry.rs`'s `CASE WHEN` rewrite.
[...]

PR #575, opened 2026-09-08
feat: Slack app_token credential, inbound columns on the wire, and PUT /api/integrations/{id}/inbound (#566)

### LINKED ISSUE #566: Slack app_token credential, inbound columns on the wire, and PUT /api/integrations/{id}/inbound
Part of epic #562. Blocked by: #563.

## What

- The Slack credentials blob accepts an optional `app_token` (`xapp-…`) beside `bot_token`; `validate_slack` (`integration_credentials.rs:326-358`) checks its prefix when present and leaves the existing `bot_token`/`oauth` modes as they are (`AUTH_MODES` unchanged unless a mode is genuinely needed).
- A SQL-computed `has_app_token` boolean on the integration read, the way `has_credentials` is computed — the token itself is never selected.
- `inbound_enabled`, `inbound_status`, `inbound_error` on the integration read, in key order after the webhook fields.
- `PUT /api/integrations/{id}/inbound` with `{ "enabled": bool }` — a 422 when enabling with no stored app token, a 404 for an unknown id, a 400 for a non-Slack type; recorded in `parity/desktop_routes.json` under a `ROUTES` const; the write ends with a spawned registry reload and a #335-style log line with a test.
- `registry.rs`'s `HostingRow` projection carries `app_token` and `inbound_enabled` so #567 can read them; nothing else in the process can.

## Why

Decision 10 of the epic. The credential must follow the `has_credentials`/`auth_mode_sql` rule: what leaves SQLite is a boolean, never the bytes. A separate enable route rather than a field on the `PUT /api/integrations/{id}` body keeps the byte-exact integration write untouched and gives the UI's switch one call.

## How

- The three-valued credentials contract from #515 already preserves an omitted blob; `app_token` rides inside it, so a Replace on the form sends a new blob and an untouched field sends none.
- `desktop_routes.json` is set-equality against the union of `ROUTES` consts — append the new const to that union or the assertion silently weakens.
- Reload on the write, not on the read; the socket worker's start/stop is the reload's effect.

## Acceptance

- [ ] A scrubbed `GET /api/integrations/{id}` never contains the `xapp-` bytes (asserted over the response bytes, not a struct field).
- [ ] `PUT …/inbound {"enabled":true}` with no app token → 422 naming `app_token`; with one → 200 and `inbound_enabled: 1` on the next read.
- [ ] The route appears in `desktop_routes.json` and the set-equality test passes; a non-Slack integration answers 400.
- [ ] The write emits `integration inbound updated id=… enabled=…` at `info`, with a `testlog` assertion.
- [ ] Existing Slack rows read back with `has_app_token: false`, `inbound_enabled: 0`, empty status and error, and the frozen integration goldens are unchanged.

---
## WHAT

**Summary:** Slack gains an `app_token` credential, the integration read exposes the inbound state as booleans/strings, and `PUT /api/integrations/{id}/inbound` is the on/off switch.

**Scope:**
- `SlackCredentials` gains `app_token`; `validate_slack` (`src-tauri/src/native/integration_credentials.rs::validate_slack`) checks the `xapp-` prefix **only when present**, on both existing modes. `AUTH_MODES` is untouched — `app_token` is a field, not a mode, so the `auth_mode_sql` allowlist and the #513 reopen behaviour are unaffected.
- `ScrubbedIntegration` gains `has_app_token` (SQL-computed, `has_credentials_sql()`'s shape) and `inbound_enabled` / `inbound_status` / `inbound_error` (migration 39's columns, already in the schema).
- `PUT /api/integrations/{id}/inbound` with `{"enabled": bool}`: 422 when enabling with no stored app token, 404 unknown id, 400 non-Slack type, 200 otherwise.
- `registry.rs::HostingRow` gains `inbound_enabled` plus an `app_token()` accessor over the `credentials` blob it already selects.

**Affected Components:**
- `native/integration_credentials.rs` — `SlackCredentials`, `validate_slack`
- `native/integrations.rs` — `ScrubbedIntegration`, `scrubbed_select`/`has_credentials_sql`, new `ROUTES` const + `claims`/`serve`
- `native/integrations/registry.rs` — `HostingRow` projection and accessor
- `parity/desktop_routes.json` — the new route
- `native/mod.rs` — `ENDPOINTS` registration (if the route lands as its own endpoint)

**Acceptance Criteria:**
- [ ] A scrubbed `GET /api/integrations/{id}` and the list read never contain the `xapp-` bytes — asserted over the **response bytes**, not a struct field.
- [ ] `PUT …/inbound {"enabled":true}` on a Slack row with no app token → 422 naming `app_token`; with one → 200, and the next read shows `inbound_enabled: true`. Unknown id → 404, non-Slack type → 400.
- [ ] The route is in `parity/desktop_routes.json` and the set-equality test passes with the new `ROUTES` const added to the union.
- [ ] The write emits `integration inbound updated id=… enabled=…` at `info` (#335 shape) with a `testlog` assertion, and ends with a spawned registry reload.
- [ ] Pre-existing Slack rows read back `has_app_token: false`, `inbound_enabled: false`, empty `inbound_status`/`inbound_error`; the credentials round-trip test from #515 still passes with `app_token` inside the blob.
- [ ] `PUT /api/integrations/{id}` (the byte-exact integration write) is unchanged — no new field on its request body.

**Out of Scope:** the socket worker itself (#567), any UI (#569), a new `auth_mode`, and any change to the Telegram webhook routes.

## WHY

**Business Value:** Without a stored app-level token and a switch to turn it on, none of the Slack inbound epic can be reached from the product — this is the smallest piece that makes the rest configurable.

**Problem Statement:** Slack credentials today carry only `bot_token`/`oauth`, the three `inbound_*` columns landed in migration 39 (#563, merged) but nothing reads or writes them, and there is no route to enable inbound.

**Expected Impact:** One extra boolean-shaped column set on an existing read (no new round trip for the Integrations list badge) and one new write route. No behavioural change for any existing row.

**Context:**
- Parent: #562
- Blocked by: #563
- Blocks: #567
- Blocks: #569
- **Corrected during refinement (the original body was written before #563 merged):**
  1. *"in key order after the webhook fields"* is not implementable — `ScrubbedIntegration` has **no** webhook fields. Telegram's webhook state is a separate route, `GET /api/integrations/{id}/webhook/status` (`native/integrations/webhook.rs`). `ScrubbedIntegration`'s key order is **alphabetical**, because Go built the object from a map (`integrations.rs:185`). The new keys therefore slot as `enabled` < `has_app_token` < `has_credentials` < `id` < `inbound_enabled` < `inbound_error` < `inbound_status` < `name`.
  2. **Assumption taken (non-interactive refinement):** the inbound state goes on the integration read rather than on a sibling `…/inbound/status` route, because the Integrations list has to badge every row and a per-id status route would cost one request per integration. Veto here if the webhook route's shape is wanted for symmetry instead.
  3. *"the frozen integration goldens are unchanged"* — there is **no** integrations-response golden in `parity/`; the guard is the struct's declaration order plus the alphabetical rule documented on it. The acceptance criterion above states the real guard.
  4. `HostingRow` already selects the raw `credentials` column (it has since #315, for Slack OAuth), so `app_token` needs **no new field** — only an accessor. Adding a second secret field would widen the leak surface the module header argues against.
  5. Line reference `integration_credentials.rs:326-358` has shifted; `validate_slack` starts at line 324. Cited by symbol above.

## HOW

**Technical Approach:** Follow the `has_credentials` discipline exactly — what leaves SQLite is a boolean computed in SQL, never the bytes. Keep the enable switch off the byte-exact `PUT /api/integrations/{id}` body so that write and its goldens stay untouched; the switch is its own route and its own `ROUTES` const.

**Implementation Details:**
- `integration_credentials.rs`: add `app_token: String` to `SlackCredentials` (with `null_is_zero_value`, like every sibling). In `validate_slack`, after the mode match, `if !creds.app_token.is_empty() && !creds.app_token.starts_with("xapp-")` → `WriteError::validation("credentials.app_token", …)`. Do **not** touch `AUTH_MODES`; add a line to the doc comment saying why this one does not need it.
- `integrations.rs`: extend the scrubbed select with `json_extract(credentials,'$.app_token') IS NOT NULL AND json_extract(credentials,'$.app_token') != ''` shaped the way `has_credentials_sql()` is (reuse the helper's `TRIM`/`null`-literal handling — `the_two_has_credentials_rules_agree` is the precedent for keeping the SQL and the in-process rule in step), plus the three plain columns.
- New handler: `pub const ROUTES: &[(&str,&str)] = &[("PUT","/api/integrations/{id}/inbound")]`, `claims` matching on the chi-routed path, `serve` on the write path: existence check (404) → type check (400) → app-token check when enabling (422) → `db::blocking` update → `tokio::spawn` registry reload → `info!` log. Validation before mutation, per the write-path rule.
- `registry.rs`: add `inbound_enabled: bool` to `HostingRow` and `pub(crate) fn app_token(&self) -> Option<&str>` parsing `credentials`; the struct still derives neither `Serialize` nor `Debug`.

**Testing Strategy:** unit tests in `integrations.rs` (scrubbed read never contains `xapp-`; key order; defaults for legacy rows), in `integration_credentials.rs` (prefix accepted/rejected/absent), the route's four status codes, the `desktop_routes.json` set-equality test, and a `testlog` assertion on the log line.

**Documentation / Configuration:** a paragraph in `native/integrations/CLAUDE.md` on the new route and the `has_app_token` rule. No new dependency, no env var.

**Files to Create / Modify:** modify `native/integration_credentials.rs`, `native/integrations.rs`, `native/integrations/registry.rs`, `native/mod.rs`, `parity/desktop_routes.json`, `native/integrations/CLAUDE.md`.

**Risks:**
- Forgetting to add the new `ROUTES` const to the union → the set-equality assertion silently weakens. Mitigation: the acceptance criterion asserts it explicitly.
- A scrub regression putting the token on the wire. Mitigation: the test asserts over response **bytes**.
- `PUT /api/integrations/{id}` with a scrubbed blob must keep preserving `app_token` (#515's three-valued contract) — a test for the omitted-blob case.

**Complexity:** Medium (5–6 files, well-trodden patterns).

---
**Refined by:** AI Issue Refiner · **Date:** 2026-09-08 · **Status:** Ready

vibexp-agent:
Closes #566

## What

- `SlackCredentials` gains an optional **`app_token`** (`xapp-…`). `validate_slack` checks its prefix **only when present**, on both existing modes; `AUTH_MODES` is untouched.
- `ScrubbedIntegration` gains **`has_app_token`** (SQL-computed, `has_credentials`'s shape) and migration 39's **`inbound_enabled` / `inbound_error` / `inbound_status`**. Key order stays alphabetical: `enabled` < `has_app_token` < `has_credentials` < `id` < `inbound_enabled` < `inbound_error` < `inbound_status` < `name`.
- **`PUT /api/integrations/{id}/inbound`** with `{"enabled": bool}` — 404 unknown id, 400 non-Slack type, 422 naming `credentials.app_token` when enabling a row that stores none, 200 otherwise. Recorded in `parity/desktop_routes.json` through a new `integrations::ROUTES` const, added to the set-equality union.
- `registry::HostingRow` gains `inbound_enabled` and an `app_token()` accessor over the `credentials` blob it already selects, so #567 can read them and nothing else in the process can.

## Why

Decision 10 of epic #562. The credential follows the `has_credentials`/`auth_mode_sql` rule: what leaves SQLite is a boolean, never the bytes. A separate enable route rather than a field on `PUT /api/integrations/{id}` keeps that byte-exact write untouched and gives the UI's switch one call.

## How

- `has_app_token_sql()` is a nested `CASE` (`json_extract` raises on a column that is not JSON), narrowed by `json_type(...) = 'text'`, trimming the same four ASCII bytes `has_credentials_sql` does. `stores_an_app_token` is the in-process twin; `the_two_has_app_token_rules_agree` holds them to each other over 16 stored shapes.
- `HostingRow::app_token` is the **third** spelling of that rule and the only one that yields the value. `an_app_token_is_read_exactly_when_the_scrubbed_read_reports_one` pins it to the other two — the wire promising a token the worker cannot find is the disagreement that matters.
- The write refuses before it mutates, encodes its response before the `UPDATE`, and ends in `registry::reload_blocking` with the `#335`-shaped `integration inbound updated id=… enabled=…` log line.
- #515's three-valued `credentials` contract covers `app_token` too: a `PUT` that omits the blob preserves it, and the response reports what is stored rather than what the request implies.

### Deviations from the issue's HOW

- `HostingRow::app_token` returns **`Option<String>`**, not `Option<&str>`. A JSON string is escaped in the blob, so the decoded token is not always a slice of it, and a `&str` cannot be borrowed out of a parsed `serde_json::Value`. Every other credential accessor in that module hands back an owned value.
- The 200 body is `{"inbound_enabled": <bool>}` — the column under the name the integration read spells it with — so a caller holding a `ScrubbedIntegration` can patch the one field. The issue did not specify a body.
- `inbound_status`, `inbound_error` and `updated_at` are deliberately **not** touched by this write. The first two are the worker's (#567) and a disable that cleared them would erase the reason the user is looking at; `updated_at` tracks the integration record that `PUT /api/integrations/{id}` writes.

## Tests

`the_two_has_app_token_rules_agree` · `a_scrubbed_read_never_carries_the_app_token` (over response **bytes**, both reads) · `a_row_written_before_the_inbound_columns_reads_back_empty` · `the_inbound_switch_refuses_before_it_writes` (404/400/422 + nothing written + disable always allowed) · `enabling_inbound_shows_on_the_next_read` (200 body, next read, `updated_at`/status/error untouched, `testlog` assertion on the log line) · `a_put_that_omits_credentials_preserves_the_app_token` · `a_slack_app_token_is_checked_only_for_its_prefix_and_only_when_present` · `an_app_token_is_read_exactly_when_the_scrubbed_read_reports_one` · `claims` and the wire-shape golden extended.

## Out of scope

The socket worker (#567), any UI (#569), a new `auth_mode`, and the Telegram webhook routes.


vibexp-agent:
**Automated pipeline hand-off — ready for a human, not auto-merged.**

CI is green on `ec3b8c8` (all 8 checks, Windows unit tests included) and every acceptance criterion of #566 is met and tested. The automated review ran its **three** rounds and never returned APPROVE, so the auto-merge gate — which requires an APPROVE within three rounds — did not open. That is the only reason this is not merged.

What the three rounds found, and what changed:

| round | finding | fix |
|---|---|---|
| 1 | `PUT /api/integrations/{id}` could leave `inbound_enabled: true` on a row that no longer stores an app token — the exact state `update_inbound`'s 422 exists to prevent, reachable because #515 makes a sent blob replace wholly and the Slack form emits only its mode's fields | `clears_inbound` on the replace arm (`bfe4daa`) |
| 1 | `has_app_token`'s rustdoc had been appended onto `has_credentials`'s block, leaving that field undocumented | split (`bfe4daa`) |
| 1 | `src/lib/types.ts` did not mirror the four new wire fields, as #515 did in its own commit | mirrored, type-only (`bfe4daa`) |
| 1 | `native/CLAUDE.md`'s registry index did not name the route | named (`bfe4daa`) |
| 2 | The same class on the **type** axis: a `PUT` changing `type` away from `slack` stranded `inbound_enabled` with no way back, because the 400 refuses a non-Slack row whichever way the switch is moved | `clears_inbound` extended to `type != "slack"`, applied on **both** `UPDATE` arms (`1ebd89b`) |
| 2 | `has_app_token`'s doc claimed the SQL reads `false` for every non-Slack row; it is not `type`-scoped | corrected (`1ebd89b`) |
| 3 | Round 2's own bullet split in `integrations/CLAUDE.md` left a dangling "It is desktop-only…" attributing the `desktop_routes.json` rule to a SQL helper | moved back onto the route, and the dropped "disabling is never refused on the credential axis" sentence restored (`ec3b8c8`) |

Round 3's verdict was REQUEST CHANGES on that documentation item alone; its own summary said "functionally this PR is ready" and the verdict was REQUEST CHANGES "strictly because the skill's gate forbids approving with any unapplied `now` item". That item is now applied and CI is green on it, but the round budget is spent — so the PR is handed to you rather than merged by the bot.

Two judgement calls made in place of a human, both recorded in the PR body:
- `HostingRow::app_token` returns `Option<String>`, not the issue's `Option<&str>` — a JSON string is escaped in the blob, so the decoded token is not always a slice of it.
- The 200 body is `{"inbound_enabled": <bool>}`, the column under the name the integration read spells it with; the issue specified no body.

PR #581, opened 2026-09-08
docs: Slack Socket Mode guide, troubleshooting and a live end-to-end test (#571)

### LINKED ISSUE #571: Slack Socket Mode: user guide, troubleshooting, and a live end-to-end test
Part of epic #562. Blocked by: #568, #569, #570.

## What

- `docs/user-guide.md` → Integrations → Slack: a **Socket Mode** subsection with the app manifest to paste (Socket Mode on; bot scopes `app_mentions:read`, `chat:write`, `channels:read`; event `app_mention`; app-level token with `connections:write`), the flow (create app from manifest → install → copy `xoxb` and `xapp` → paste both → turn on Socket Mode → invite the app to a channel → add a rule), how threads map to chats, the one-way rule, and the security note (anyone in a configured channel can run the rule's agent).
- `docs/troubleshooting.md` → Slack: the three status values, the `apps.connections.open` 401 case, the "mention ignored" cases (no rule, disabled specific rule, thread not started by Agento, bot message), and the log lines to grep.
- A `#[ignore]`d live test `tests/slack_socket_live.rs` that, given `AGENTO_SLACK_TEST_*` env vars, connects to a real workspace, posts a mention through the Web API, and asserts a reply and a resume in the thread.
- `.claude/skills/local-verify`: a recipe for the live test and for reading `inbound_status`.

## Why

The epic's security posture lives in the guide, not the code; and a feature whose only proof is a fake server has never met Slack's real envelope shapes.

## Acceptance

- [ ] A fresh Slack app created from the manifest in the guide connects and answers on the first try (recorded in the PR).
- [ ] The live test passes against that app and is skipped without the env vars.
- [ ] `docs/` mention every status value and every ignore case the code has.

---
## WHAT

**Summary:** Document Slack Socket Mode end to end — setup guide, troubleshooting, a local-verify recipe — and prove it against a real workspace with an `#[ignore]`d live test.

**Scope:**
- `docs/user-guide.md` → Integrations → **Slack Socket Mode** subsection: the app manifest to paste (Socket Mode on; bot scopes `app_mentions:read`, `chat:write`, `channels:read`; event subscription `app_mention`; app-level token with `connections:write`), the flow (create app from manifest → install → copy `xoxb` and `xapp` → paste both → turn on Socket Mode → invite the app to a channel → add a trigger rule), how a Slack thread maps to one chat, the one-way rule (UI turns are not mirrored back), and the security note: **anyone who can post in a configured channel can run that rule's agent with its permission mode and working directory.**
- `docs/troubleshooting.md` → a **Slack** section: every `inbound_status` value the code can store, the `apps.connections.open` 401 case, the "mention ignored" cases (no matching rule, the matching rule disabled, thread not started by Agento, message from a bot), and the exact log lines to grep.
- `src-tauri/tests/slack_socket_live.rs` — `#[ignore]`d; given `AGENTO_SLACK_TEST_*` env vars it connects to a real workspace, posts a mention through the Web API, and asserts both a threaded reply and a resume in that thread.
- `.claude/skills/local-verify` — a recipe for running the live test and for reading `inbound_status` back over `/api`.

**Affected Components:** `docs/user-guide.md` (Slack rows at `:246`, integration setup at `:253`), `docs/troubleshooting.md` (**no Slack section exists today**), `src-tauri/tests/` (new file), `.claude/skills/local-verify/`.

**Acceptance Criteria:**
- [ ] A **fresh** Slack app created from the manifest exactly as written in the guide connects and answers a mention on the first try, with no undocumented step — recorded in the PR.
- [ ] `cargo test --test slack_socket_live -- --ignored` passes against that app; the plain `cargo test` run skips it, and it skips (not fails) when the `AGENTO_SLACK_TEST_*` vars are unset.
- [ ] Every `inbound_status` value the worker can write and every ignore branch the handler has appears in `docs/troubleshooting.md` — checked against the code, not from memory.
- [ ] The security note is in the guide, states who can trigger a run, and names the permission mode / working directory a rule can carry.
- [ ] The live test names no secret in any failure message or log line.
- [ ] `local-verify` gains the recipe and it works as written.

**Out of Scope:** in-app help text, screenshots of the Slack UI, and any code change to the worker — if the docs cannot describe a behaviour truthfully, that is a bug filed against #567/#568, not a fix here.

## WHY

**Business Value:** The epic's security posture lives in the guide, not in the code — a user who does not know that a channel is an execution surface will configure it wrongly. And a feature whose only proof is a fake server has never met Slack's real envelope shapes.

**Problem Statement:** `docs/troubleshooting.md` has no Slack content at all and the user guide names Slack only as an outbound tool provider. Nothing tells a user how to create the app, which token goes where, or what a red badge means.

**Expected Impact:** Setup succeeds unattended from the guide; a failing socket is diagnosable from the docs; real Slack envelope shapes are exercised before release.

**Context:**
- Parent: #562
- Blocked by: #567, #568, #569, #570 — **the original body listed only #568**; #567 (the worker) is what produces the status values this issue documents, and #569/#570 are the surfaces it describes, so all four are prerequisites. Documenting statuses that do not exist yet is the failure mode here.
- Verified against `main` @ `4184eea`: `docs/` is `architecture.md`, `development.md`, `installation.md`, `README.md`, `releasing.md`, `troubleshooting.md`, `user-guide.md`; `src-tauri/tests/` already holds `trigger_run.rs`, `headless_resume.rs` and several `*_live.rs` suites — `insights_live.rs` / `search_live.rs` are the `#[ignore]`d-with-env-vars pattern to copy. `.claude/skills/local-verify` exists.
- **`src-tauri/tests/` does not compile on Windows** (the fake-CLI suites use `PermissionsExt`), so this test is Unix-only in CI by construction; that is the existing state, not a regression.

## HOW

**Technical Approach:** Docs are written **from the code as merged**, not from the issue bodies — read the worker's status writes and the handler's ignore branches and enumerate what is actually there. The live test follows the existing `*_live.rs` shape: `#[ignore]`, read config from env, skip cleanly when unset, and drive real HTTP.

**Implementation Details:**
- Manifest goes in the guide as a fenced YAML block, copy-pasteable whole, with each scope annotated with why it is needed.
- Troubleshooting entries are `symptom → what it means → what to do`, each naming the grep-able log line.
- `slack_socket_live.rs`: read `AGENTO_SLACK_TEST_BOT_TOKEN`, `_APP_TOKEN`, `_CHANNEL` (document the exact names in the file header and in `local-verify`); start an instance against a scratch `HOME`; configure the integration + one rule over `/api`; post a mention via `chat.postMessage`; poll `conversations.replies` with a bounded timeout for the agent's reply; post a follow-up in the thread and assert the chat resumes rather than a second chat being created; tear the fixtures down in both the pass and fail paths.
- Every HTTP client comes from `native::http::client_builder()`.

**Testing Strategy:** The live test *is* the test. Also confirm `cargo test` (no `--ignored`) is unaffected, and that the suite compiles with the env vars unset.

**Documentation / Configuration:** This issue is the documentation. New env vars are test-only: `AGENTO_SLACK_TEST_BOT_TOKEN`, `AGENTO_SLACK_TEST_APP_TOKEN`, `AGENTO_SLACK_TEST_CHANNEL`. No new runtime deps.

**Files to Create / Modify:** create `src-tauri/tests/slack_socket_live.rs`; modify `docs/user-guide.md`, `docs/troubleshooting.md`, `.claude/skills/local-verify/SKILL.md`.

**Risks:**
- **Docs drifting from the code** if written from the issues — mitigated by enumerating statuses and ignore branches from the merged source and citing them in the PR.
- A flaky live test blocking CI — mitigated by `#[ignore]` plus bounded polls and a clean skip when unconfigured.
- Leaking a workspace token into CI output — never print token values; assert on shapes.
- Scopes in the manifest drifting from what the worker actually calls — cross-check each documented scope against a real API call in the code.

**Complexity:** Low (docs + one test file, no production code)

---
**Refined by:** AI Issue Refiner · **Date:** 2026-09-08 · **Status:** Ready

vibexp-agent:
Closes #571

## What

Documents Slack **Socket Mode** end to end and proves it against a real workspace.

| file | change |
|---|---|
| `docs/user-guide.md` | `### Slack Socket Mode` under Integrations — the app manifest, the setup flow, thread↔chat mapping, the one-way rule, the security note |
| `docs/troubleshooting.md` | a new `## Slack` section — every `inbound_status` value, the `apps.connections.open` refusal, every ignore branch, the log lines to grep |
| `src-tauri/tests/slack_socket_live.rs` | new, `#[ignore]`d — a real Socket Mode connection, a real mention, a real threaded reply, and a resume |
| `.claude/skills/local-verify/SKILL.md` | the recipe for running it and for reading `inbound_status` back over `/api` |

## How

Everything documented was **enumerated from the code as merged**, not from the
issue bodies. Specifically:

- **Five `inbound_status` values, not three.** `socket.rs:81-88` defines
  `connecting`, `connected`, `reconnecting` and `error`; the fifth is the empty
  string, which the *registry* writes (boot, a row that stops being startable, a
  failed start, a Slack row that gets no socket worker) and which the UI renders
  as *Not running* via `catalog.ts::inboundState`. The worker never clears its
  own row, so a doc that offered "wait for it to clear" would be wrong.
- **Seven ignore branches**, each with its verbatim log line: no rule for the
  channel and a disabled channel-specific rule (`select_rule.rs` — a rule naming
  the channel wins *even when disabled*, so disabling silences rather than falls
  back), a thread Agento did not start, a message from a bot, nothing said to the
  bot, a redelivered `event_id`, and the app simply not being in the channel
  (Slack sends nothing, so there is no line at all). Six of the seven log at
  `debug`, which is itself documented — at `info` the section would look like it
  was lying.
- **The manifest's scopes are cross-checked against real calls.**
  `app_mentions:read` for the event, `chat:write` for `inbound.rs:487`
  (`chat.postMessage`), `channels:read` for `inbound.rs:429`
  (`conversations.info`, the chat title). `auth.test` and `chat.getPermalink`
  need no scope. Nothing else is requested.
- **`error` is a report, not a stop** (`socket.rs`, and `tests/slack_socket.rs`
  pins it), and a refusal never clears the stored credential — both are stated,
  because both change what a user should do.

### The live test

`cargo test --test slack_socket_live -- --ignored --nocapture` seeds a temp
database with one integration and one channel-scoped rule, builds the
**production** `inbound::handler` and `SocketOptions::default()`, opens a real
socket, posts a mention, waits for the agent's threaded reply, posts a follow-up
in the thread, and asserts two mentions produced **one** `chat_sessions` row and
one `inbound_threads` row keyed on the first mention's `ts`. The thread is
deleted on both the pass and the fail path, each message with the token that
posted it.

## Deviations from the HOW, and why

Two, both recorded on the issue as well.

1. **Four env vars, not three.** The HOW named `AGENTO_SLACK_TEST_BOT_TOKEN`,
   `_APP_TOKEN` and `_CHANNEL`. A mention posted with the **bot** token carries
   `bot_id` and is dropped by `socket.rs:853` before the handler sees it — that
   branch is correct and is one of the things this PR documents — so the driver
   must be a human, which needs `AGENTO_SLACK_TEST_USER_TOKEN` (`xoxp-`, with
   `chat:write` and `channels:history`). It turns out better this way: every call
   the *test* makes goes out on the user token, so the app keeps exactly the
   three scopes the guide's manifest lists and a missing scope fails the run
   instead of being masked.
2. **Library-level, not "an instance against a scratch `HOME` configured over
   `/api`".** No test in `src-tauri/tests/` starts a server or spawns the binary;
   they link `agento_lib` and drive the subsystem against a temp database, and
   `paths::data_dir` is process-global and debug-pinned so in-process `/api`
   dispatch would write to the developer's own dev instance. Seeding the temp
   database directly is what `tests/slack_socket.rs` and `slack/inbound/tests.rs`
   already do. Nothing is lost: the socket, the envelopes, the handler, the
   Slack calls and the reply are all the real ones.

Two small truthfulness fixes fell out of the sections being edited: the Slack
row in the Integrations table said Slack was outbound-only, the paragraph beside
it said Slack is OAuth (it offers both modes), and `local-verify`'s "eight
integration tests" was already stale at 21 and would have been more so at 22.

## Verification

| gate | result |
|---|---|
| `npm run build` | pass |
| `cargo fmt --check` | pass |
| `cargo clippy --all-targets -- -D warnings` | pass |
| `scripts/check-claude-md-size.sh` | pass (no `CLAUDE.md` touched) |
| `cargo test --test slack_socket_live` | 1 passed, 1 ignored — the live test is not run |
| CI on GitHub | all 9 checks green, including **Site check**, which is the heading-anchor validator |
| `cargo test --test slack_socket_live -- --ignored` with no env vars | skips: `skipping: set AGENTO_SLACK_TEST_BOT_TOKEN, _APP_TOKEN, _USER_TOKEN and _CHANNEL` |

`a_failure_message_never_names_a_token` is the one non-ignored test and is a real
regression guard: make `scrub` a no-op and it fails.

**The live acceptance criterion is not met by this PR and cannot be.** Creating a
fresh Slack app from the manifest needs a browser and workspace-admin consent,
and no `AGENTO_SLACK_TEST_*` credentials exist on the machine this was written
on. So criterion 1 ("a fresh app from the manifest connects and answers on the
first try, recorded in the PR"), the passing half of criterion 2, and criterion 6
("`local-verify` … works as written") are **open and need a human run**. That is
why this is not self-merged.

## Review

**Seven automated review rounds; the seventh returned ✅ APPROVE on the head that
is here now (`61435d6`), with CI green on it.** 29 findings across the rounds,
all applied — the maintainer's round cap was lifted so the loop could run to an
approval rather than stopping at three.

The findings worth knowing about, because each was a documented claim that was
wrong rather than a style note:

- The live test seeded `filter_chat_ids` with a bare channel id.
  `dispatcher::decode_list` answers an empty `Vec` for anything it cannot
  parse, so the rule was the match-every-channel default — and passed anyway.
- **The suite claimed to keep the manifest honest and did not, for one scope of
  the three.** `Inbound::channel_name` swallows a refused `conversations.info`
  and falls back to the raw channel id, so an app without `channels:read` passed
  green. The chat's stored title is now the gate: `[Slack] #<name>` when the call
  worked, `[Slack] #<id>` when it did not.
- **The resume assertions could have held without a resume.** The wait was
  keyed on a count of bot messages, and one answer over 4000 characters is
  posted as two — so a long first answer satisfied it before the second run
  finished. Each wait is now keyed on the `ts` of the mention that provoked it.
- **An OAuth-mode Slack row can turn Socket Mode on, reach *Connected*, and
  never receive a mention** — `SLACK_SCOPES` has no `app_mentions:read` and
  nothing on the start path gates on `auth_mode`.
- The pasted manifest is the Socket Mode minimum, while the Integrations table
  on the same page advertises the outbound tools. Five of the seven need scopes
  it does not grant, and `search_messages` needs one no bot token can hold.
- Three of the four causes of "Sorry, something went wrong." are logged
  *instead of* `slack mention matched`, not after it, so the chat id the
  section told you to grep for does not exist for them.
- Both `already processed` and `no rule for the channel` have a second,
  database-failure cause whose remedy is the opposite of the configuration
  advice they carried.
- "Reading the logs" said no prompts are logged; `slack mention prompt` logs
  one at `debug`, and is the only such site in the tree.

One finding was deferred: the trigger-rule form offers **Prefix** and
**Keywords** for Slack rules and `inbound.rs` consults neither — #582. The
guide's security note states the asymmetry as the interim mitigation.

One nit was deliberately not taken. `await_reply_after` returns on the first bot
message newer than its anchor, so in principle a straggling chunk of a
multi-part first answer could end the second wait early. `FIRST_PROMPT` asks for
one word, so a second chunk cannot arise; and the resume claim is carried by the
`chat_sessions` / `inbound_threads` counts, not by which message the wait
returned, so an early return could not fake it either way.

One thing found and **not** in scope here: `integrations/oauth/mod.rs`'s
`SLACK_SCOPES` requests `search:read` as a **bot** scope, which Slack grants only
to user tokens. Untouched by this diff and unprovable without a workspace, but
worth a separate look.

## What the maintainer has to do

Create a Slack app from the manifest exactly as written, export the four
`AGENTO_SLACK_TEST_*` variables, and run

```bash
cargo test --test slack_socket_live -j 4 -- --ignored --nocapture
```

That closes acceptance criteria 1, 2 and 6 in one go. The recipe, including
what each token needs and what it costs, is in
`.claude/skills/local-verify/SKILL.md` under *Does Slack Socket Mode actually
reach a real workspace?*.


vibexp-agent:
Round-2 review findings are applied in `04f06f6`.

The one `ticket` finding from round 1 — the trigger-rule form offers **Prefix** and **Keywords** for Slack rules and `inbound.rs` never consults either — is **deferred to #582**. This PR documents the asymmetry in `docs/user-guide.md` → *Slack Socket Mode* → *Who can run an agent this way* as the interim mitigation, since #571 puts worker code changes out of scope.

vibexp-agent:
Round 7 returned **✅ APPROVE** on `61435d6`, and CI is green on that head (CI, Site check and CodeQL all success, nothing pending).

Seven rounds, 29 findings, all applied. Round 6 approved `d9df21d` with three nits; two were doc corrections worth taking, so they were applied and round 7 ran to put the approval on the tree that is actually here. The third nit — `await_reply_after` returning on the first bot message newer than its anchor — was deliberately declined: a second chunk needs an answer over 4000 characters and `FIRST_PROMPT` asks for one word, and the resume claim is pinned by the `chat_sessions`/`inbound_threads` counts rather than by which message the wait returned.

**Still not merged, and not because of the review.** Acceptance criteria 1, the passing half of 2, and 6 need a fresh Slack app created from the manifest in a real workspace, which needs a browser and workspace-admin consent. Everything else is done and green. The run is four env vars and one command — see the PR body's *What the maintainer has to do*.

vibexp-agent:
Coordination note from #584 (issue #582) — **this paragraph has to change before #581 merges.**

`docs/user-guide.md`, under `#### Who can run an agent this way`, this branch currently says:

> **The channel list is the only filter a Slack rule applies.** The rule form also offers a **Prefix** and **Keywords** … but a mention over Socket Mode is matched on its channel alone, so neither narrows what triggers a run. Do not reach for them as a safety measure.

#584 makes that false. Slack's inbound path now gates the already-selected rule through `trigger::match_rule`, so **Prefix and Keywords do narrow what triggers a run**, checked against the text with the `@mention` removed — and they gate a reply inside a thread Agento started as well as the mention that opened it. Empty prefix and empty keywords behave exactly as this paragraph describes today.

Leaving the wording as-is would ship user-facing *security* guidance on `main` that contradicts the code, whichever order the two land in. #582's spec asked for the two to be coordinated; recording it here rather than editing this branch, since it is not mine to edit.

Issue #580, opened 2026-09-08
src-tauri/src/native/CLAUDE.md is 11 characters under the 40,000 CI limit


vibexp-agent:
Split out of the review of #579 (issue #570).

## What

`src-tauri/src/native/CLAUDE.md` is **39,989 characters**. `scripts/check-claude-md-size.sh` fails at 40,000 and CI runs it on every PR, so the next change that adds a single line to that file fails CI on something unrelated to itself.

It was 39,830 before #579, which added a two-line note about the chats `inbound` key — deliberately compressed to one bullet for exactly this reason, with the full reasoning put in `ChatSession::inbound`'s doc comment instead.

## Why

The budget exists because Claude Code truncates a `CLAUDE.md` past 150k and the root file was 304k when #553 split it. That is still the right rule; this file has simply grown back to the edge of it. The failure mode is bad: an unrelated PR goes red on a check the author did not touch, and the cheapest way out is to shorten someone else's note rather than one's own.

## How (sketch)

Move subsystem-deep material out of `native/CLAUDE.md` and into a `CLAUDE.md` beside the code it describes, per the root file's *Where the notes live*. Candidates worth measuring first: the wire-format traps list (some entries are per-module and already have a module doc), and the layout table, which duplicates what the tree already shows.

Not a trim-for-trim's-sake pass — the root file is explicit that a pinned-divergence list *is* the specification and must move intact rather than be summarised.

---
## WHAT

**Summary:** Make the repository hold exactly one `CLAUDE.md`, at the root, by relocating all 20 nested subsystem notes into `docs/internal/` and pointing the root file at them — which also removes the 39,989-character file that is 11 characters from failing CI.

**Scope:**
- Move each of the 20 tracked nested `CLAUDE.md` files (326,277 characters in total) to `docs/internal/<area>.md`, content-preserving.
- Rewrite the root file's *Where the notes live, and how to add to them* table to point at the new paths, and state the new rule ("one `CLAUDE.md`, at the root; subsystem notes are `docs/internal/*.md`, read before working in that area").
- Update every cross-reference to a nested `CLAUDE.md` — in the root file, in the relocated files themselves, in `docs/`, and in `//!` module docs.
- Extend `scripts/check-claude-md-size.sh` so it enforces both the 40,000 limit **and** that no `CLAUDE.md` exists outside the root.

**Affected Components:**
- `CLAUDE.md` — the notes table and the "where a new note goes" rule.
[...]

Issue #565, opened 2026-09-07
Trigger rules apply their execution settings; most-specific rule selection for Slack channels


vibexp-agent:
Part of epic #562. Blocked by: #563.

## What

- The trigger dispatcher (`native/trigger/dispatcher.rs`) honours a rule's `model`, `working_directory`, `settings_profile_id`, `permission_mode` and `timeout_minutes` when it builds the run, for Telegram and for the Slack path #568 adds. Empty means what it means today (agent default; 5-minute timeout).
- A second selection function, `select_rule_for_channel(rules, channel_id)`, for Slack: the enabled or disabled rule whose `filter_chat_ids` names the channel wins over a rule with none; **a disabled channel-specific rule means silence** (no fall-through to the default); with no specific rule, the enabled default (empty `filter_chat_ids`) applies; no rule at all is silence. Telegram keeps `find_matching_rule` unchanged.

## Why

Decisions 3 and 4 of the epic. A workspace-wide default makes every channel the app is in respond; a channel needs an off switch, and "disable the specific rule" is the one that reads naturally in the rules list. Changing Telegram's first-enabled-by-creation selection would change stored behaviour for existing users, so the new rule is Slack's only.

## How

- Thread the five fields from the matched rule into `create_trigger_session` (title, working dir, model, profile, mode are already `insert_session` parameters, passed as empty strings today at `dispatcher.rs:255-256`) and into the `RunSpec`; the timeout replaces the flat `RUN_TIMEOUT` when non-zero.
- `select_rule_for_channel` lives beside `match_rule.rs` and is pinned by a vector file the way `parity/trigger_match_vectors.json` pins matching.

## Acceptance

- [ ] Vectors: specific enabled beats default; specific disabled → `None` even with an enabled default; no specific → enabled default; disabled default only → `None`; two specific rules for one channel → the earlier `created_at`.
- [ ] `parity/trigger_match_vectors.json` and every Telegram dispatcher test are byte-identical and green.
- [ ] A Telegram rule with `working_directory` set runs the CLI in that directory (fake-CLI test records cwd) and one with `timeout_minutes: 1` times out at one minute.
- [ ] A rule with `permission_mode: "plan"` reaches `build_options` as `plan`.

---
## WHAT
**Summary:** The trigger dispatcher honours a rule's execution settings when it builds a run, and gains a **second, Slack-only** rule-selection function whose semantics are most-specific-first with a per-channel off switch.

**Scope:**
[...]

PR #584, opened 2026-09-09
feat: Slack inbound honours a trigger rule's prefix and keywords (#582)

### LINKED ISSUE #582: Slack trigger rules show Prefix and Keywords, and the inbound path ignores both
## What

The trigger-rule form renders **Prefix** and **Keywords** for every provider (`src/views/IntegrationsView.tsx:2038-2054`), but the Slack inbound path never consults either. `Inbound::rule_for` (`src-tauri/src/native/integrations/slack/inbound.rs:184-199`) picks a rule with `select_rule_for_channel` and nothing else, and `Inbound::turn` never calls `dispatcher::match_rule` — so `filter_prefix` and `filter_keywords` are read off the row (`native/trigger/dispatcher.rs:252-254`) and then dropped.

Telegram, which the form was built for, does apply them.

## Why it matters

A user who sets `Prefix: /ask` to narrow what triggers an agent gets no narrowing at all: every mention in a configured channel still starts a run, with that rule's permission mode and working directory. That is a control that looks like a safety measure and is not one — the failure is silent and in the direction of running more, not less.

## What to decide

Two ways, and it is a product call rather than a bug fix:

- **Hide them for Slack** — gate the two inputs on the provider, the way `supportsInbound` / `triggerTargets` already vary the form in `src/views/integrations/catalog.ts`. Smallest change; the honest one if channel-scoping is meant to be the whole filter.
- **Honour them in `inbound.rs`** — run the mention text through the same matcher Telegram uses. More useful, and it changes worker behaviour, so it needs its own thought about what "prefix" means for a message that starts with `<@Uxxxx>`.

## Context

Found by the automated review of #581 (issue #571), which documents this asymmetry in `docs/user-guide.md` → *Slack Socket Mode* → *Who can run an agent this way* as the interim mitigation. #571 put worker code changes explicitly out of scope, hence this issue rather than a fix there.

---
## WHAT

**Summary:** Make Slack's inbound path honour a trigger rule's `filter_prefix` and `filter_keywords` against the mention text with the bot mention stripped, so the two inputs the form already shows mean what they say.

**Scope:**
- Apply prefix and keyword filtering to Slack app-mention events, on the **already-selected** rule — channel selection stays `select_rule_for_channel`'s most-specific-wins (#565); the filters are a gate on that rule, never a reason to fall through to another one.
- Filter against the text with `<@Uxxx>` mentions removed, not the raw event text, and use the prefix-stripped remainder as the prompt.
- Silence stays silence: a mention that fails the filter starts no run and posts nothing.
- Update the Slack trigger-rule help text so Prefix and Keywords describe Slack's behaviour.

**Decision taken:** *honour them*, not *hide them*. Hiding the two inputs removes a control users reasonably want (a channel where only `/ask …` should start an agent) and leaves stored non-empty filters on existing Slack rows silently inert. Honouring them makes the form's promise true and moves the failure in the safe direction — fewer runs, not more. Recorded here because the issue posed it as an open product call.

**Affected Components:**
- `src-tauri/src/native/integrations/slack/inbound.rs` — `accept` / `rule_for` / `run`: strip the mention before filtering, and gate on `match_rule`.
- `src-tauri/src/native/trigger/match_rule.rs` — reused as-is; a Slack-shaped entry point may be added beside `match_rule`, but the Telegram/Go-pinned function itself is not modified.
- `src/views/integrations/catalog.ts` — Slack's trigger help strings.

**Acceptance Criteria:**
- [ ] A rule with `filter_prefix = "/ask"` on a Slack channel: `@bot /ask what is the status` runs with prompt `what is the status`; `@bot what is the status` starts nothing and posts nothing.
- [ ] Prefix matching sees the stripped text, so it works regardless of whether the user typed `@bot /ask …` or `/ask @bot …`.
- [ ] A rule with `filter_keywords = ["deploy"]`: a mention containing `Deploy` (any case) runs; one without it does not.
- [ ] Empty prefix and empty keywords behave exactly as today — every mention in a matched channel runs.
- [ ] A mention that matches the channel but fails the rule's filters does **not** fall through to a broader rule; it is dropped, and the drop is logged at `debug` with no Slack-derived text above `debug`.
- [ ] A bare `@bot /ask` with nothing after the prefix starts nothing (empty prompt is not a match), matching `match_rule`'s existing rule.
- [ ] Telegram's behaviour is byte-identically unchanged: `parity/trigger_match_vectors.json` and `native::trigger::match_rule` tests still pass untouched.
- [ ] Unit tests in `inbound.rs` cover: prefix hit, prefix miss, keyword hit, keyword miss, both empty, prefix-only-text, mention in the middle of the text.

**Out of Scope:**
- Changing `select_rule_for_channel` or the most-specific-rule semantics from #565.
- Any change to Telegram's dispatcher or to `match_rule`'s pinned Go-parity behaviour.
- Regex or pattern filters, per-user allowlists, or any new filter field.
- Migrating or rewriting stored rows.

## WHY

**Business Value:** A filter that is displayed and stored but ignored is worse than no filter — the user believes they have narrowed what can start an agent and they have not.

**Problem Statement:** The trigger-rule form renders Prefix and Keywords for every provider (`src/views/IntegrationsView.tsx:2038-2054`), and the Slack path never reads either. `Inbound::rule_for` (`slack/inbound.rs:184-199`) selects with `select_rule_for_channel` alone, and nothing on the Slack path calls `dispatcher::match_rule`, so `filter_prefix` and `filter_keywords` are loaded off the row (`native/trigger/dispatcher.rs:252-254`) and dropped. Telegram, which the form was built for, applies both.

**Expected Impact:** Every mention in a configured channel currently starts a run under that rule's permission mode and working directory, however the filters are set. After this, a user can scope a channel to a prefix or a keyword and have it hold. The failure mode changes from "silently runs more than asked" to "runs exactly what was asked".

**Context:**
- Found by the automated review of #581 (issue #571). **#581 is still open**, so the interim mitigation it documents (`docs/user-guide.md` → *Slack Socket Mode* → *Who can run an agent this way*) is not on `main` yet; that section describes the current, unfiltered behaviour and must be corrected when this lands — coordinate on ordering.
- #571 put worker code changes explicitly out of scope, which is why this is its own issue.
- Related: #565 (most-specific rule selection for Slack channels).

## HOW

**Technical Approach:** The mention text is only usable for filtering **after** `strip_mention`, and today that runs in `Inbound::run` (`inbound.rs:292`), long after `accept` has picked a rule and enqueued a job. So the ordering changes: resolve the bot user id and strip the mention *before* the filter decision, then gate the selected rule through the existing `match_rule` and carry its returned prompt forward instead of re-deriving one.

**Implementation Details:**
- `slack/inbound.rs`:
  - In `accept`, after `rule_for(&mention.channel)` returns a rule, obtain `bot_user_id().await` and compute `strip_mention(&mention.text, bot_user_id)`. Drop (log at `debug`) when the id is unavailable or the stripped text is empty — `run` already treats an empty prompt that way, so the check simply moves earlier.
  - Gate on `match_rule(&rule.filters, &stripped, &mention.channel)`. `Some(prompt)` → enqueue with that prompt on the `Job`; `None` → log at `debug` and return, with no fall-through to another rule. Note `match_rule` also re-checks `chat_ids`, which `select_rule_for_channel` has already satisfied for a channel-specific rule; for a catch-all rule (`chat_ids` empty) the matcher reads it as "everything", so the two agree — assert this with a test rather than special-casing it.
  - `Job` gains the resolved `prompt`; `run` uses it instead of calling `strip_mention` again. Keep one code path that produces the prompt, so the text that was filtered is exactly the text that runs.
  - **Keywords are matched against the stripped text**, which diverges from Telegram deliberately: `match_rule`'s header pins keywords to the *whole* message because Go did, but on Slack "the whole message" always begins with `<@Uxxx>`, a bot id no user typed and no keyword can sensibly target. Say this in a comment at the call site, in the shape the rest of the module uses (rule → why → where enforced).
- `src/views/integrations/catalog.ts` — Slack's entry gains prefix/keyword help wording matching the behaviour ("checked after the `@mention` is removed"). Do not add a new capability flag; nothing is being hidden.

**Testing Strategy:**
- Unit tests in `slack/inbound.rs`'s existing `mod tests` over the filter decision (a helper taking `(filters, raw_text, bot_user_id)` keeps them free of the socket), one per acceptance criterion above.
- `select_rule.rs` and `match_rule.rs` suites must pass unchanged — that is the guard that Telegram and Go parity did not move.
- Manual: two rules on one Slack integration (a channel-specific one with `/ask`, a catch-all without), verify the four mention shapes in a real workspace.

**Documentation / Configuration:** `docs/user-guide.md`'s Slack Socket Mode section — replace the interim "prefix and keywords do not apply on Slack" mitigation once #581 has merged. `src-tauri/src/native/integrations/slack/CLAUDE.md` gains the ordering rule (strip → filter → enqueue) and the keywords-on-stripped-text divergence. No new deps, no env vars, no schema change.

**Files to Create / Modify:** `src-tauri/src/native/integrations/slack/inbound.rs`, `src-tauri/src/native/integrations/slack/CLAUDE.md`, `src/views/integrations/catalog.ts`, `docs/user-guide.md`.

**Risks:**
- *A Slack rule with a non-empty prefix stored today stops firing on upgrade* → that is the fix, but it is a behaviour change users will feel; call it out in the release notes and in the docs section.
- *Resolving `bot_user_id` earlier adds an `auth.test` call to the accept path* → it is a `OnceCell` fetched at most once per connection, and `run` already awaits it; the cost moves rather than doubles.
- *Reusing `match_rule` drags Go parity into the Slack path* → it is reused read-only and the parity vectors stay authoritative for Telegram; any Slack-specific choice lives at the call site, not inside the matcher.

**Complexity:** Medium (one worker file plus tests, one UI string change, docs)

---
**Refined by:** AI Issue Refiner · **Date:** 2026-09-09 · **Status:** Ready

vibexp-agent:
Closes #582

## What

Slack's inbound path now honours a trigger rule's `filter_prefix` and `filter_keywords` against the mention text with the bot's own `<@Uxxx>` removed. The two inputs the form has always shown for every provider now mean what they say on Slack.

## Why

The form renders Prefix and Keywords for every provider, and the Slack path read neither — both were loaded off the row by `dispatcher::load_rules` and dropped. A filter that is displayed, stored and then ignored is worse than no filter: it looks like a safety measure, and it fails in the direction of running *more* than was asked for.

## How

- **`slack/inbound.rs`** — `accept` resolves the bot user id, strips the mention and gates the **already selected** rule through `trigger::match_rule::match_rule`, before the queue. `Job` carries the resolved `prompt`, so the text that was filtered is exactly the text that runs and `turn` no longer re-derives one.
  - The filters are a gate on the rule `select_rule_for_channel` picked (#565), never a reason to fall through to a broader one.
  - **Keywords are matched against the stripped text**, a deliberate divergence from Telegram: `match_rule`'s header pins keywords to the whole message because Go did, and on Slack the whole message always begins with a bot id no user typed. Stated at the call site.
  - `match_rule`'s own `chat_ids` clause is a no-op here (selection has already returned a rule naming this channel, or one naming none) and is left in place rather than routed around — asserted by a test.
- **`native/trigger/match_rule.rs`** — untouched. `parity/trigger_match_vectors.json` stays Telegram's authority.
- **`src/views/integrations/catalog.ts` / `IntegrationsView.tsx`** — `TriggerTargets` gains a required `filterHelp`, rendered under the Prefix/Keywords row (the pattern `targets.help` already sets for the chat-ids row). Required rather than optional so a provider cannot gain the inputs without someone writing down what they match against there.
- **`docs/internal/native-integrations-slack.md`** — the strip → filter → enqueue ordering, the keywords divergence, and the `auth.test`/`ERROR_REPLY` split.

## Deviations from the issue's HOW, and why

1. **`auth.test` failing is still `ERROR_REPLY`, not a silent drop.** The HOW folded the unavailable-bot-id case into the same drop as an empty remainder. That would change an error-path behaviour the issue did not ask about and that `inbound.rs`' module header specifies. `Job::prompt` is an `Option` instead: `None` is that one failure, and `turn` answers it exactly where it did before, after the thread is known to be Agento's. `accept` itself posts nothing, so moving the *fetch* earlier is safe.
2. **The ordering note went to `docs/internal/native-integrations-slack.md`, not `src-tauri/src/native/integrations/slack/CLAUDE.md`.** That file cannot exist: since #580/#583 the repository may contain exactly one `CLAUDE.md`, at the root, and `scripts/check-claude-md-size.sh` fails CI otherwise.
3. **No `docs/user-guide.md` change.** The HOW asks for the *Slack Socket Mode → Who can run an agent this way* interim mitigation to be corrected. That section is #581's and **#581 is still open**, so nothing on `main` says prefix and keywords do not apply on Slack — there is nothing to correct. **#581 should have that sentence updated before it merges**; flagged rather than edited, since it is another PR's text.

## Acceptance criteria

- [x] `filter_prefix = "/ask"`: `@bot /ask what is the status` runs with prompt `what is the status`; `@bot what is the status` starts nothing and posts nothing
- [x] Prefix matching sees the stripped text — `@bot /ask …`, `/ask @bot …` and `/ask … @bot` are one message
- [x] `filter_keywords = ["deploy"]`: `Deploy` (any case) runs, a mention without it does not
- [x] Empty prefix and empty keywords behave exactly as today
- [x] A mention that matches the channel but fails the filters does not fall through to a broader rule; the drop is logged at `debug` with no Slack-derived text
- [x] A bare `@bot /ask` starts nothing
- [x] Telegram unchanged: `parity/trigger_match_vectors.json` and `native::trigger::match_rule` pass untouched
- [x] Unit tests cover prefix hit/miss, keyword hit/miss, both empty, prefix-only text and a mention in the middle

## Tests

Seven new tests in `slack/inbound/tests.rs` — five unit tests over `filtered_prompt` (the call site itself, not a re-spelling of it) and two end-to-end tests: one proving no fall-through to the workspace default, one proving the prefix-stripped remainder is what reaches the chat and the title *and* that a follow-up in Agento's own thread is gated the same way.

**Revert-proof, measured:** stubbing `filtered_prompt`'s `match_rule` call out fails all seven (`7 failed; 12 passed`).

## A decision the issue did not pose

The filter is decided in `accept`, before the thread map is read, so **a reply inside a thread Agento started is gated exactly like the mention that opened it** — a rule with `Prefix: /ask` wants `/ask` on every follow-up too.

Kept deliberately: #582's WHAT says "apply prefix and keyword filtering to Slack app-mention events", unqualified, and it is the direction that runs *less*. Stated in `inbound.rs`'s module header, at `filtered_prompt`, in the area doc and in Slack's `filterHelp`, and pinned by `a_prefixed_rule_gates_the_follow_up_in_its_own_thread_too`. Exempting mapped threads is the other defensible answer and would be a behaviour change, not a doc change.

## Review

Four automated review rounds, every finding applied: 3 → 2 → 1 → **APPROVE with no findings**. Round 1 surfaced the follow-up-gating decision above; rounds 2–3 were documentation accuracy (a contradicting caveat on the still-open #581, a stale consequence count, two doc sentences this change had falsified).

## Local gates

`cargo fmt --check`, `cargo clippy --all-targets -- -D warnings`, `cargo clippy --lib -- -D warnings` (the Windows arms), `cargo test` (1601 lib + every integration suite, 0 failed), `npm run build`, `scripts/check-claude-md-size.sh`, `web/scripts/sync-docs.mjs` — all green.

### DIFF
--- a/docs/internal/native-integrations-slack.md
+++ b/docs/internal/native-integrations-slack.md
@@ -284,6 +284,16 @@ declines to start one.
     would fire on every mention. `keywords_are_case_insensitive_and_never_see_the_bots_own_id`
     is that case. The prefix is matched on the stripped text for the same reason,
     which is also what makes `@bot /ask …` and `/ask @bot …` one message.
+  - **The filters gate a reply inside one of Agento's own threads too.** `accept`
+    has not read `inbound_threads` at that point and deliberately does not — the
+    mapping lookup belongs inside the per-thread worker, per the bullet above —
+    so a rule carrying a prefix asks for that prefix on every message in the
+    channel, follow-ups included. That is the reading of #582's *apply the
+    filters to app-mention events*, and it is the direction that runs less: the
+    alternative exempts anyone who can reply in a thread from the filter the
+    channel was given. Slack's `filterHelp` in `views/integrations/catalog.ts`
+    says so to the user, and
+    `a_prefixed_rule_gates_the_follow_up_in_its_own_thread_too` is the guard.
   - **`match_rule`'s own `chat_ids` clause is a no-op here, and is left in
     place.** Selection has already returned either a rule naming this channel or a
     rule naming none, and the matcher reads the second as "everything", so the two
@@ -310,9 +320,10 @@ declines to start one.
   rule and the filters all need; a failure to get it is a failed turn and answers
   `ERROR_REPLY`, because a Slack that will not answer `auth.test` will not accept
   `chat.postMessage` either. Moving the call into `accept` moves its cost rather
-  than doubling it — the `OnceCell` is per handler, and `turn` already awaited it. `conversations.info` (chat title) and
-  `chat.getPermalink` are best-effort on the start path only — the channel id and
-  an empty permalink are the fallbacks, and neither refuses the run.
+  than doubling it — the `OnceCell` is per handler, and `turn` already awaited it.
+  `conversations.info` (chat title) and `chat.getPermalink` are best-effort on the
+  start path only — the channel id and an empty permalink are the fallbacks, and
+  neither refuses the run.
 - **`mrkdwn.rs` escapes `&`, `<` and `>` before it converts anything**, over the
   whole answer including its fenced blocks. Slack renders those entities back as
   themselves everywhere, so escaping costs the answer nothing — and not escaping
--- a/docs/internal/native-integrations-slack.md
+++ b/docs/internal/native-integrations-slack.md
@@ -232,8 +232,8 @@ declines to start one.
   otherwise find no `inbound_threads` row — the row is written by the run it is
   queued behind — and be dropped as a mention in a thread Agento did not start.
   This is the one ordering constraint the FIFO exists for, and it is why
-  classification is split across `accept` (the rule) and `turn` (the mapping, the
-  bot-id strip, start-or-resume). The order the queue preserves is **enqueue**
+  classification is split across `accept` (the rule, the strip, the filters) and
+  `turn` (the mapping, start-or-resume). The order the queue preserves is **enqueue**
   order, not arrival order: `socket.rs::dispatch` spawns a task per envelope and
   each awaits its dedup claim before a handler runs, so two mentions posted a
   millisecond apart can reach the queue either way round and nothing downstream
@@ -250,13 +250,46 @@ declines to start one.
   line saying a mention was seen and a reply appearing minutes later are two
   different things to anyone reading a log — but that wait now costs a parked
   task and no permit.
-- **`auth.test` is resolved after the mapping decision, not before it.** It is
-  the one Slack failure the handler cannot work around, and its answer is
-  `ERROR_REPLY` — so resolving it first would post that sentence into a thread
-  Agento never started, which is exactly what the ignore rule exists to prevent.
-  Reading the thread map likewise distinguishes *no row* from *could not read*:
-  `busy_timeout` is five seconds, and a database busy behind the session scanner
-  would otherwise make a resume look like a stranger's thread and answer nothing.
+- **`auth.test` is resolved in `accept`; its `ERROR_REPLY` is still posted in
+  `turn`.** The two used to be one step, and #582 separated them. The bot user id
+  is what the `<@Uxxx>` strip needs and the strip is what the filters read, so the
+  id has to be in hand *before* the filter decision, which is before the queue.
+  The sentence it fails with does not move: `accept` posts nothing at all, and the
+  `ERROR_REPLY` for an unresolvable id waits in `turn` until the mapping decision
+  has established the thread as Agento's — otherwise it lands in a thread Agento
+  never started, which is exactly what the ignore rule exists to prevent. A `Job`
+  therefore carries `prompt: Option<String>`, and `None` is that one failure
+  rather than a drop. Reading the thread map likewise distinguishes *no row* from
+  *could not read*: `busy_timeout` is five seconds, and a database busy behind the
+  session scanner would otherwise make a resume look like a stranger's thread and
+  answer nothing.
+- **Strip, then filter, then enqueue — and the filters read the stripped text
+  (#582).** The trigger-rule form has always shown Prefix and Keywords for every
+  provider, and until #582 the Slack path read neither: `filter_prefix` and
+  `filter_keywords` were loaded off the row by `dispatcher::load_rules` and
+  dropped. A control that is displayed, stored and ignored fails in the direction
+  of running *more* than was asked for, so `accept` now gates the **already
+  selected** rule through `trigger::match_rule::match_rule` — the same function
+  Telegram calls, unmodified, with `parity/trigger_match_vectors.json` still its
+  authority. Three consequences worth stating:
+  - **The filters are a gate on the selected rule, never a reason to look for
+    another one.** `select_rule_for_channel`'s most-specific-wins (#565) picks the
+    rule; a rule that then refuses the message means silence, not a fall-through
+    to the workspace default. `a_mention_the_channels_rule_refuses_never_falls_through_to_the_default`
+    is the guard.
+  - **Keywords are matched against the stripped text, which is a deliberate
+    divergence from Telegram.** `match_rule`'s header pins keywords to the *whole*
+    message because Go did; on Slack the whole message always begins with a
+    `<@Uxxx>` no user typed, so a keyword equal to a fragment of the bot's own id
+    would fire on every mention. `keywords_are_case_insensitive_and_never_see_the_bots_own_id`
+    is that case. The prefix is matched on the stripped text for the same reason,
+    which is also what makes `@bot /ask …` and `/ask @bot …` one message.
+  - **`match_rule`'s own `chat_ids` clause is a no-op here, and is left in
+    place.** Selection has already returned either a rule naming this channel or a
+    rule naming none, and the matcher reads the second as "everything", so the two
+    agree. `the_matchers_channel_check_agrees_with_the_rule_selection` asserts that
+    rather than the code routing around it — one matcher, called the same way from
+    both transports.
 - **Queue teardown is under the map lock.** A worker that found its channel
   empty, released nothing and then removed its entry would lose a job queued in
   between, so the drain re-checks the channel while holding the lock that hands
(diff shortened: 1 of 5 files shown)


vibexp-agent:
**Not auto-merged — handing off.**

CI is green on `9a07cb1`: all 9 checks pass (`Typecheck, lint and build`, `build`, `Windows unit tests`, `Site check`, and every CodeQL analysis including `Analyze (rust)`).

The automated review ran its full budget of **3 rounds** and never returned APPROVE, which is a hard stop for the unattended pipeline that opened this PR — so it stops here rather than merging. It is not a red flag on the code: each round's findings were smaller than the last (3 `now` → 2 `now` → 1 🟡), all were applied, and round 3's single finding was a stale doc comment, fixed in `9a07cb1` and unreviewed only because the budget ran out.

**The one thing worth a human's eye** is a product decision the issue did not pose. The filter decision is made in `accept`, before the thread map is read, so **a reply inside a thread Agento started is gated exactly like the mention that opened it** — a rule with `Prefix: /ask` wants `/ask` on every follow-up too. I kept that reading (#582's WHAT says "apply prefix and keyword filtering to Slack app-mention events", unqualified, and it is the direction that runs *less*), stated it in the module header, at `filtered_prompt`, in the area doc and in Slack's `filterHelp`, and pinned it with `a_prefixed_rule_gates_the_follow_up_in_its_own_thread_too`. Exempting mapped threads is the other defensible answer and would be a behaviour change, not a doc change.

Also outstanding, and not this branch's to fix: PR #581's user-guide paragraph saying Prefix and Keywords do not apply on Slack becomes false when this merges — recorded at #581#issuecomment-5593909313.

Issue #595, opened 2026-09-18
Kill in-flight agent runs on app exit


vibexp-agent:
Part of epic #593. Blocked by #594.

## Scope

Add an app-exit hook that terminates every in-flight run's process group
before Agento quits, so closing/quitting the app actually stops the work
instead of leaving it running as an orphan.

Today `src-tauri/src/lib.rs` has exactly one `.on_window_event` (macOS-only,
for traffic-light repositioning) and the builder ends in a plain
`.run(tauri::generate_context!())`. `RunEvent::ExitRequested` is never
intercepted, so the graceful shutdown path already implemented for normal
turn-end/error paths (`terminate()` → grace period → `kill()` in
`claude/process.rs`) never runs on quit, and Rust destructors (`kill_on_drop`,
`Stream::drop`) do not run on the abrupt termination a GUI quit performs.

## Changes

- Switch `lib.rs`'s builder to the `.run(|app, event| ...)` form and handle
  `RunEvent::ExitRequested`: iterate in-flight runs (scheduled + manual +
  chat), send `SIGTERM` to each stored process group (from #594), wait
  up to the existing grace period, then `SIGKILL` any still alive.
- Ensure this runs to completion (or a bounded timeout) before the process
  actually exits — `ExitRequested` allows deferring exit via
  `api.prevent_exit()` for exactly this kind of cleanup; use it, but bound the
  wait so a hung child can't wedge app shutdown indefinitely.
- Mark the affected `job_history` rows `failed` with a reason such as
  `"terminated: app quit"` before exit, if the write can complete in the
  bounded window — if not, leave them for the startup reaper (#REAPER_ISSUE)
  to pick up next launch.

## Acceptance criteria

- [ ] Quitting Agento while a scheduled/manual run is in flight terminates the
      child process tree within the bounded grace window (verified with `ps`
      showing no surviving `claude` process after quit).
- [ ] The corresponding `job_history` row ends up `failed` (either written
      during the bounded exit window, or picked up by the reaper on next
      launch) rather than staying `running` forever.
- [ ] App shutdown itself is still bounded — a child that refuses to die does
      not prevent Agento from eventually exiting.
- [ ] No regression to normal chat-turn or scheduled-run shutdown behavior
      (existing graceful shutdown on ordinary turn-end is untouched).

## Complexity

Medium.

## Blockers

#594 (needs process-group ownership + a stored PID to signal).

---
## WHAT
[...]

Issue #594, opened 2026-09-18
Process-group ownership + PID tracking for spawned agent runs


vibexp-agent:
Part of epic #593.

## Scope

Give every spawned Claude CLI run its own OS process group, and persist the
child's PID (and the process-group id, if distinct) on its `job_history` row.
This is the prerequisite for anything outside the run's own future being able
to identify and terminate it — the app-exit hook (#EXIT_ISSUE) and the startup
reaper (#REAPER_ISSUE) both depend on it.

Today `claude/process.rs`'s `spawn_and_stream` builds the `tokio::process::Command`
with `.kill_on_drop(true)` only — explicitly documented as "a backstop only" —
and no `setsid`/`process_group(0)` (Unix) or `CREATE_NEW_PROCESS_GROUP`
(Windows) is set anywhere in the spawn path. `kill_on_drop` only fires on an
ordinary Rust `Drop`, which does not happen on the abrupt process termination
a GUI quit performs, so it does not help the sibling issues.

## Changes

- Set process-group/session ownership at spawn time (Unix: `setsid` or
  `process_group(0)`; Windows: `CREATE_NEW_PROCESS_GROUP`) for both the
  scheduled/manual `query` path and the chat `Session` path, so a signal to
  the group reaches the whole subprocess tree the CLI itself may spawn.
- Add a column (or equivalent) to `job_history` to store the spawned PID
  (and start time, to guard against PID reuse — see the epic's risk note).
  Migration follows the existing pattern in `native/db.rs`/`migrate.rs`.
- Write the PID as soon as the child is spawned, before any output is read,
  so a crash immediately after spawn still leaves a row that names the
  process to check.

## Acceptance criteria

- [ ] A spawned run's child process is the leader of its own process group
      (verified with `ps -o pid,pgid` showing pid == pgid for the child).
- [ ] `job_history` persists the PID (and start time) for every run, written
      before the run's output is consumed.
- [ ] Sending `SIGTERM`/`SIGKILL` to the stored PID terminates the entire
      subprocess tree the CLI spawned, not just the immediate child.
- [ ] Existing `tests/scheduled_run.rs` / `tests/chat_turn.rs` suites still
      pass unmodified in behavior (this issue only adds ownership + a
      recorded PID, not new termination logic).
- [ ] Windows equivalent verified or explicitly deferred with a follow-up note
      if `CREATE_NEW_PROCESS_GROUP` behavior differs enough to need its own
      issue.

## Complexity

Low/medium.

[...]

Issue #593, opened 2026-09-18
Epic: Reliable termination and orphan recovery for scheduled/manual agent runs


vibexp-agent:
### Problem

A scheduled or manually-triggered task's Claude CLI subprocess is not tied to
Agento's own process lifetime, and its `job_history` row is never finished by
anything other than the process that started the run. Two consequences,
observed directly:

- **Closing or restarting Agento does not stop the run.** `claude/process.rs`
  spawns the child with no process-group ownership (no `setsid`/
  `process_group(0)`, no Windows `CREATE_NEW_PROCESS_GROUP`), and `lib.rs` has
  no `ExitRequested`/`CloseRequested` handler. The graceful shutdown path
  (SIGTERM → grace → SIGKILL in `terminate()`/`kill()`) only runs on ordinary
  turn-end/error paths, never on app quit. When the window closes, the parent
  process terminates without running its destructors, and the child `claude`
  process is reparented to init and keeps running to completion, untracked.
- **The `job_history` row is orphaned along with it.** Nothing else ever
  writes `finished`/`failed` to that row, so it displays `RUNNING` forever in
  the UI — this is what produced a job showing 16+ hours of runtime with no
  token count, in a case that prompted this epic.

### Solution

Four independent-where-possible pieces:

1. Give every spawned run process-group ownership and record its PID on the
   `job_history` row, so a run can be identified and killed from outside its
   own future.
2. Add an app-exit hook that terminates all in-flight run process groups
   before the app quits.
3. Add a startup reaper that reconciles `job_history` rows stuck `running`:
   kill any surviving orphaned process, mark the row `failed` with a clear
   reason, and free the scheduler's semaphore slot.
4. Add a configurable per-task wall-clock timeout, enforced by the executor
   during the run itself, independent of app lifecycle.

### Design decisions

1. Process-group ownership is the prerequisite for both the exit hook and the
   reaper being able to reliably kill a run — it lands first.
2. The exit hook and the startup reaper are complementary, not redundant: the
   exit hook handles the clean-quit path, the reaper is the recovery net for
   anything that bypassed it (SIGKILL, crash, power loss).
3. The timeout is fully independent of the other three — it bounds a run that
   is alive and owned by a live Agento process but simply stuck/looping.

### Security / risk notes

- Killing a process group must not affect unrelated processes — ownership is
  scoped per spawned run via its own process group, never a signal to `-1` or
  the whole session.
- The reaper must distinguish "PID alive but reused by an unrelated process"
[...]

PR #612, opened 2026-09-18
feat: reap a previous session's running job_history rows at startup (#596)

### LINKED ISSUE #596: Startup reaper for orphaned/stuck job_history rows
Part of epic #593. Blocked by #594.

## Scope

Add a startup reaper that reconciles any `job_history` row left `running`
from a previous session — the recovery net for anything that bypassed the
exit hook (#595): a `SIGKILL`, a crash, a power loss, or an update
that replaced the binary mid-run.

This is what turns a stuck `RUNNING` row (observed at 16+ hours with no
token count, with no live watcher able to ever finish it) into a bounded,
explained `failed` row instead of a permanent UI artifact — and it's also
where an actually-still-running orphaned subprocess (left behind because
Agento didn't cleanly exit) finally gets killed.

## Changes

- On startup, before (or as part of) `Scheduler::start`, query
  `job_history` for rows still `running`.
- For each: check whether its stored PID (from #594) is alive **and**
  matches the recorded start time/command signature (to guard against PID
  reuse by an unrelated process — see the epic's risk note).
  - Alive and matching → it's a surviving orphan: terminate its process
    group (reuse the same terminate/grace/kill logic as #595), then
    mark the row `failed` with reason `"orphaned: recovered on startup"`.
  - Not alive → the process already exited (or the row predates PID
    tracking) — mark the row `failed` with reason `"orphaned: app did not
    exit cleanly"`.
- Free the scheduler's semaphore slot / in-flight tracking for the affected
  task id, per the existing `Scheduler::in_flight` refcount (documented in
  `docs/internal/native-schedule.md`), so a reaped task is immediately
  eligible to run again on its next scheduled fire.
- This is a sweep, not a fix to `reconcile`'s existing fingerprint-based timer
  sync — it runs once at startup (or on the same 60s cadence as `reconcile`)
  and only touches rows, not installed timers.

## Acceptance criteria

- [ ] A `job_history` row left `running` by a killed Agento process, with its
      `claude` child still alive, is killed and marked `failed` with a
      distinguishing reason on the next launch.
- [ ] A `running` row whose process already exited (or has no PID recorded,
      e.g. written before this feature) is marked `failed` without attempting
      to signal an unrelated/reused PID.
- [ ] A reaped task's scheduler semaphore slot is released, and the task
      fires normally on its next due time (not permanently blocked by a
      stale in-flight count).
- [ ] The reaper cannot itself become a startup blocker — matches the "sweep
      armed before the first read can fail" principle already established
      for `reconcile` (see `docs/internal/native-schedule.md`): a transient
      DB error here must not leave the app in a state with no recovery until
      restart.
- [ ] Covered by an integration test analogous to `tests/scheduled_run.rs`
      that seeds a stale `running` row and asserts the post-startup state.

## Complexity

Medium.

## Blockers

#594 (needs a stored PID to check liveness against).

---
## WHAT
**Summary:** Add a startup reaper that reconciles `job_history` rows left `running` from a previous session, killing any surviving orphaned process and marking the row `failed` with a clear reason.

**Scope:**
- On startup (before or alongside `Scheduler::start`), query `job_history` for rows still `running`.
- For each: check the stored PID + start time (from #594) for liveness *and* identity match (guards against PID reuse).
  - Alive & matching → terminate its process group (reuse #595's terminate/grace/kill sequence), mark `failed`, reason `"orphaned: recovered on startup"`.
  - Not alive (or no PID recorded — pre-existing rows) → mark `failed`, reason `"orphaned: app did not exit cleanly"`.
- Release the task's slot in `Scheduler::in_flight`'s refcount so a reaped task is immediately eligible to fire again.

**Affected Components:**
- `src-tauri/src/native/schedule/runtime.rs` — houses `Scheduler::in_flight` (the refcounted task-id map added for the manual-run `409`, per `docs/internal/native-schedule.md`) and `Scheduler::reconcile` (the existing 60s timer-sync sweep, fingerprinted on `schedule_type`+`schedule_config`). This reaper is a **sibling sweep over rows**, not a change to `reconcile`'s timer-fingerprint logic.
- `src-tauri/src/native/schedule/executor.rs` — reuse `record_failed_run`'s shape (already used for the "agent tools unavailable" case) for writing the reaped row's terminal state.
- `src-tauri/src/claude/process.rs` — reuse the terminate/grace/kill primitives from #595.

**Acceptance Criteria:**
- [ ] A `job_history` row left `running` by a killed Agento process, whose `claude` child is still alive, is killed and marked `failed` with a distinguishing reason on next launch.
- [ ] A `running` row whose process already exited (or predates PID tracking) is marked `failed` without signaling an unrelated/reused PID.
- [ ] A reaped task's `Scheduler::in_flight` slot is released, and the task fires normally on its next due time.
- [ ] The reaper cannot itself become a startup blocker — mirrors the existing principle "the sweep is armed before the first read can fail" (`docs/internal/native-schedule.md`, re: `reconcile`'s SQLITE_BUSY-at-boot fix): a transient DB error here must not leave the app with no recovery path until restart.
- [ ] Covered by an integration test analogous to `tests/scheduled_run.rs` that seeds a stale `running` row (with and without a live matching PID) and asserts the post-startup state.

**Out of Scope:** UI surfacing of "this run was orphaned" beyond the `job_history.error_message`/status fields already used for other terminal states (e.g. the `success`-with-`error_message` convention documented in `docs/internal/native-schedule.md`) — a richer UI treatment can be a follow-up if wanted.

## WHY
**Business Value:** Turns a permanently stuck `RUNNING` row (observed at 16+ hours, no token count, in the report that prompted this epic) into a bounded, explained `failed` row — and is the only mechanism that recovers from anything #595's clean-exit hook didn't catch (`SIGKILL`, crash, power loss, binary replaced mid-run by an update).

**Problem Statement:** Today nothing ever finishes a `job_history` row once the process that started it is gone — there is no watchdog, so the UI shows indefinite `RUNNING` with no way back short of manual DB surgery.

**Expected Impact:** Every orphaned run is resolved within one app launch instead of persisting forever; the scheduler's in-flight accounting stays accurate so a reaped task isn't wrongly blocked from its next scheduled fire.

**Context:**
- Parent: #593
- Blocked by: #594 (needs a stored PID to check liveness against)
- Related: #595 (the clean-exit path this reaper backstops)

## HOW
**Technical Approach:** A startup sweep, run once during boot (candidate hook point: the same place `Scheduler::start`'s existing sweep is armed), that lists `running` job_history rows, checks each PID with `sysinfo`/`kill(pid, 0)`-style liveness plus the stored start-time signature, and resolves each to a terminal state.

**Implementation Details:**
- New function alongside `Scheduler::reconcile`, e.g. `Scheduler::reap_stale_runs`, called once at the same point `reconcile`'s first pass is armed (so it inherits the "armed before the first read can fail" guarantee already established there).
- Liveness+identity check: read `/proc/<pid>` start time on Linux (or equivalent) and compare against the stored start time within a tolerance, rather than trusting PID alone.
- Write path: same `job_history` update `finish`/`record_failed_run` use, so the write-then-effect discipline (`root CLAUDE.md`: "a write does more than write a row") is preserved — releasing the `in_flight` slot is the "effect" here.

**Testing Strategy:** `tests/scheduled_run.rs`-style test seeding a `job_history` row with `status = "running"` and a PID pointing at (a) a real still-running fake process, (b) an already-exited PID, (c) no PID at all — assert each resolves to the correct `failed` reason and that `in_flight` no longer holds the task id afterward.

**Documentation / Configuration:** Add this reaper to `docs/internal/native-schedule.md`'s description of `Scheduler::start`, next to the existing `reconcile` documentation, since both are startup-time sweeps with related but distinct responsibilities (timers vs. rows).

**Files to Create / Modify:** `src-tauri/src/native/schedule/runtime.rs`, `src-tauri/src/native/schedule/executor.rs`, `src-tauri/tests/scheduled_run.rs`.

**Risks:** Killing a process whose PID was reused by an unrelated program → mitigated by the start-time cross-check (epic-level risk note); a reaper that itself panics/errors blocking startup → mitigated by the same "armed before first read can fail" discipline `reconcile` already uses.

**Complexity:** Medium (new sweep function, PID+start-time liveness check, in-flight refcount release, new test coverage).

---
**Refined by:** AI Issue Refiner · **Date:** 2026-09-18 · **Status:** Ready

vibexp-agent:
Closes #596 (part of epic #593).

## What
On startup, every `job_history` row a **previous** session left `running` is failed with an explicit reason, and a CLI from that session that is provably still running is stopped first.

## Why
Nothing else ever finishes such a row — the process that would have is gone (SIGKILL, crash, power loss, an update replacing the binary, or a row the #595 exit hook's 2 s did not see written). It showed as `RUNNING` forever.

## How
- **`claude/process.rs`**: `stop_orphan(pid, started_at, grace)` — signals only when the pid is alive **and** its start time (`ps -o etime=`, one probe identical on Linux and macOS) is within 5 s of the recorded `pid_started_at`; then SIGTERM to the group, SIGKILL after `SIGKILL_GRACE`, exactly as the exit hook does. `group_alive`'s unix body moved to a free `group_exists` so both share it. `is_live_run(pid)` exposes the #595 live set. On Windows nothing is signalled (no start-time probe); rows are only marked.
- **`native/tasks.rs`**: `list_running_jobs` and `reap_job_history` — the update is guarded by `status = 'running'` so a run finishing concurrently wins.
- **`native/schedule/runtime.rs`**: `Scheduler::reap_stale_runs`. Reasons: `orphaned: recovered on startup` (stopped) and `orphaned: app did not exit cleanly` (exited, no pid, or pid reused). Only rows started before the scheduler was built (`booted_at`) and whose pid is not an un-reaped child of this process are touched. `start` spawns it beside the reconcile sweep, on the blocking pool, off the boot path, retried every 60 s until one pass succeeds ("armed before the first read can fail").
- Docs: `docs/internal/native-schedule.md`.

**Deviation from the HOW — no `in_flight` slot is released.** `Scheduler::in_flight` and the semaphore are per-process memory and start empty, so a previous session's row never held a slot; decrementing would corrupt the count of a run this session owns. The acceptance criterion (a reaped task is not blocked and fires normally) is asserted by a test instead. `executor.rs` needed no change: the reaper's write is a dedicated guarded UPDATE rather than `finish_job_history`, whose unguarded update could clobber a concurrent finish.

## Tests
- `scheduled_run.rs::the_startup_reaper_stops_a_surviving_orphan_and_fails_every_stale_row` — a live group-leading orphan with a grandchild (stopped, group gone), an exited pid, a pre-#594 row, a live pid with the wrong start time (**not signalled**), and a row of this session (**untouched**); second pass is a no-op. Verified as a regression guard: disabling the start-time check or the `booted_at` guard turns it red.
- `a_task_whose_stale_run_was_reaped_fires_normally_afterwards` — no 409, and the next fire records `success`.
- `process::tests::an_orphan_is_stopped_only_when_its_start_time_matches`, `etime_parses_every_width_ps_prints`.

Local: `cargo fmt --check`, `cargo clippy --all-targets -D warnings`, full `cargo test` green.


vibexp-agent:
Deferred to #613: stopping a surviving orphan on Windows (needs a start-time probe; this PR only marks the row there).

Issue #604, opened 2026-09-18
Credentials Checker: /api endpoints — list findings, whitelist/false-positive, scan status


vibexp-agent:
Part of #597.

Add `native/security_scan/mod.rs` and register it as an `Endpoint` in `native/mod.rs`'s `ENDPOINTS` (one appended line, per that file's convention). Routes:

- `GET /api/security-scan/findings` — list findings (session, rule, confidence, masked snippet, status, detected_at), filterable by status
- `POST /api/security-scan/findings/{id}/whitelist` — mark a finding whitelisted (by value hash)
- `POST /api/security-scan/findings/{id}/false-positive`
- `GET`/`POST`/`DELETE /api/security-scan/whitelist` — manage whitelist entries directly (rule-level or value-level), for the Settings UI
- `GET /api/security-scan/status` — worker enabled/running, last sweep time, counts

State-changing routes need `Content-Type: application/json` per the guard convention; every route needs the bearer token like the rest of `/api`.

Design doc: https://vibexp.shaharialab.com/artifacts/203ca41f-c0ae-4a6c-92c8-62fd3c61b71b/design-credentials-checker (§5.1, §6)

Blocked by:
- Store (this is the read/write surface over it)

---
## WHAT
**Summary:** A new `native/security_scan/api.rs` module exposing six `/api/security-scan/*` routes over the Credentials Checker store — list findings, mark whitelisted/false-positive, manage whitelist entries, and report scan status — registered in `native/mod.rs`'s `ENDPOINTS`.

**Scope:**
- `GET /api/security-scan/findings` — list findings, filterable by `status` (`open` | `whitelisted` | `false_positive`), newest-first
- `POST /api/security-scan/findings/{id}/whitelist` — look up the finding's `match_hash`, then whitelist by that hash (retroactively suppresses every other finding sharing it, via the store's existing behavior)
- `POST /api/security-scan/findings/{id}/false-positive` — mark one finding `false_positive` (a verdict the store never overwrites, including on rescan)
- `GET /api/security-scan/whitelist` — list whitelist entries
- `POST /api/security-scan/whitelist` — add an entry, body carries exactly one of `rule_id` or `match_hash` (mutually exclusive) plus an optional `reason`
- `DELETE /api/security-scan/whitelist/{id}` — remove an entry (path form inferred from the issue's ambiguous `DELETE /api/security-scan/whitelist`; matches the existing `DELETE /api/security/tokens/{id}` shape)
[...]

Issue #603, opened 2026-09-18
Credentials Checker: background scan worker (boot sweep + incremental)


vibexp-agent:
Part of #597.

Add `native/security_scan/worker.rs`, following the shape of `native/insights/worker.rs`: a dedicated `std::thread::spawn` loop (not tokio — the same reasoning as insights, everything here is blocking rusqlite + transcript reads), a boot sweep over the full corpus on first enable, then an incremental pass keyed off whatever `scan.rs` already uses to notice new/changed sessions (no second filesystem-watch mechanism).

**Gated by the settings toggle**: the worker does not start at all while `credentials_checker_enabled` is false, and flipping the toggle stops/starts the worker rather than having it poll a flag every pass.

What gets scanned per session: user/assistant message text plus both `input` and the textual parts of `output` on every tool call (`Bash` stdout/stderr, `Read`/`Write` file contents, `WebFetch` bodies, MCP tool results), decoded the same way `native/insights/transcript.rs` already does. Binary/non-text tool outputs are skipped.

Design doc: https://vibexp.shaharialab.com/artifacts/203ca41f-c0ae-4a6c-92c8-62fd3c61b71b/design-credentials-checker (§4.3, §5.1, §6)

Blocked by:
- Store (needs the persistence layer to write findings to)
- Settings field (needs the toggle to gate on)

---
## WHAT
**Summary:** Add `native/security_scan/worker.rs` — a `std::thread`-based background worker that keeps `credential_scan_state`/`credential_findings` in sync with the session corpus: a boot sweep, an incremental hookup off `scan.rs`'s existing change-notification, and toggle-driven start/stop wired to `credentials_checker_enabled`.

**Scope:**
- New module `native/security_scan/worker.rs`, declared in `security_scan/mod.rs`
- A **stoppable** worker lifecycle (start/stop/reload) driven by the `credentials_checker_enabled` setting — unlike `insights::worker`, which cannot be stopped once started
- Boot sweep over the corpus using `store::needs_scanning`
- Incremental path: a second `security_scan::worker::enqueue(...)` call alongside the existing `super::insights::worker::enqueue(outcome.notifications...)` in `scan.rs`'s `ensure_scan_inner` (`src-tauri/src/native/scan.rs:558`), fed from the same `outcome.notifications`
[...]

Issue #599, opened 2026-09-18
Credentials Checker: vendor high-confidence credential detection rules


vibexp-agent:
Part of #597.

Add the `regex` crate to `src-tauri/Cargo.toml` (not currently a dependency) and a new `native/security_scan/rules.rs` holding a vendored, high-confidence-only rule table, ported from gitleaks' ruleset (Apache-2.0) rather than written from scratch — see design doc §3–§4.1 for why.

**v1 rules (high confidence only, no entropy heuristics):**
- AWS access-key-id (`AKIA[0-9A-Z]{16}`) + paired secret key when adjacent
- GCP service-account JSON `"private_key"` blocks
- Azure connection strings
- GitHub tokens (`ghp_`, `gho_`, `github_pat_`)
- Slack tokens (`xox[baprs]-`)
- Stripe (`sk_live_`, `rk_live_`)
- OpenAI (`sk-[A-Za-z0-9]{20,}`), Anthropic (`sk-ant-`), npm (`npm_`)
- Postgres/MySQL URLs with embedded `user:password@host`
- PEM private key blocks
- JWT-shaped triple — the one "medium" confidence rule, everything else ships "high"

Also add `native/security_scan/scan.rs`: a pure function `transcript -> Vec<Finding>`, unit-tested against fixtures for each rule (true positives) and known false-positive shapes (UUIDs, hashes, minified code) to confirm the "no plausible false positive" bar from the design doc.

Design doc: https://vibexp.shaharialab.com/artifacts/203ca41f-c0ae-4a6c-92c8-62fd3c61b71b/design-credentials-checker (§3, §4.1–§4.2)

No dependencies — this is the foundational rule engine everything else builds on.

---
## WHAT
**Summary:** Vendor a high-confidence, format-specific credential-detection rule set and a pure `transcript -> Vec<Finding>` scan function, as the foundation everything else in the Credentials Checker epic builds on.

**Scope:**
- Add the `regex` crate to `src-tauri/Cargo.toml` (currently absent from the tree)
- `native/security_scan/rules.rs`: a static rule table — id, compiled regex, provider label, fixed confidence tier (`high` | `medium`)
- `native/security_scan/scan.rs`: `pub fn scan(text: &str) -> Vec<Finding>`, pure, no I/O
[...]

Issue #597, opened 2026-09-18
Epic: Credentials Checker — detect leaked credentials in session transcripts


vibexp-agent:
## Summary

Agento stores every Claude Code session transcript locally, and nothing today scans that corpus for accidentally-leaked credentials (an API key echoed in a `Bash` tool's stdout, a `.env` file read into a tool result, a secret pasted into a prompt).

This epic tracks a **Credentials Checker**: a background scanner over session transcripts using high-confidence regex rules (ported from gitleaks' ruleset — no mature Rust crate exists for this), surfaced in a new **Security** nav section with a confidence score per finding, an opt-in toggle in Settings (default **off**), and whitelist/false-positive handling. A future phase adds real-time detection via Claude Code hooks.

Full design doc: https://vibexp.shaharialab.com/artifacts/203ca41f-c0ae-4a6c-92c8-62fd3c61b71b/design-credentials-checker

## Scope

**v1 (this epic's core path):**
- [ ] Rule engine — vendored high-confidence credential regexes
- [ ] Schema — `credential_findings` + `credential_whitelist` tables
- [ ] Persistence — whitelist-aware store
- [ ] Settings field — enable/disable, default off
- [ ] Background worker — boot sweep + incremental scan, gated by the setting
- [ ] `/api` endpoints — list findings, whitelist/false-positive, scan status
- [ ] UI — Security nav section + Credentials Checker page
- [ ] UI — Settings → Security: Credentials Checker sub-section

**Explicitly sequenced after v1 (tracked, not blocking):**
- [ ] Real-time detection via Claude Code hooks
- [ ] Entropy-adjusted / generic-secret detection tuning
- [ ] Provider verification of detected credentials (TruffleHog-style)

## Non-goals for v1

- No entropy-only heuristic (too noisy — see design doc §4.1)
- No provider-side verification of a candidate secret
- No editing/redaction of the transcript itself — detection and reporting only

Sub-issues and their native GitHub relationships (parent/blocked-by) are linked below.

PR #617, opened 2026-09-18
feat: Credentials Checker store — whitelist-aware persistence for findings (#602)

### LINKED ISSUE #602: Credentials Checker: store — whitelist-aware persistence for findings
Part of #597.

Add `native/security_scan/store.rs`, modeled on `native/insights/store.rs`: what needs (re)scanning, the upsert, keyed on `(session_id, project_path)`. Before writing an `open` finding, check `credential_whitelist` by both match-hash and rule_id — a whitelisted match must never appear as `open`, not merely be hidden after the fact (design doc §5.2). A ruleset-version bump (from the rules issue) must requeue every session for rescan, the same mechanism as `CURRENT_PROCESSOR_VERSION` in insights.

Also covers: re-evaluating existing `open` findings when a new whitelist entry is added (§6) — a whitelist add should retroactively suppress matching existing findings, not just future ones.

Design doc: https://vibexp.shaharialab.com/artifacts/203ca41f-c0ae-4a6c-92c8-62fd3c61b71b/design-credentials-checker (§5.1–§5.2, §6)

Blocked by:
- Rule engine (rules.rs / scan.rs — needs rule ids + confidence tiers)
- Schema migration (needs the two tables)

---
## WHAT
**Summary:** `native/security_scan/store.rs` — the persistence layer over the tables from #600 — plus the migration that gives `credential_findings` the `match_hash` column its own by-hash whitelisting acceptance criterion requires.

**Scope:**
- A new appended migration (next number after 41 — re-derive from `migrate.rs`'s tail at implementation time) adding a nullable `credential_findings.match_hash TEXT` column and an index on it
- `needs_scanning(conn, ruleset_version) -> Vec<Pending>` — sessions whose `credential_scan_state` row is missing or behind `ruleset_version`, modeled on `insights::store::needs_processing`
- `record_scan(conn, session_id, project_path, ruleset_version, findings) -> Result<(), String>` — one transaction that writes/updates `credential_scan_state` and upserts each non-whitelisted finding into `credential_findings`, **now including `match_hash`**
- `is_whitelisted(conn, rule_id, match_hash) -> bool` (or a batch equivalent) — checked before any finding is written as `open`
- `add_whitelist_entry` / `remove_whitelist_entry` / `list_whitelist` — CRUD on `credential_whitelist`
- Re-evaluation: adding a whitelist entry updates every existing matching `open` finding's `status` to `whitelisted` in the same call, matching **either** `rule_id` **or** the new `match_hash` column

**Affected Components:**
- `src-tauri/src/native/migrate.rs` — new migration entry + doc-comment line
- `parity/migrations_vectors.json` — the new migration appended as *text*
- `src-tauri/src/native/security_scan/store.rs` — new

**Acceptance Criteria:**
- [ ] The new migration adds `credential_findings.match_hash TEXT` (nullable — a finding pre-dating this migration has none) and an index on it, additive, and updates the hardcoded migration count assertion in `migrate.rs`'s tests
- [ ] `record_scan` computes and stores `match_hash` for every finding it writes, using **one hashing scheme shared with `add_whitelist_entry`'s** `credential_whitelist.match_hash` write — both are first written by this same module, so the scheme is decided once here and used identically by both call sites, not independently
- [ ] `needs_scanning` returns a session when it has no `credential_scan_state` row, or when that row's `ruleset_version` is behind the current one — keyed on `(session_id, project_path)`, never `session_id` alone
- [ ] `record_scan` is one transaction: writes/updates the `credential_scan_state` row unconditionally (even when zero findings), and inserts a `credential_findings` row per match **only if** it is not whitelisted by hash or by rule_id
- [ ] Re-running `record_scan` for the same `(session_id, project_path, rule_id, location_start)` is idempotent (the `UNIQUE` constraint from #600 backs this — upsert, not insert-or-fail)
- [ ] Adding a whitelist entry **by hash** immediately flips every existing `open` finding whose `match_hash` matches to `whitelisted` — verified by a test that seeds an `open` finding with a known `match_hash`, adds a matching whitelist entry, and asserts the finding's status changed with no rescan. Adding one **by `rule_id`** does the equivalent for every finding with that rule.
- [ ] A ruleset-version bump (from #599's `CURRENT_RULESET_VERSION`) makes `needs_scanning` return every session again
- [ ] No function in this module ever holds a raw secret value in memory beyond the scope of computing the masked snippet and hash handed to it by the caller — it stores what #599's `Finding` + rules already redacted, plus the hash

**Out of Scope:**
- The worker loop that calls `needs_scanning`/`record_scan` on a schedule (#603)
- The `/api` surface exposing whitelist CRUD to the UI (#604) — #604 reads a finding's `match_hash` from this migration's new column, so it stays blocked on this issue, not on a separate schema issue

## WHY
**Business Value:** This is the only place scan results and whitelist state are read or written — every other v1 component (worker, API, UI) goes through it rather than touching the tables directly, keeping the redaction and whitelist-suppression invariants in one place.

**Problem Statement:** Without this layer, a whitelisted match would need to be filtered at every read site instead of never being written as `open` in the first place (design §5.2). Without `match_hash` on `credential_findings`, the retroactive-whitelist-by-hash requirement is unmeetable: nothing in the database can tell which existing findings share a whitelisted value's hash once the raw value is (correctly) never persisted.

**Expected Impact:** Correct suppression semantics — including retroactive by-hash suppression — and no repeated full-corpus rescans once a session is at the current ruleset version.

**Context:**
- Parent: #597
- Blocked by: #599 (rule ids/confidence tiers), #600 (schema)
- Blocks: #603 (worker), #604 (API — its "whitelist by value hash" endpoint needs this column)
- **This corrects the previous refinement's premise.** The prior version assumed the by-hash retroactive-whitelist criterion could be met against migration 41's tables as merged, with `UPDATE … WHERE status = 'open' AND (rule_id = ?1 OR <matches by hash>)`. It cannot: `credential_findings` as merged in #600/#615 has no `match_hash` column, only `rule_id`, `masked_snippet` and byte offsets — `match_hash` exists only on `credential_whitelist`. The 2026-09-18 implementation run stopped at HOW revalidation for exactly this reason (see the issue comment) and flagged that #604 hits the same wall. Rather than splitting the column into a separate schema issue that blocks this one (the run's other suggested option), it is folded into this issue's scope, the same way #600 was itself amended in place for the `credential_scan_state` table this issue depends on — one implementer touches `store.rs` and its migration together instead of two issues serializing on a one-column ALTER TABLE.
- Design doc: https://vibexp.shaharialab.com/artifacts/203ca41f-c0ae-4a6c-92c8-62fd3c61b71b/design-credentials-checker (§5.1–§5.2, §6)

## HOW
**Technical Approach:** Directly mirror `native/insights/store.rs`'s shape — a `Pending`-like struct, a `LEFT JOIN`-based `needs_scanning` query, and a single-transaction upsert — substituting `claude_session_cache` as the left side of the join. Precede it with a small additive migration (mirrors migration 38's single-`ALTER TABLE` shape) adding `match_hash` to `credential_findings`.

**Implementation Details:**
```rust
// migrate.rs — appended migration (renumber from the file's actual tail;
// 42 as of 2026-09-18, re-derive at implementation time)
"ALTER TABLE credential_findings ADD COLUMN match_hash TEXT;
 CREATE INDEX idx_credential_findings_match_hash ON credential_findings(match_hash);"
```
```rust
pub struct Pending {
    pub session_id: String,
    pub project_path: String,
    pub file_path: String,
}

pub fn needs_scanning(conn: &Connection, ruleset_version: i64) -> Result<Vec<Pending>, String> {
    // SELECT DISTINCT c.session_id, c.project_path, c.file_path
    // FROM claude_session_cache c
    // LEFT JOIN credential_scan_state s
    //        ON c.session_id = s.session_id AND c.project_path = s.project_path
    // WHERE s.session_id IS NULL OR s.ruleset_version < ?1
}

pub fn record_scan(
    conn: &mut Connection,
    session_id: &str,
    project_path: &str,
    ruleset_version: i64,
    findings: &[crate::native::security_scan::scan::Finding],
) -> Result<(), String> {
    let tx = conn.transaction().map_err(|e| e.to_string())?;
    // upsert credential_scan_state (session_id, project_path) unconditionally
    for f in findings {
        // let hash = hash_match(&f.raw_value);  // same scheme as credential_whitelist.match_hash
        // skip if is_whitelisted(&tx, f.rule_id, &hash)
        // INSERT ... (..., match_hash, ...) ... ON CONFLICT(session_id, project_path, rule_id, location_start) DO UPDATE ...
    }
    tx.commit().map_err(|e| e.to_string())
}
```
Whitelist re-evaluation: `add_whitelist_entry` runs its insert plus one `UPDATE credential_findings SET status = 'whitelisted' WHERE status = 'open' AND (rule_id = ?1 OR match_hash = ?2)` in the same transaction.

**Testing Strategy:** Inline `#[cfg(test)] mod tests`, in-memory `rusqlite::Connection` with migrations applied (reuse `migrate.rs`'s `the_credential_tables_key_on_the_session_pair` test's setup pattern). Cases: needs_scanning on empty state, after a scan at current version, after a ruleset bump; record_scan writing zero and non-zero findings with `match_hash` populated; whitelist-by-hash and whitelist-by-rule both suppressing at write time **and** retroactively against a pre-seeded `open` finding.

**Documentation / Configuration:** Update `migrate.rs`'s module doc comment with the new migration's entry. No new dependency.

**Files to Create / Modify:**
- Modify: `src-tauri/src/native/migrate.rs`
- Modify: `parity/migrations_vectors.json`
- Create: `src-tauri/src/native/security_scan/store.rs`

**Risks:**
- Migration number collision if another issue appends one first — re-derive from `migrate.rs`'s tail at implementation time.
- **Hashing scheme is undecided — this is the first and only consumer of `match_hash` on either table** (nothing in the codebase writes it yet). Pick one deterministic scheme (e.g. SHA-256 of the raw matched bytes) and use it identically in `record_scan` and `add_whitelist_entry`; a mismatch between the two call sites makes by-hash matching silently never fire. Document the choice in `store.rs` itself, since #600 deferred it here.

**Complexity:** Medium (multi-statement transaction logic plus a small additive migration, directly modeled on an existing store with no new architecture).

---
**Refined by:** AI Issue Refiner · **Date:** 2026-09-18 · **Status:** Ready

vibexp-agent:
Closes #602 (part of #597).

## What
- **Migration 43** — `credential_findings.match_hash TEXT` (nullable) + `idx_credential_findings_match_hash`. Additive, appended to `parity/migrations_vectors.json` as text; count/version assertions in `migrate.rs` bumped 42 → 43.
- **`native/security_scan/store.rs`** — the only read/write path over the credential tables:
  - `needs_scanning(conn, ruleset_version)` — sessions with no `credential_scan_state` row or one behind the ruleset, joined on `(session_id, project_path)`.
  - `record_scan(conn, session_id, project_path, ruleset_version, text, findings)` — one transaction: scan-state upsert (unconditional, zero findings included), then a finding upsert per match that is not whitelisted by rule id or value hash. Rescan keeps `status` and `detected_at`, and deletes this session's `open`/`whitelisted` findings the new scan did not produce (retired/tightened rules), so a later whitelist removal cannot resurrect them; `false_positive` rows are kept.
  - `add_whitelist_entry` / `remove_whitelist_entry` / `list_whitelist` / `is_whitelisted`. Adding an entry flips every matching `open` finding to `whitelisted` in the same transaction (no rescan); removing one re-opens the `whitelisted` findings no remaining entry covers. `false_positive` is never touched.
  - `hash_match` — **the single hash scheme: SHA-256 of the matched bytes, lowercase hex** (`ring`, already a dependency). `add_whitelist_entry` refuses a hash not of that shape, so the two call sites cannot drift.
  - `mask` — first 4 + `********` + last 4 chars (fixed run, so length is not recorded); matches under 16 chars are fully masked.

## Deviation from the HOW
The HOW sketches `hash_match(&f.raw_value)`, but `scan::Finding` (#599) deliberately carries only a byte range. So `record_scan` takes the scanned `text` alongside the findings and slices each match only long enough to compute the mask and hash — which is what `security_scan/mod.rs`'s header already said the store would do. An out-of-range finding fails the whole call and writes nothing.

## Tests
17 inline tests in `store.rs` (pair keying, ruleset bump, clean scan state, upsert keeping the user's verdict, write-time suppression by hash and by rule, retroactive suppression by hash across sessions and by rule, malformed hash refused, whitelist CRUD, removal re-opening only uncovered findings, rescan retracting stale open and whitelisted findings, rescan re-opening a whitelisted row whose hash changed, out-of-range rollback, SHA-256 known vector). Disabling the whitelist SQL fails 4 of them.

## Known limitation (for the maintainer)
Removing a whitelist entry re-opens only findings that have a row. A match suppressed *at write time* (whitelisted before it was first found) is never stored — the issue's AC requires that — and nothing requeues its session on removal, so it reappears only at the next ruleset bump. Fixing that means either requeueing affected sessions on removal or storing suppressed matches as `whitelisted` rows (which changes this issue's AC), so it is left as a spec decision; documented on `remove_whitelist_entry`.

Out of scope: the worker (#603), the `/api` surface (#604).


vibexp-agent:
Automated run stopped short of auto-merge: the automated review did not reach APPROVE within its 3 rounds. Round 3's last finding (a rescan leaving an un-covered row `whitelisted` after its hash changed) is fixed in 2565271 with two tests, but that commit has not been through a review round. CI is green on it. See the PR body's *Known limitation* for the one open spec question.

